PR #67, opened 2023-08-04
feat: Okta login

scott-codecov:
### Purpose/Motivation

Login to Codecov using Okta

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/145
https://github.com/codecov/engineering-team/issues/146

### What does this PR do?

Adds a new OAuth flow for Okta available at `/login/okta`

### Notes to Reviewer

Unlike other OAuth providers we've integrated with, Okta requires an "issuer" which is a domain like `https://{orgname}.okta.com`.  When clicking on a tile in Okta the issuer is passed as an `iss` query string param which we store in a cookie for the duration of the OAuth flow.

If we ever want to support some sort of "Single Sign On" button in Codecov then we would first need to ask for this issuer domain (or just `orgname`).

This will only be deployed in ECDN environments to start.  We'll configure the `OKTA_ISS` for a specific org and set `DISABLE_GIT_BASED_LOGIN` to true.  This will force members of that org to use Okta to login and only then can they link their Git provider account.


codecov_auth/views/okta.py
```diff
@@ -0,0 +1,187 @@
...
+            dict(
+                response_type="code",
+                client_id=settings.OKTA_OAUTH_CLIENT_ID,
+                scope="openid email profile",
+                redirect_uri=settings.OKTA_OAUTH_REDIRECT_URL,
+                state=iss,
+            )
+        )
+        redirect_url = f"{iss}/oauth2/v1/authorize?{qs}"
+        response = redirect(redirect_url)
```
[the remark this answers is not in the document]
scott-codecov: Maybe makes sense to check the `iss` value against a regex.  I believe it will always be of the form `https://{orgname}.okta.com`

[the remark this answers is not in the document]
scott-codecov: I think this is OK for now since the ISS will be coming from the config.

codecov_auth/migrations/0034_oktauser.py
```diff
@@ -0,0 +1,41 @@
...
+
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("codecov_auth", "0033_sentryuser"),
```
adrian-codecov, line 13: You might want to redo this migration since this PR was done a while ago, and we're in migration 0039

codecov_auth/migrations/0034_oktauser.py
```diff
@@ -0,0 +1,41 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="OktaUser",
+            fields=[
+                ("id", models.BigAutoField(primary_key=True, serialize=False)),
+                ("external_id", models.UUIDField(default=uuid.uuid4, editable=False)),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
```
adrian-codecov, line 22: Hmmm, when we use the `BaseCodecovModel`, we don't save this with the `...WithoutTZField`, we okay with that?

codecov_auth/migrations/0034_oktauser.py
scott-codecov: Oh yeah, thanks - looks like all the tests broke when I merged `main`

codecov_auth/views/github.py
```diff
@@ -111,6 +111,10 @@ def actual_login_step(self, request):
         return response
 
     def get(self, request):
+        print("HERE", settings.DISABLE_GIT_BASED_LOGIN, request.user.is_anonymous)
```
adrian-codecov, line 114: 👮‍♂️ 🚓 🧯

codecov_auth/views/okta.py
```diff
@@ -0,0 +1,191 @@
...
+def validate_id_token(iss: str, id_token: str) -> dict:
+    res = requests.get(f"{iss}/oauth2/v1/keys")
+    jwks = res.json()
+
+    public_keys = {}
+    for jwk in jwks["keys"]:
+        kid = jwk["kid"]
+        public_keys[kid] = jwt.algorithms.RSAAlgorithm.from_jwk(json.dumps(jwk))
+
+    print(jwt.get_unverified_header(id_token))
```
adrian-codecov, line 32: Do we want this?

codecov_auth/migrations/0034_oktauser.py
scott-codecov: I think it's probably fine.  We're extending from `BaseCodecovModel` in a lot of other models as well.

codecov_auth/views/okta.py
```diff
@@ -0,0 +1,191 @@
...
+            )
+
+        login(request, current_user)
+        return current_user
+
+    def get(self, request):
+        if request.GET.get("code"):
+            return self._perform_login(request)
+        else:
+            iss = settings.OKTA_ISS or request.GET.get("iss")
```
adrian-codecov, line 187: Optional: not sure if `iss` over `issuer` is understandable if you're familiar with OIDC, but otherwise I'd push for the latter for readability

codecov_auth/views/okta.py
```diff
@@ -0,0 +1,191 @@
...
+            iss,
+            max_age=86400,  # Same as state validatiy
+            httponly=True,
+            domain=settings.COOKIES_DOMAIN,
+        )
+
+    def _perform_login(self, request: HttpRequest) -> HttpResponse:
+        code = request.GET.get("code")
+        iss = settings.OKTA_ISS or request.COOKIES.get("_okta_iss")
+        if iss is None:
```
adrian-codecov, line 101: Could we ever hit this? Since at the very least we'd have the iss from `_redirect_to_consent`

scott-codecov: I'd say this would be unexpected.  IDK - maybe somebody has cookies disabled or something.

codecov_auth/views/okta.py
```diff
@@ -0,0 +1,191 @@
...
+        id_payload = validate_id_token(iss, id_token)
+
+        okta_id = id_payload["sub"]
+        user_email = id_payload["email"]
+        user_name = id_payload["name"]
+
+        okta_user = OktaUser.objects.filter(okta_id=okta_id).first()
+
+        current_user = None
+        if request.user is not None and not request.user.is_anonymous:
```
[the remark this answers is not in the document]
scott-codecov: Yeah, they'll have a single `users` record and then an `okta_users` record + `owners` record that both reference that single user.

codecov_auth/views/okta.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # TEMPORARY: we're assuming a single owner for the time being since there's
+        # no supporting UI to select which owner you'd like to view
+        owner = current_user.owners.first()
+        if owner is not None:
+            service = get_short_service_name(owner.service)
+            response = redirect(f"{settings.CODECOV_DASHBOARD_URL}/{service}")
+        else:
+            # user has not connected any owners yet
+            response = redirect(f"{settings.CODECOV_DASHBOARD_URL}/sync")
```
adrian-codecov, line 118: Is this a new url? I don't think I've seen this sync endpoint 👀

codecov_auth/views/okta.py
scott-codecov: The query string param has to be `iss` since that's what Okta sends.  I think since it's referred to as `iss` elsewhere (i.e. in the ID token, query param, etc.) then I'm inclined to keep it as `iss` so you know it's the same thing.  I expect you'd need some familiarity with OIDC to fully follow this code anyway.

codecov_auth/views/okta.py
adrian-codecov: Yeaah that's fair, asking in case this could be dead code that adds complexity, but I think that's fair. In that scenario they wouldn't be able to use okta hmmm, unlikely, but we can keep it in here and see if we hit this

codecov_auth/views/okta.py
scott-codecov: Yeah a new page that Nick created to support "Login with Sentry".  Once a user is authenticated but doesn't have any linked Git providers we send them to this page (where they can sync a Git provider).

scott-codecov: Oh it's not merged yet: https://github.com/codecov/gazebo/pull/2235

codecov_auth/views/okta.py
adrian-codecov: Gotchaa, yeap that makes sense! We can keep it as is

adrian-codecov (approved):

PR #71, opened 2023-08-08
Implement local shared toggling/hot reloading

joseph-sentry:
### Purpose/Motivation
This commit enables developers to create dev images of the API that use the local copy of shared, and
optionally allow for hot reloading of shared.

### Links to relevant tickets
https://github.com/codecov/worker/pull/29

### What does this PR do?
- Create `dev` and `dev-shared` make targets used to create images that use local/hot-reloaded shared
- Move shared dependency from targeting git to targeting local shared
- Add shared as a submodule 
- Create dev-shared script to enable toggling of hot reload shared


dev-shared.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
matt-codecov, line 1: api's `dev.sh` uses `gunicorn --reload` which already gives us hot-reloading. ideally we don't need to add another layer with `watchmedo` here

however, `gunicorn --reload` doesn't appear to have a way to ignore files matching a pattern like `watchmedo` does which is a bummer

because this is just in our dev environment, we could possibly try moving `dev.sh` to use `uvicorn` instead which has `--reload-exclude` https://www.uvicorn.org/#command-line-options

(i googled the difference: `gunicorn` is a WSGI server which means it processes requests serially and `uvicorn` is an ASGI server which means it can process requests concurrently.)

alternatively, a hot-reload honestly isn't that disruptive, if we run it for shared no matter what it wouldn't be the worst thing in the world

in any case, i think we can add what we want to `dev.sh` and skip adding this second file
```bash
if [ "$EDITABLE_SHARED" = "y" ]
then
    pip install -e ./shared
    uvicorn blah blah blah
else
    gunicorn blah blah blah
fi
```

.dockerignore
```diff
@@ -14,5 +14,4 @@ nosetests.xml
 coverage.xml
 *,cover
 *.log
-.git
```
matt-codecov, line 17: :o what did this fix? do we need it in worker too?

joseph-sentry: nope, if we're ignoring .git we don't have any information about this directory being a git repository so we wouldn't be able to run any git commands, `git submodule` fails because we aren't in a git repo in the container when `.git` is ignored

PR #72, opened 2023-08-08
fix: lru_cache issues + meta info missing 

giovanni-guidini:
Context: https://github.com/codecov/engineering-team/issues/119

So the real issue with the meta info is fixed in https://github.com/codecov/shared/pull/22.
spoiler: reusing the report details cached values and _changing_ them is not a good idea.

However in the process of debuging that @matt-codecov pointed out that we were not using lru_cache correctly.
Check this very well made video: https://www.youtube.com/watch?v=sVjtp6tGo0g

So the present changes upgrade shared so we fix the meta info stuff AND address the cache issue.
There are further complications with the caching situation, which explain why I decided to add the cached value in the
`obj` instead of `self`. The thing is that there's only 1 instance of `ArchiveField` shared among ALL instances of
the model class (for example, all `ReportDetail` instances). This kinda makes sense because we only create an instance
of `ArchiveField` in the declaration of the `ReportDetail` class.

Because of that if the cache is in the `self` of `ArchiveField` different instances of `ReportDetails` will have dirty cached value of other `ReportDetails` instances and we get wrong values. To fix that I envision 3 possibilities:
1. Putting the cached value in the `ReportDetails` instance directly (the `obj`), and checking for the presence of that value.
If it's there it's guaranteed that we put it there, and we can update it on writes, so that we can always use it. Because it is
for each `ReportDetails` instance we always get the correct value, and it's cleared when the instance is killed and garbage collected.

2. Storing an entire table of cached values in the `self` (`ArchiveField`) and using the appropriate cache value when possible. The problem here is that we need to manage the cache ourselves (which is not that hard, honestly) and probably set a max value. Then we will populate the cache and over time evict old values. The 2nd problem is that the values themselves might be too big to hold in memory (which can be fixed by setting a very small value in the cache size). There's a fine line there, but it's more work than option 1 anyway.

3. We move the getting and parsing of the value to outside `ArchiveField` (so it's a normal function) and use `lru_cache` in that function. Because the `rehydrate` function takes a reference to `obj` I don't think we should pass that, so the issue here is that we can't cache the rehydrated value, and would have to rehydrate every time (which currently is not expensive at all in any model)

This is an instance cache, so it shouldn't need to be cleaned for the duration of the instance's life
(because it is updates on the SET)

closes https://github.com/codecov/engineering-team/issues/119


matt-codecov (commented):
i really don't understand the issue here so it's hard to evaluate whether this is the correct fix. how are you reproducing this? does the issue consistently happen without your fix and then consistently not happen with your fix?

the field data takes a few forms before we return it to the caller: the raw file string, the deserialized json, and the rehydrated result. you are finding that the issue goes away if we move the caching from the rehydrated result to the raw file string, which suggests the problem is either in `json.loads()` or `self.rehydrate_fn()`. could you try caching the deserialized json to see if that also fixes the issue? it would help narrow down the cause, and also deserializing might be expensive so caching it would be nice

also, i wonder if the issue happens if we cache the old fashioned way:
```
if self.cached_val = None:
    self.cached_val = self.rehydrate(json.loads(archive_service.read(filename)))
return self.cached_val
```

that all said, my gut tells me it has to be an edge case. where is "meta" supposed to actually come from? i can find places that assume it exists, but none that create it.

incidentally, i actually found an issue with how we use `@lru_cache()`, but it shouldn't be the cause of this issue. using `@lru_cache()` on instance methods this way winds up leaking memory as the cache will hold onto references to `self` forever. instead you can do something like:
```
class Foo:
    def __init__(self):
        self.cached = functools.lru_cache(maxsize=1)(_uncached)

    def _uncached(self):
        return "foo"
```

giovanni-guidini:
> how are you reproducing this? 
I just setup the environment locally and made some upload locally and all that. With some special logs to snoop each part of the de-serialization. In the context issue there's more about the specifics.

> does the issue consistently happen without your fix and then consistently not happen with your fix?
Exactly

> which suggests the problem is either in json.loads() or self.rehydrate_fn(). 
Not quite, because I got consistently expected results _before_ returning. The issue really has to be in the caching of the object that comes out of `json.loads()` (a dict). The rehydrate function in this case does nothing (`lambda self, x: x`) so I don't think there's an issue there.

> also, i wonder if the issue happens if we cache the old fashioned way
I think that's simple and universally understandable. I am willing to make this change.

> where is "meta" supposed to actually come from?
It's created here: https://github.com/codecov/shared/blob/39f6f4a383d37755a69d4dfbe40390fcac45536b/shared/reports/types.py#L260 when you serialize a `Report`

giovanni-guidini:
> that fixes the memory leak problem + behaves like your option 2. and it picks a default cache size, but leaves it open to configure per field if we ever need to

My understanding of the "memory leak" is that it holds on to the references of the objects passed to `lru_cache`. That was the case of `self` in the video. Your suggestion does fixes that (although we can argue that it was not a problem because we only have 1 `ArchiveField` instance that lives until the end of the process anyway).

We then need to be careful with the `obj` reference that is passed to `_uncached_get_value_from_archive`. Otherwise those references might leak, and that would be a big leak. It is doable, but I personally think that if we are to do that it's better to go straight to option 3 and move the function outside `ArchiveField` completely... maybe compromise somewhere between 2 and 3 and use the `lru_cache` inside the `ArchiveField` class is a good option

matt-codecov (approved):

PR #73, opened 2023-08-08
fix: read github client_id from config / use new default

matt-codecov:
the old development app redirected to port 80 upon successful auth and relied on the user having nginx running and reverse-proxying port 80 to the API. it was also an OAuth app and not a new-style GitHub App.

the new app is a GitHub app, and also by making this value be read from config, you can keep using the old one or sub in a new one more easily (for e.g. pointing webhooks at localhost).

### Legal Boilerplate


codecov/settings_dev.py
```diff
@@ -22,7 +22,7 @@
 
 CORS_ALLOW_CREDENTIALS = True
 
-GITHUB_CLIENT_ID = "3d44be0e772666136a13"
+GITHUB_CLIENT_ID = get_config("github", "client_id", default="Iv1.220bfc5d9b825038")
```
adrian-codecov, line 25: Is the default a sensitive secret we'd want to hide?

Also I see some failing tests

matt-codecov: oh it didn't occur to me to run tests, wow. and email is so noisy i didn't see the notice. thanks for the catch!

the ID's not sensitive, there's a separate `client_secret` field which is + is in the yaml only

giovanni-guidini (approved):

PR #75, opened 2023-08-10
Make search case insensitive

adrian-codecov:
### Purpose/Motivation
Codecov's client can search for different files in a coverage report, and it is currently case sensitive. This PR is to enable case insensitivity on the backend when searching for a Codecov file.

### What does this PR do?
- Adjust ReportPath's functionality to be case insensitive if there is a search term parameter
- Adjust relevant tests


graphql_api/tests/test_branch.py
```diff
@@ -386,6 +386,7 @@ def test_fetch_path_contents_with_files_and_search_value(
         critical_files.return_value = [CriticalFile("folder/fileB.py")]
 
         data = self.gql_request(query_files, variables=variables)
+        print("dataaa", data)
```
RulaKhaled, line 389: interesting

services/path.py
```diff
@@ -160,7 +160,10 @@ def __init__(
 
         if search_term:
             self._paths = [
-                path for path in self.paths if search_term in path.relative_path
+                path
+                for path in self.paths
+                if search_term in path.relative_path
+                or search_term.lower() in path.relative_path
```
RulaKhaled, line 166: guess you can just do `path for path in self.paths if search_term in path.relative_path if search_term.lower() in path.relative_path.lower()`

graphql_api/tests/test_branch.py
adrian-codecov: 🙃 I'd removed this in my local and didn't push

RulaKhaled (approved):

PR #77, opened 2023-08-10
chore: update shared + adopt mock_metrics and mock_config_helper test utils

matt-codecov:
https://github.com/codecov/shared/pull/23

many test files roll their own way to mock metrics and config, and i've copy-pasted more-or-less these helpers in a couple files myself. bit the bullet and put them in `shared` and converted a couple tests in `codecov-api` to use them as illustrative examples

not going to sit down and convert everything, but hopefully folks can follow this pattern in the future and we can gradually convert older files when we're doing other work in them

### Legal Boilerplate


giovanni-guidini (approved):
I like this standardization in testing methods. Thanks for biting the bullet for us :P

PR #79, opened 2023-08-11
Add local testing support

adrian-codecov:
### Purpose/Motivation
This PR is to add more explicit instructions for testing codecov-api local changes.

### What does this PR do?
- Adds a docker-compose file for serving necessary persistence containers for testing standalone codecov-api
- Adjusts readme.md to give instructions clearer instructions for local testing

### DIFF
--- a/docker-compose-local-testing.yml
+++ b/docker-compose-local-testing.yml
@@ -9,10 +9,10 @@ services:
     networks:
       - codecovapi
     ports:
-      - "5436:5432"
+      - "5432:5432"
     environment:
-      - POSTGRES_USER=postgres
-      - POSTGRES_PASSWORD=postgres
+      - POSTGRES_HOST_AUTH_METHOD: trust
+      - POSTGRES_PASSWORD: postgres
   redis:
     hostname: redis
     image: redis:4.0-alpine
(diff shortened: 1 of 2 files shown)


docker-compose-local-testing.yml
```diff
@@ -0,0 +1,32 @@
...
+  postgres:
+    hostname: postgres
+    image: postgres:14.4-alpine
+    volumes:
+      - postgres-volume:/var/lib/postgresql/data
+    networks:
+      - codecovapi
+    ports:
+      - "5436:5432"
+    environment:
```
joseph-sentry, line 13: For the environment I would set `POSTGRES_HOST_AUTH_METHOD=trust` instead of using `POSTGRES_USER` and `POSTGRES_PASSWORD`

docker-compose-local-testing.yml
```diff
@@ -0,0 +1,32 @@
...
+services:
+  postgres:
+    hostname: postgres
+    image: postgres:14.4-alpine
+    volumes:
+      - postgres-volume:/var/lib/postgresql/data
+    networks:
+      - codecovapi
+    ports:
+      - "5436:5432"
```
joseph-sentry, line 12: I would prefer to have this use port 5432 and have redis use port 6379 since I think the settings use those ports as default and it would be easier for devs to avoid modifying those

PR #85, opened 2023-08-17
feat: support gh refresh tokens

giovanni-guidini:
Depends on codecov/shared#27
TODO: after that is merged update the sha reference in requirements.in

Adds support for github app refresh tokens

codecov/engineering-team#162

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


services/repo_providers.py
```diff
@@ -70,7 +70,9 @@ def get_generic_adapter_params(owner: Owner, service, use_ssl=False, token=None)
             key=getattr(settings, f"{service.upper()}_CLIENT_ID", "unknown"),
             secret=getattr(settings, f"{service.upper()}_CLIENT_SECRET", "unknown"),
         ),
-        on_token_refresh=get_token_refresh_callback(owner, service),
+        on_token_refresh=(
+            get_token_refresh_callback(owner, service) if "username" in token else None
```
scott-codecov, line 74: What is `if "username" in token` checking for here?

scott-codecov: Oh I saw the code in https://github.com/codecov/worker/pull/69 about checking whether we're using the GH app integration.  Is this a proxy for that?  If so maybe just a quick comment here

giovanni-guidini: Yeah that's the idea. Since we define "username" in the token on line 62 - IFF we know what's the owner of said token - then we can use that to check if the token belongs to an owner (and thus can/should be refreshed).

I'll update with a comment

PR #89, opened 2023-08-22
Fix enterprise build

juho9000:
### Purpose/Motivation
The enterprise build is currently broken.

### What does this PR do?
The utils/config.py file contains a function that is missing in enterprise/config.py and this results in a broken build. The plan-module also does not make it into the enterprise build due to the way the cythonize scripts work. The added init file fixes this. 

### Notes to Reviewer
Try building and running the enterprise-runtime or even pulling and running the newest stable enterprise-runtime Docker image and you will notice it is broken.

### Legal Boilerplate


zetaab:
cc @JerrySentry codecov-api master is currently broken. This is fixing the issue with enterprise build.

PR #92, opened 2023-08-28
Make uses_invoice field on Owner non-nullable

joseph-sentry:
### Purpose/Motivation
This is a migration to create the uses_invoice field in order to keep track of which users are using invoice billing. 

### Links to relevant tickets
Fixes: https://github.com/codecov/engineering-team/issues/131

### What does this PR do?
- Create migration to create uses_invoice nullable column in Owner table
- Create migration to set default of uses_invoice column to false
- Create migration to make uses_invoice field non-nullable


scott-codecov (approved):
Just a note/reminder that since the code depends on the column existing (since it was defined in the model) make sure that these risky migrations are run before merging/deploying this PR.

joseph-sentry:
> Just a note/reminder that since the code depends on the column existing (since it was defined in the model) make sure that these risky migrations are run before merging/deploying this PR.

Yes, the risky migrations have been run already!

PR #95, opened 2023-08-29
[WIP] fix: Remove duplicate values in owners.admins and fix adding dupes in the future (#193)

JerrySentry:
### Purpose/Motivation

There is a bug when the membership list is loaded in the app by an admin of the org, their ownerid will be added to the admins column regardless if it was already there. On the admin page, it will show hundreds of the same admins for an org.

This fix will remove all duplicate ownerids in `owners.admins` column for every row. Should lead to improved DB disk usage and API query speed as well.

Note that the migration script runs the "risky" check, which means it doesn't run in CI when building/deploying the app in production. Need to run it manually by dev-ops after the release is deployed.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/193

### What does this PR do?
- Fix bug in django
- Provide a django migration script to remove existing duplicates


.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 repos:
```
JerrySentry, line 1: upgrading this as the previous version there's a bug with some error something like:
`ImportError: cannot import name '_unicodefun' from 'click'`

PR #98, opened 2023-08-30
ref: Add agreement in user and sync updates with owner agreement (#390)

JerrySentry:
### Purpose/Motivation
What is the feature? Why is this being done?

This is for auth refactor, add signed agreements flag to user table for users signed in via a non-git provider (ie not row in owners table).

Part 1 of 3 releases for this overall task.

Adds new columns and updates user.agreements when owner.agreements is updated.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/390

PR #100, opened 2023-08-31
ref: Populate agreement columns in user table (#98)

JerrySentry:
### Purpose/Motivation

- Migrate all the values from OwnerProfile term agreement columns to User term agreement columns
- When updating term agreement will update only User columns (via `saveTermsAgreement` mutation)
- Add retrieving term agreement from User columns (via `/internal/user` endpoint and GraphQL `me` endpoint)
- Tested locally and on staging. Checked staging DB for migrations applied and endpoints reading/writing in correct tables.

### Links to relevant tickets

### What does this PR do?
Release 2 of https://github.com/codecov/engineering-team/issues/390#issuecomment-1698094168

### Legal Boilerplate


codecov_auth/admin.py
```diff
@@ -49,18 +49,6 @@ def impersonate_owner(self, request, queryset):
 impersonate_owner.short_description = "Impersonate the selected owner"
```
JerrySentry, line 51: Remove the owner profile section in admin page as there's nothing left without agreements, instead add it to user form

codecov_auth/migrations/0037_sync_user_terms_agreement.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 4.2.2 on 2023-08-30 20:30
```
JerrySentry, line 1: Copy every owner objects' term agreement values (via owner profile) to its associated user's agreement columns

codecov_auth/commands/owner/interactors/tests/test_save_terms_agreement.py
```diff
@@ -28,45 +28,39 @@ def execute(
...
+        user: User = User.objects.filter(id=self.current_owner.user_id).first()
+        assert user.terms_agreement == True
+        assert user.terms_agreement_at == timezone.datetime(2022, 1, 2)
 
         self.current_owner.refresh_from_db()
         self.current_owner.business_email == before_refresh_business_email
 
     @freeze_time("2022-01-03T00:00:00")
-    def test_update_owner_and_profile_when_email_isnt_empty(self):
+    def test_update_owner_and_user_when_email_isnt_empty(self):
```
adrian-codecov, line 55: nitpick: change `isnt` to `is_not`. I originally created the test with `isnt`, so not your doing, but I recently installed this extension, https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker, that highlights typos, so in the spirit of slowly moving our codebase to proper names

codecov_auth/migrations/0037_sync_user_terms_agreement.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.db import migrations
+
+
+def sync_agreements(apps, schema):
+    owners = apps.get_model("codecov_auth", "Owner")
```
adrian-codecov, line 7: Is there a benefit of doing it this way over importing the Owner model like `from codecov_auth.models import Owner` and getting rid of this line?

codecov_auth/tests/test_admin.py
```diff
@@ -190,25 +185,6 @@ def test_inline_orgwide_add_token_permission_no_token_user_not_in_enterprise_clo
...
-    def test_inline_owner_profile_permissions(self):
-        owner = OwnerFactory(name="test-owner", service="github")
-        owner.save()
-        request_url = reverse(f"admin:codecov_auth_owner_change", args=[owner.ownerid])
-        request = RequestFactory().get(request_url)
-        request.user = self.staff_user
-        inlines = self.owner_admin.get_inline_instances(request, owner)
-        inline_instance = inlines[1]
-        assert inline_instance.has_delete_permission(request, owner) == False
-        assert inline_instance.has_change_permission(request, owner) == True
```
adrian-codecov, line 210: Did you confirm this logic still works in the django admin for the User model? People using the django admin have come with issues in the past where they are changing other fields in an Owner/User model, and they can't save them because some fields (in this case the terms fields) are required. Wanting to confirm if you can go edit a User entry and have the terms fields not required to save the change.

graphql_api/tests/test_current_user_ariadne.py
```diff
@@ -179,7 +179,7 @@ def test_fetch_terms_agreement_and_business_email_when_owner_profile_is_null(sel
         }
         """
         data = self.gql_request(query, owner=current_user)
-        assert data == {"me": {"businessEmail": None, "termsAgreement": None}}
+        assert data == {"me": {"businessEmail": None, "termsAgreement": False}}
```
adrian-codecov, line 182: Why did this value go from None to False?

For termsAgreement here:
- True means they accepted our terms
- None means they haven't agreed/disagreed with our terms (they haven't interacted with the forms)
- False means they declined our terms

So there's an important distinction here. I think this might be a side-effect of your factory defaulting to False instead of None.

graphql_api/types/me/me.py
```diff
@@ -104,11 +104,7 @@ def resolve_tracking_data(current_user, _, **kwargs):
 @me_bindable.field("termsAgreement")
 @sync_to_async
 def resolve_terms_agreement(current_owner: Owner, _, **kwargs) -> Optional[bool]:
-    try:
-        owner_profile = current_owner.profile
-        return owner_profile.terms_agreement
-    except OwnerProfile.DoesNotExist:
-        return None
+    return current_owner.user.terms_agreement
```
adrian-codecov, line 107: just to confirm, an owner will always have a user right?

api/internal/user/serializers.py
```diff
@@ -14,6 +14,8 @@ class Meta:
             "name",
             "external_id",
             "owners",
+            "terms_agreement",
+            "terms_agreement_at",
```
adrian-codecov, line 18: I believe we don't need to add these here as we're consuming them via the graphql resolvers. Or did you have another reason to add them here for?

codecov_auth/migrations/0037_sync_user_terms_agreement.py
JerrySentry: I just did it to keep it consistent with other migrations files are retrieving models in `codecov_auth`, not sure about other apps

codecov_auth/tests/test_admin.py
JerrySentry: Interesting! It certainly does error in admin page. (When the fields are null it errors, not when there is existing value.)

graphql_api/tests/test_current_user_ariadne.py
JerrySentry: I set agreement to False by default when a user is created because it follows the logic of setting agreement in owner profile to False when an Owner is created.
https://github.com/codecov/codecov-api/blob/ab2eaaac8e0a007b6e1232b79a9fa1df0522b05c/codecov_auth/signals.py#L17

graphql_api/types/me/me.py
JerrySentry: I believe so as this query requires user to be authenticated already.

api/internal/user/serializers.py
JerrySentry: This is requested for a change in the `internal/user` endpoint as part of being able to login without a service provider.
https://github.com/codecov/engineering-team/issues/390#issue-1871820224
The current graphql endpoints are dependent on a service provider, however another ticket is to add graphql queries to be independent of services.

codecov_auth/commands/owner/interactors/tests/test_save_terms_agreement.py
JerrySentry: Done. Cool tool, installed too.

codecov_auth/tests/test_admin.py
JerrySentry: Fixed. See [codecov_auth/models.py](https://github.com/codecov/codecov-api/pull/100/files#diff-e2e21324fdebb10e0effbcbc3cc096e121165c4476b1ff8428bb3a51b34ef14b)

codecov_auth/migrations/0037_sync_user_terms_agreement.py
```diff
@@ -0,0 +1,31 @@
...
+    for owner in owners.objects.all():
```
scott-codecov, line 8: This could take a long time to loop through all the owners in the database - I wonder if it makes sense to do this in a Django management command instead of a migration.

Also, consider using https://docs.djangoproject.com/en/4.2/ref/models/querysets/#iterator to avoid loading the full set of owner records into memory here.

scott-codecov: @trent-codecov What are your thoughts here?  Is it OK for the migrate pod to execute a really long process like this?

scott-codecov: Actually, looking a little closer at the code inside the loop - I think this could all be replaced with some SQL that should be much more efficient:

```sql
UPDATE users
SET users.terms_agreement_at = subquery.terms_agreement_at
FROM (
    SELECT
        owners.user_id, codecov_auth_ownerprofile.terms_agreement_at
    FROM owners
    INNER JOIN codecov_auth_ownerprofile
    ON codecov_auth_ownerprofile.owner_id = owners.ownerid
) subquery
WHERE subquery.user_id = users.id;
```

Look into `RiskyRunSQL` if you want to take this approach instead of the code-based approach.

JerrySentry: @scott-codecov  Switched to using SQL which seems like a better option for this.
Regarding `RiskyRunSQL`, my understanding is we don't need it since we're not altering the owners table right? A side effect with using `RiskyRunSQL` is that there will be an interval of time between the new code changes being deployed and whenever we manually run the SQL, during that interval data is reading/writing from users table before the data is migrated over from the ownerprofile table. This can lead to writes being overwritten and reads being wrong. There is probably a small chance for this to happen though, thoughts?

codecov_auth/tests/test_admin.py
adrian-codecov: peeerfect, thanks!

graphql_api/tests/test_current_user_ariadne.py
adrian-codecov: I believe though since we're moving from profile to user here, then the test for this should be adapted to deleting the user instead of the owner profile. I think it's odd that the logic is meant to "remain the same" so our test resutls shouldn't change, but they are here 👀

api/internal/user/serializers.py
adrian-codecov: A gotcha, was not aware, ty for that!

codecov_auth/migrations/0037_sync_user_terms_agreement.py
trent-codecov: @scott-codecov generally prefer to avoid long running migrate pods when possible but it's fine if necessary. I like your sql approach.

@JerrySentry the main point of risky sql / migrations is to allow you to deploy your code decoupled from sql that can/will lock the db. Then the infra team can run the sql after hours. 

Ideally you design your code such that the deploy with the risky sql can be done first. In the case that is not possible, we can just do the merge after hours, downside being you will need to be around to help just in case.

PR #101, opened 2023-09-01
Add changes for monthly uploads to account for trialing customer

adrian-codecov:
### Purpose/Motivation
Uploads while undergoing a trial shouldn't affect the monthly uploads post trial. This is a code change to adjust this in the API level (change in worker coming soon)

### What does this PR do?
- Adding logic to determine the monthly upload created date + tests
- Adding some properties + tests to the plan service

### Notes to Reviewer

### Legal Boilerplate


codecov_auth/commands/owner/interactors/get_uploads_number_per_user.py
```diff
@@ -18,11 +16,23 @@ def execute(self, owner: Owner):
...
+                plan_service.trial_status == TrialStatus.EXPIRED.value
+                and plan_service.has_trial_dates
+            ):
+                trial_duration_in_days = plan_service.trial_duration_in_days()
+                time_since_trial_end_in_days = (
+                    datetime.utcnow() - plan_service.trial_end_date
+                ).days
+                if time_since_trial_end_in_days < trial_duration_in_days:
+                    created_at = plan_service.trial_end_date
+
```
JerrySentry, line 30: Isn't there the case where the trial start and end dates are within the last month? For example,
Today (now): Apr 30
Trial End: Apr 25
Trial Start: Apr 11
-> This will compute the count of interval from Apr 25 to Apr 30 (now), but it is missing Apr 1 to Apr 11 interval

adrian-codecov: hmmm, you're right. The solution is to exclude uploads that happened during the trial. I think the solution is to count the amount of days someone was on trial, add those to the time delta, so `created_at__gte= now - (30 + time_on_trial)`, and then exclude the uploads during the days of trial

codecov_auth/commands/owner/interactors/get_uploads_number_per_user.py
```diff
@@ -18,11 +16,23 @@ def execute(self, owner: Owner):
...
             uploads_used = ReportSession.objects.filter(
+                upload_type="uploaded",
                 report__commit__repository__author_id=owner.ownerid,
                 report__commit__repository__private=True,
-                created_at__gte=timezone.now() - timedelta(days=30),
-                report__commit__timestamp__gte=timezone.now() - timedelta(days=60),
-                upload_type="uploaded",
+                created_at__gte=created_at,
+                report__commit__timestamp__gte=datetime.utcnow() - timedelta(days=60),
             )[:monthly_limit].count()
```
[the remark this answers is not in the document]
adrian-codecov: I believe so yeah

scott-codecov (approved):

PR #102, opened 2023-09-05
feat: Django command to enqueue commit backfill tasks

scott-codecov:
### Purpose/Motivation

A way to systematically backfill commit data for our entire set of commits.  Depends on the task to actually do the backfilling here: https://github.com/codecov/worker/pull/78

This is just a way to orchestrate the running of that task.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/190

### What does this PR do?

Creates a new Django management command that will be run on an interval (i.e. via cron).  It grabs a batch of commits and equeues a backfill task for each commit.  The next time it runs it will grab the preceding batch, etc. until all commits have eventually been processed.

We'll want to manually set the `backfill_commits_id` key in Redis (as a starting point for commits to backfill) before running this the 1st time.  We'll want to tune the frequency at which this runs such that we consistently drain the queue between runs but also don't sit idle for too long.


core/management/commands/backfill_commits.py
```diff
@@ -0,0 +1,63 @@
...
+        if queue_length > 0:
+            log.warning(
+                "Backfill commits queue not drained",
+                extra=dict(
+                    queue_length=queue_length,
+                ),
+            )
+            return
+
+        commits = Commit.objects.all()
```
giovanni-guidini, line 43: Will this `all()` actually fetch ALL commits in the database? seems excessive, and commits are fairly complex objects...

I know you can define the query and the filters without actually fetching any data (and only do that at the end). I think it's worth doing that, if we're not doing so yet.

core/management/commands/backfill_commits.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # this stores the oldest commit id that has already been backfilled
+        commit_id = storage_redis.get("backfill_commits_id")
+        if commit_id:
+            commit_id = int(commit_id)
+            commits = commits.filter(id__lt=commit_id)
+
+        commits = commits.order_by("-id")[:batch_size]
```
giovanni-guidini, line 51: Can the batch size be added as a limit to the query? So we only fetch the commits we want from the Database.

Also if we can not fetch the entire commit, that would be ideal.

giovanni-guidini (commented):
Left some comments around getting the commit info to enqueue tasks. They are base on my assumptions of how Django ORM works, and I might be wrong 😅 but anyway

core/management/commands/backfill_commits.py
scott-codecov: This just returns a `QuerySet` which is a lazy representation of the actual SQL to be executed.  Below we do a `queryset[:batch_size]` which triggers the actual execution of the query and adds a `limit {batch_size}` to the SQL.

core/management/commands/backfill_commits.py
scott-codecov: See https://github.com/codecov/codecov-api/pull/102#discussion_r1316161249

core/management/commands/backfill_commits.py
scott-codecov: https://docs.djangoproject.com/en/4.2/topics/db/queries/#limiting-querysets

core/management/commands/backfill_commits.py
scott-codecov: > Also if we can not fetch the entire commit, that would be ideal.

Oh yeah, good idea - will just fetch the `id` and `commitid`

core/management/commands/backfill_commits.py
giovanni-guidini: ah perfect, perfect. Thanks for the docs too

giovanni-guidini (approved):
seal_of_approval

PR #106, opened 2023-09-06
Don't compile since source is available

trent-codecov:
### Purpose/Motivation
Don't compile

### What does this PR do?
Removes a bunch of things related to compilation. 

### Notes to Reviewer

### Legal Boilerplate


scott-codecov (approved):
This is really nice.  So much easier to run in enterprise mode now!

PR #115, opened 2023-09-06
Adjust donwload_url link

adrian-codecov:
### Purpose/Motivation
Adjust the download_url surfaced to the client.

### What does this PR do?
- Adjusts download_url + test related to it

### Legal Boilerplate


RulaKhaled (approved):
Looks good, just wondering if we can instead alter to https from the source build_absolute_uri itself? i can't really find the function anywhere

PR #128, opened 2023-09-12
feat: Filter flags by flags for pathContents

adrian-codecov:
### Purpose/Motivation
The client needs to be able to filter out files for pathcontents. This PR enables that by exposing filters to the client and the ability to reflect those files accordingly. 

#### Meat of the PR 
- Adds logic in `service/reports.py` to calculate sessions based on flags, and files belonging to specific sessions
- Adds logic to interface those changes to path contents

- Adds the ability to filter files by flags for path contents related changes
- Handles errors when selected flag leads to no result
- Adds tests

### Legal Boilerplate


services/path.py
```diff
@@ -165,6 +171,19 @@ def __init__(
...
             ]
 
+    @property
+    def files_accounting_flags(self) -> List[str]:
+        if not self.filter_flags:
+            return self.report.files
+        files = report_service.files_belonging_to_flags(
+            commit_report=self.report, flags=self.filter_flags
+        )
+        self._filter_commit_report()
```
adrian-codecov, line 181: Not sure what I think about this, thought?

graphql_api/types/inputs/path_contents_filters.graphql
```diff
@@ -2,9 +2,10 @@ input PathContentsFilters {
   searchValue: String
   displayType: PathContentDisplayType
   ordering: PathContentsOrdering
+  flags: [String]
```
scott-codecov, line 5: Maybe make this `[String!]` so someone can't pass `[None]`

services/path.py
scott-codecov: I would maybe just do this in the constructor so we don't have to rely on this method being called.

```python
self.report = report
if self.flags_filter:
    self.report = self.report.filter(flags=self.flags_filter)
```

services/path.py
```diff
@@ -165,6 +171,19 @@ def __init__(
                 if search_term.lower() in path.relative_path.lower()
             ]
 
+    @cached_property
+    def files_accounting_flags(self) -> List[str]:
```
scott-codecov, line 175: I think you could just call this `files` (which takes into account it's own state).  Outside callers don't necessarily care about the flag part (other than passing them into the constructor).

services/report.py
```diff
@@ -224,3 +224,41 @@ def build_files(commit_report: CommitReport) -> dict[str, ReportFileSummary]:
...
+        commit_report=commit_report, flags=flags
+    )
+    session_ids = list(sessions_for_specific_flags.keys())
+    files_in_specific_sessions = calculate_files_in_sessions(
+        commit_report=commit_report, session_ids=session_ids
+    )
+    return files_in_specific_sessions
+
+
+def calculate_sessions_with_specific_flags(
```
scott-codecov, line 240: Not a huge fan of including things like `calculate` or `compute` in function names since that's basically the function of a computer.  Maybe just `sessions_with_flags`?

scott-codecov (approved):
Looks good!  Just a few minor naming things you can feel free to ignore if you want

services/path.py
adrian-codecov: I considered this, but the issue is that all the logic in the `files_accounting_flags` relies on a `Report` and not a `FilteredReport`. I suppose what I need to do is change that logic to use internally self.report(which is a `FitleredReport` by that point).report_file instead. I'll play around that

services/report.py
adrian-codecov: Yeah that's fine, I remember seeing somewhere this as a "good naming convention" as this is describing a verb instead of a noun, where nouns should be more like @propertys (which I can change to too)

services/path.py
scott-codecov: Oh yeah, good point.  It's kinda strange to me that `Report` and `FilteredReport` don't offer the exact same API.

services/report.py
scott-codecov: Hmm, interesting.  I'd say if it's mutating state somewhere then a verb definitely makes sense.  If it's just returning data then maybe less important to make that distinction?

adrian-codecov: Yeaaah makes sense to me, I can adjust this accordingly 👌

PR #129, opened 2023-09-13
Filter file comparisons by flags

adrian-codecov:
### Purpose/Motivation
We need to filter out file comparisons by flags. This PR introduces that logic

### What does this PR do?
- Adjusts impacted files to filter out those without relevant flags
- Adds param to gql to add flags
- Adds tests
- Add readonly report to head_report in `Comparison` class

### Legal Boilerplate


services/comparison.py
```diff
@@ -701,7 +704,9 @@ def base_report(self):
     @cached_property
     def head_report(self):
         try:
-            report = report_service.build_report_from_commit(self.head_commit)
+            report = report_service.build_report_from_commit(
+                self.head_commit, report_class=ReadOnlyReport
```
adrian-codecov, line 708: Could this be unsafe?

compare/commands/compare/interactors/fetch_impacted_files.py
```diff
@@ -13,13 +22,28 @@ class ImpactedFileParameter(enum.Enum):
...
+    ):
         parameter = filters.get("ordering", {}).get("parameter")
         direction = filters.get("ordering", {}).get("direction")
         if parameter and direction:
             impacted_files = self.sort_impacted_files(
                 impacted_files, parameter, direction
             )
+        flags = filters.get("flags", [])
+        if flags and comparison:
+            head_commit_report = comparison.head_report
```
adrian-codecov, line 39: Brought the comparison all the way down here to only compute the head_report if the flags are provided

compare/commands/compare/compare.py
```diff
@@ -1,8 +1,20 @@
...
 from .interactors.fetch_impacted_files import FetchImpactedFiles
 
 
 class CompareCommands(BaseCommand):
-    def fetch_impacted_files(self, comparsion, filters):
-        return self.get_interactor(FetchImpactedFiles).execute(comparsion, filters)
+    def fetch_impacted_files(
+        self,
+        comparison_report: ComparisonReport,
+        comparison: Union[PullRequestComparison, Comparison],
```
scott-codecov, line 15: Just `Comparison` is fine here since `PullRequestComparison` inherits from `Comparison` and we're not using anything specific to `PullRequestComparison`.

compare/commands/compare/interactors/fetch_impacted_files.py
```diff
@@ -13,13 +22,28 @@ class ImpactedFileParameter(enum.Enum):
 
 
 class FetchImpactedFiles(BaseInteractor):
-    def _apply_filters(self, impacted_files, filters):
+    def _apply_filters(
+        self,
+        impacted_files: Optional[List[ImpactedFile]],
+        comparison: Union[PullRequestComparison, Comparison],
```
scott-codecov, line 28: Same here - can just be `Comparison`

compare/commands/compare/interactors/fetch_impacted_files.py
```diff
@@ -13,13 +22,28 @@ class ImpactedFileParameter(enum.Enum):
...
                 impacted_files, parameter, direction
             )
+        flags = filters.get("flags", [])
+        if flags and comparison:
+            head_commit_report = comparison.head_report
+            if set(flags) & set(head_commit_report.flags):
+                files = files_belonging_to_flags(
+                    commit_report=head_commit_report, flags=flags
+                )
+                impacted_files = list(
```
scott-codecov, line 44: This might read a little easier:

```python
impacted_files = [
    file
    for file in impacted_files
    if file.head_name in files
]
```

services/comparison.py
scott-codecov: I don't think the API does any report modification so this seems like a good move to me.

scott-codecov (approved):

PR #139, opened 2023-09-15
feat: Add authenticated view

scott-codecov:
### Purpose/Motivation

Gazebo isn't able to easily handle the 401 from `/internal/users` as a way to determine that the user is not authenticated.

### What does this PR do?

Adds a new endpoint `GET /internal/authenticated` which always returns a 200 and has a JSON response body of the form:

```
{"authenticated": true/false}
```


nicholas-codecov (approved):
Looks all good to me 👍

PR #141, opened 2023-09-18
feat: Add `trial_fired_by` to owner model 

RulaKhaled:
### Purpose/Motivation
to activate user post trial, we need to keep track of their ownerid 

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/406

### SQL migration 
```
BEGIN;
--
-- Add field trial_fired_by to owner
--
ALTER TABLE "owners" ADD COLUMN "trial_fired_by" integer NULL;
COMMIT;
```

### Notes to Reviewer
looks like terms condition `default ` wasn't migrated

### Legal Boilerplate


codecov_auth/migrations/0041_owner_trial_fired_by_alter_user_terms_agreement.py
```diff
@@ -0,0 +1,23 @@
...
+        ("codecov_auth", "0040_oktauser"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="owner",
+            name="trial_fired_by",
+            field=models.IntegerField(null=True),
+        ),
+        migrations.AlterField(
```
adrian-codecov, line 18: Huuuuh, what's this doing here 👀 is this part of your PR?

RulaKhaled: it shouldn't be, but it looks like the last migration applied to this field didn't capture "default":
`field=models.BooleanField(null=True),
`
it should be:
`
field=models.BooleanField(default=False, null=True),`

RulaKhaled: or i have no idea

RulaKhaled: synced w/@jerry it's safe not to include it

adrian-codecov (approved):

Issue #46, opened 2023-07-28
Allow API to be called on Reviewable


thomasrockhu-codecov:
Currently, when a service like Reviewable uses the API to make an (unauthenticated) call, it errors out, showing a CORS error. Current solution looks to be updating one of the Django middlewares to deal with `Access-Control-Allow-Origin:` and `Access-Control-Allow-Credentials`

cc: @scott-codecov

Smit-Parmar:
Can we have more details on this like how are we allowing API access to everyone is it any subscription based or public for all? If public then should we allow every origin? If not then how are we managing list of origin? Thanks
@thomasrockhu-codecov @scott-codecov

scott-codecov:
We're thinking of dynamically setting `Access-Control-Allow-Origin: <requesting origin>` (perhaps using https://docs.djangoproject.com/en/dev/ref/request-response/#django.http.HttpRequest.get_host).  This will allow anyone to make public CORS requests.

Most of the time we'll omit the `Access-Control-Allow-Credentials` header except when the request originates from the Codecov frontend (`app.codecov.io`) in which case we'll set `Access-Control-Allow-Credentials: true`.  This will allow only the Codecov frontend to utilize the authenticated session.

Apparently the CORS spec does not allow setting `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Credentials` together which is why we need the dynamic origin.

We're currently using https://github.com/adamchainz/django-cors-headers for setting the CORS headers.  Maybe this use case can be supported via the various signals that package uses otherwise it may require a custom Django middleware.  I'm not entirely sure w/o a little investigation.

trent-codecov:
I just pulled this issue into our next sprint. We will have SOME answer in <2 weeks (our sprints are 2 weeks).

PR #144, opened 2023-09-18
feat: Allow CORS requests from all origins, credentials whitelisted

scott-codecov:
Resolves https://github.com/codecov/codecov-api/issues/46

We'd like to allow CORS requests to public resources from all origins.  For Codecov whitelisted origins (like Gazebo) we'll allow authenticated requests.


matt-codecov (approved):
brushed up on CORS, the concept of the change is good but double-check the ways in which it differs from the upstream function:
- (nit) upstream uses `urlsplit` instead of `urlparse`
- upstream early-exits and sets no headers if CORS is not enabled, we basically bypass that
- upstream will set `ACCESS_CONTROL_ALLOW_CREDENTIALS` but *not* set `ACCESS_CONTROL_ALLOW_ORIGIN` if [this condition](https://github.com/adamchainz/django-cors-headers/blob/67f1f9214fc8be556ee6580c53f4036b8fa52abd/src/corsheaders/middleware.py#L148-L153) is met. we set both together unconditionally
- if the above condition is met, there are other `ACCESS_CONTROL_*` headers that are skipped

i don't think the differences will really impact behavior today, but it's the kind of thing that'll be silently/subtly wrong in (hypothetical) eighteen months when we start adding Django "signals" or whatever. i think you can eliminate them by:
- checking if CORS is enabled up top
- using `urlsplit` instead of `urlparse`
- skipping setting `ACCESS_CONTROL_ALLOW_ORIGIN` (upstream already sets it to the same value we do if it determines it should be set at all

approving as i'm sure you'll address appropriately

PR #146, opened 2023-09-18
ref: Read/Write agreements from user table only

JerrySentry:
### Purpose/Motivation

When updating term agreement will update only User columns (via saveTermsAgreement mutation)
Add retrieving term agreement from User columns (via /internal/user endpoint and GraphQL me endpoint)

Note: To be merged when https://github.com/codecov/codecov-api/pull/145 is merged and risky SQL is ran.

### Links to relevant tickets

### What does this PR do?

Release 3 of https://github.com/codecov/engineering-team/issues/390#issuecomment-1698094168

### Legal Boilerplate


codecov_auth/admin.py
```diff
@@ -133,7 +123,7 @@ class OwnerAdmin(AdminMixin, admin.ModelAdmin):
     search_fields = ("username__iexact",)
     actions = [impersonate_owner]
     autocomplete_fields = ("bot",)
-    inlines = [OrgUploadTokenInline, OwnerProfileInline]
+    inlines = [OrgUploadTokenInline]
```
adrian-codecov, line 126: Have you ensured this works appropriately in the django admin?

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -23,20 +23,9 @@ def validate(self, input: TermsAgreementInput):
...
-        owner_profile.terms_agreement = input.terms_agreement
-        owner_profile.terms_agreement_at = now
-        owner_profile.save()
-
-        # Store agreements in user table as well
-        user = User.objects.get(id=self.current_owner.user_id)
-        user.terms_agreement = input.terms_agreement
-        user.terms_agreement_at = now
-        user.save()
+        self.current_user.terms_agreement = input.terms_agreement
```
adrian-codecov, line 26: `self.current_user` here is of `User` instance right?

codecov_auth/models.py
```diff
@@ -71,8 +71,8 @@ class User(BaseCodecovModel):
     is_staff = models.BooleanField(null=True, default=False)
     is_superuser = models.BooleanField(null=True, default=False)
     external_id = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
-    terms_agreement = models.BooleanField(null=True, default=False)
-    terms_agreement_at = DateTimeWithoutTZField(null=True)
+    terms_agreement = models.BooleanField(null=True, default=None, blank=True)
```
adrian-codecov, line 74: We had this signal, https://github.com/codecov/codecov-api/blob/main/codecov_auth/signals.py, when an owner was created to create an `OwnerProfile` with `terms_agreement` set to false by default. I don't think we want to set this to None here, rather to `False` as None and False are treated different by the UI

adrian-codecov (commented):
Left some comments 👀

codecov_auth/admin.py
adrian-codecov: We confirmed it does

codecov_auth/commands/owner/interactors/save_terms_agreement.py
JerrySentry: Yes

codecov_auth/models.py
JerrySentry: Updated to default new users to have False value.

adrian-codecov (approved):

PR #147, opened 2023-09-19
fix: Remove duplicate values in owners.admins and fix adding dupes in…

JerrySentry:
### Purpose/Motivation

There is a bug when the membership list is loaded in the app by an admin of the org, their ownerid will be added to the admins column regardless if it was already there. On the admin page, it will show hundreds of the same admins for an org.

This fix will remove all duplicate ownerids in owners.admins column for every row. Should lead to improved DB disk usage and API query speed as well.

Note that the migration script runs the "risky" check, which means it doesn't run in CI when building/deploying the app in production. Need to run it manually by dev-ops after the release is deployed.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/193

### What does this PR do?

Fix bug in django
Provide a django migration script to remove existing duplicates

### Legal Boilerplate


codecov_auth/tests/test_migrations.py
```diff
@@ -3,106 +3,73 @@
 from utils.test_utils import TestMigrations
```
JerrySentry, line 3: The test being replaced here is expected, that test will be deleted in PR #154

adrian-codecov (approved):

PR #154, opened 2023-09-21
ref: drop ownerprofile terms agreement columns

JerrySentry:
### Purpose/Motivation

Remove two columns from ownerprofile table that is no longer used, current data in those columns are migrated to the new columns in the user table, though the data is not meaningful at the moment anyways until we turn on the agreements feature on the frontend.

### Links to relevant tickets

### What does this PR do?
- Drop two unused columns in the DB

### Legal Boilerplate


codecov_auth/tests/test_migrations.py
```diff
@@ -1,108 +0,0 @@
-from datetime import datetime
```
JerrySentry, line 1: Remove this test for previous migration since these columns are deleted in this change anyways.

graphql_api/types/me/me.py
```diff
@@ -104,6 +104,8 @@ def resolve_tracking_data(current_user, _, **kwargs):
 @me_bindable.field("termsAgreement")
 @sync_to_async
 def resolve_terms_agreement(current_owner: Owner, _, **kwargs) -> Optional[bool]:
+    if current_owner.user is None:
```
JerrySentry, line 107: Apparently it is possible for user to be null through impersonation?

adrian-codecov (approved):

PR #159, opened 2023-09-25
chore: Allow GraphQL introspection + playground in all envs

scott-codecov:
Enables GraphQL schema introspection in all envs


nicholas-codecov (approved):
Send it

PR #169, opened 2023-10-03
ATS Docker Action

trent-codecov:
### Purpose/Motivation
Fixes ATS step. Uses ATS Docker action.

### Legal Boilerplate


Makefile
```diff
@@ -200,7 +200,7 @@ test_env.upload:
 	docker-compose -f docker-compose-test.yml exec api make test_env.container_upload CODECOV_UPLOAD_TOKEN=${CODECOV_UPLOAD_TOKEN} CODECOV_URL=${CODECOV_URL}
 
 test_env.container_upload:
-	codecovcli -u ${CODECOV_URL} do-upload --flag unit-latest-uploader --flag unit  \
+	codecovcli -u ${CODECOV_URL} upload-process --flag unit-latest-uploader --flag unit  \
```
thomasrockhu-codecov, line 203: did we change the name of this?

trent-codecov: do-upload is a single upload step.
upload-process is a shortcut for create-commit, create-report, do-upload.

Since we aren't doing parallel testing and have just 1 upload (sort of) for API, I am using the shortcut.

scott-codecov (approved):

PR #170, opened 2023-10-03
chore: add overview ats metrics

giovanni-guidini:
Adding some general, high-level metrics to static analysis and label analysis.
We still want more detailed metrics with higher cardinality, but those will probably be in a different backend.

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/labelanalysis/serializers.py
+++ b/labelanalysis/serializers.py
@@ -1,4 +1,5 @@
 from rest_framework import exceptions, serializers
+from shared.metrics import metrics
 
 from core.models import Commit
 from labelanalysis.models import (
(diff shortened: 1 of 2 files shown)


labelanalysis/serializers.py
```diff
@@ -22,21 +23,25 @@ def to_internal_value(self, commit_sha):
...
             return commit
         if not self.accepts_fallback:
+            metrics.incr("label_analysis_request.errors.static_analysis_not_found")
             raise serializers.ValidationError("No static analysis found")
         attempted_commits = []
         for _ in range(10):
             attempted_commits.append(commit.commitid)
             commit = commit.parent_commit
             if commit is None:
+                metrics.incr("label_analysis_request.errors.static_analysis_not_found")
```
thomasrockhu-codecov, line 38: should we differentiate this from the previous metric?

giovanni-guidini: I don't think so... it's the same error, but in the 2nd case we try to fallback.
But they track the same thing.

Same reasoning for the metric below this one.

joseph-sentry (approved):

PR #172, opened 2023-10-03
feat: Allow upload requests originating from Shelter to set storage path

scott-codecov:
### Purpose/Motivation

Shelter will be buffering uploads on a queue and then calling the API async.  Since Shelter won't know about the report and upload's external ID we need to construct our own storage path inside Shelter.

### Links to relevant tickets

TODO: link to Shelter PR

### What does this PR do?

This PR introduces a new config option which is a secret that will be shared between API and Shelter.  Shelter will add this to some API requests in a custom header: `X-Shelter-Token`

When this token is present in the upload handler then we'll allow the storage path to be set explicitly instead of generating that path in the API.


upload/views/base.py
```diff
@@ -10,7 +11,19 @@
...
+class ShelterMixin:
+    def is_shelter_request(self) -> bool:
+        """
+        Returns true when the incoming request originated from a Shelter.
+        Shelter adds an `X-Shelter-Token` header which contains a shared secret.
+        Use of that shared secret allows certain priviledged functionality that normal
+        uploads cannot access.
+        """
+        shelter_token = self.request.META.get("HTTP_X_SHELTER_TOKEN")
+        return shelter_token and shelter_token == settings.SHELTER_SHARED_SECRET
```
adrian-codecov, line 23: Makes sense. Is this a go-to approach when trying to identify specific requests? (for my own learning heh)

upload/tests/views/test_uploads.py
```diff
@@ -288,6 +291,60 @@ def test_uploads_post(mock_metrics, db, mocker, mock_redis):
...
+            "storage_path": "shelter/test/path.txt",
+        },
+        headers={
+            "X-Shelter-Token": "shelter-shared-secret",
+        },
+    )
+    upload = ReportSession.objects.filter(
+        report_id=commit_report.id, upload_extras={"format_version": "v1"}
+    ).first()
+    assert response.status_code == 201
```
adrian-codecov, line 342: I'm confused: the other code I see is being gated by `not self.is_shelter_request()`, but this endpoint works. Is this intended or did I miss something? As far as I understand this, the existing logic should only work if it is not a shelter request

scott-codecov: The logic within `not self.is_shelter_request()` is the default behavior of generating the storage path in the API.  Only requests from Shelter are allowed to pass in their own storage path.

upload/views/base.py
adrian-codecov: Makes sense, I recall doing some identification through headers, cool!

adrian-codecov (approved):

PR #174, opened 2023-10-03
fix: Address various security scanning vulerabilities

scott-codecov:
### Purpose/Motivation

Addresses multiple vulnerabilities detected by GH code scanning alerts.

### Links to relevant tickets

https://github.com/codecov/internal-issues/issues/53
https://github.com/codecov/internal-issues/issues/54
https://github.com/codecov/internal-issues/issues/55
https://github.com/codecov/internal-issues/issues/70
https://github.com/codecov/internal-issues/issues/71

Fixes https://github.com/codecov/codecov-api/security/code-scanning/5
Fixes https://github.com/codecov/codecov-api/security/code-scanning/6
Fixes https://github.com/codecov/codecov-api/security/code-scanning/7

These tickets boil down to 2 vulnerabilities described below. 

### What does this PR do?

* Encrypts the OAuth token data before saving it in a cookie
  - for BitBucket and BitBucket server
  - also signs the cookie for good measure

* Addresses a potential DoS on a path regex that originates from user input
  - completely changed the logic related to this and removed the regex altogether
  - @adrian-codecov Can you verify that this still makes sense?

### DIFF
--- a/codecov_auth/views/bitbucket.py
+++ b/codecov_auth/views/bitbucket.py
@@ -12,6 +12,7 @@ from shared.torngit import Bitbucket
 from shared.torngit.exceptions import TorngitServerFailureError
 
 from codecov_auth.views.base import LoginMixin
+from utils.encryption import encryptor
 
 log = logging.getLogger(__name__)
 
@@ -63,8 +64,10 @@ class BitbucketLoginView(View, LoginMixin):
             + b"|"
             + base64.b64encode(oauth_token_secret.encode())
         ).decode()
-        response.set_cookie(
-            "_oauth_request_token", data, domain=settings.COOKIES_DOMAIN
+        response.set_signed_cookie(
+            "_oauth_request_token",
+            encryptor.encode(data).decode(),
+            domain=settings.COOKIES_DOMAIN,
         )
         self.store_to_cookie_utm_tags(response)
         return response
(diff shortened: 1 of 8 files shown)


services/path.py
```diff
@@ -121,25 +121,19 @@ def is_subpath(full_path: str, subpath: str):
     return full_path.startswith(f"{subpath}/") or full_path == subpath
 
 
-# Regex that separates string into 2 groups, taken from https://stackoverflow.com/a/33021907
-def calculate_path_file_and_dir_groups(path: str) -> tuple[str, str]:
-    result = re.search(r"((?:[^/]*/)*)(.*)", path)
-    return result.groups()
-
-
-def is_file(path) -> bool:
```
scott-codecov, line 130: @adrian-codecov I'm not sure this logic is correct.  Couldn't you have a file w/o a `.` in it?  Or a file with a leading `.`?

I think the only way to differentiate file vs. directory path is by checking the report.

codecov_auth/views/bitbucket.py
```diff
@@ -63,8 +64,10 @@ def redirect_to_bitbucket_step(self, request):
             + b"|"
             + base64.b64encode(oauth_token_secret.encode())
         ).decode()
-        response.set_cookie(
-            "_oauth_request_token", data, domain=settings.COOKIES_DOMAIN
+        response.set_signed_cookie(
+            "_oauth_request_token",
+            encryptor.encode(data).decode(),
```
matt-codecov, line 69: so i am guessing `encryptor.encode()` encrypts and `.decode()` is to convert from bytes to str, right? awkward names

matt-codecov (approved):
lgtm, but keep an eye on deploying in case we learn the hard way why that "filename must have dot but not at the start" logic is in there

codecov_auth/views/bitbucket.py
scott-codecov: Yeah that's right.  I agree it's confusing.  `encryptor.encrypt(data).decode()` would at least make a little more sense.

PR #176, opened 2023-10-04
Add ServiceMiddleware to handle short service names in API

joseph-sentry:
### Purpose/Motivation
To allow users to use short service names such as `gh` and `bb` in their calls to the API when specifying the service parameter, for example we want: `https://api.codecov.io/api/v2/gh/getsentry/repos/sentry` and `https://api.codecov.io/api/v2/github/getsentry/repos/sentry` to both work.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/479

### What does this PR do?
- Creates the ServiceMiddleware
  - this checks the kwargs passed to a view, and if the kwargs contain a `service` parameter, it checks if the parameter is a short service name
  - if it is, it replaces it with the long service name
- Adds the ServiceMiddleware to the Django app


codecov/settings_base.py
```diff
@@ -57,6 +57,7 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "api.shared.middleware.ServiceMiddleware",
```
scott-codecov, line 60: Since this applies to all views (I think) maybe move this into `core.middleware`?

scott-codecov (approved):

PR #179, opened 2023-10-05
feat: Notify that commit is currently processing

scott-codecov:
### Purpose/Motivation

Send an early notification that an upload is being processed.
Depends on https://github.com/codecov/worker/pull/140

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/553

### What does this PR do?

Enqueues a notify task as soon as we receive an upload (but before it's processed).  This notification will issue a PR comment that says something like "We're processing your upload".


JerrySentry (approved):

services/task/task.py
```diff
@@ -147,17 +148,36 @@ def upload(
...
             "app.tasks.notify.Notify",
             kwargs=dict(
                 repoid=repoid,
                 commitid=commitid,
                 current_yaml=current_yaml,
                 empty_upload=empty_upload,
             ),
+        )
+
+    def notify(self, repoid, commitid, current_yaml=None, empty_upload=None):
```
dana-yaish, line 178: since now we're using different kinds of notifications here, can we rename the empty_upload to be something more descriptive? like notification_type for example?

dana-yaish (approved):
LGTM, I like the idea that we're checking the existence of previous notifications. Left some minor notes but this is looking good

services/task/task.py
```diff
@@ -34,7 +34,7 @@
 
 
 class TaskService(object):
-    def _create_signature(self, name, args=None, kwargs=None):
+    def _create_signature(self, name, args=None, kwargs=None, immutable=False):
```
dana-yaish, line 37: this is for my own understanding, what's the immutable for?

services/task/task.py
scott-codecov: Yeah I think that would make sense.  I can update it in the worker too but will have to support both `empty_upload` and `notification_type` for a little bit so as to preserve backward compatibility.

services/task/task.py
scott-codecov: Yeah, I had to look this up in the Celery docs.  When we use Celery's `chain` or `link` then the results of the previous task are passed as arguments to the next task in the chain (this is the default and is when `immutable=False`).  I left a small comment about this where I pass `immutable=True` and the goal was to prevent the results of the notify task from being passed as args to the upload task.

upload/helpers.py
```diff
@@ -637,16 +638,39 @@ def dispatch_upload_task(task_arguments, repository, redis):
...
         report_code=task_arguments.get("report_code"),
-        countdown=max(
-            countdown, int(get_config("setup", "upload_processing_delay") or 0)
-        ),
+        # this prevents the results of the notify task (below) from being
+        # passed as args to this upload task
+        immutable=True,
     )
 
+    notified = CommitNotification.objects.filter(commit__commitid=commitid).exists()
```
scott-codecov, line 652: I'm realizing we probably need to store this per-PR since when we make another commit on the same branch we don't necessarily want to show the "processing" comment again. 🤔

upload/helpers.py
```diff
@@ -649,7 +649,20 @@ def dispatch_upload_task(task_arguments, repository, redis):
...
+    # correct but works for now since the early notify is only for PR comments.
+
+    notified = False
+    commit = Commit.objects.filter(commitid=commitid).first()
+    if commit and commit.pullid is not None:
+        pull = Pull.objects.filter(
+            repository_id=commit.repository_id, pullid=commit.pullid
+        ).first()
+        if pull and pull.commentid:
+            notified = True
```
scott-codecov, line 664: @dana-yaish What do you think about this approach?  I had been checking `CommitNotification` previously but that resulted in us showing the "processing" comment for a brief period each time a new commit was pushed to a PR.

giovanni-guidini: Isn't that more accurate though? Every time I push a new commit in a PR I expect the CI to run again and update me of the status of the new commit.

If we had 2 slightly different versions of "we are processing the upload" and "you updated this PR and we are processing the upload" that seems the most accurate to me.
It somehow seems better than the comment with the red banner saying "the head of this PR doesn't match these calculations, please upload new coverage"

PR #181, opened 2023-10-06
feat: 411 add available paid plans resolver

adrian-codecov:
### Purpose/Motivation
We want to surface the plans available to a customer based on some parameters. This is adding a resolver to expose that information. This resolver mimics a function shown in via a REST endpoint, with the intent to transition to gql only.

This is not the final version of this change, as I want some input on how to keep parity with this very similar function, https://github.com/codecov/codecov-api/blob/main/billing/helpers.py#L23. There's two main differences with this and the newly added resolver: (Edit: these have been addressed)
- Original fn doesn't have logic to add Lite plans if we want to use those (can be added)
- Offers just the basic plan rather than the free plan representation, which I'm unsure if I'm making the right assumption
- Original fn only adds sentry plans to sentry users and not pro plans

### What does this PR do?
- Adds `availablePlans` resolver to an owner
- Adds tests
- Adds fn in plan service to get the available_plans
- Adjusted existing `available_plans` function to include lite plans

### Legal Boilerplate


billing/helpers.py
```diff
@@ -33,5 +34,15 @@ def available_plans(owner: Optional[Owner]) -> List[dict]:
...
         plans += list(SENTRY_PAID_USER_PLAN_REPRESENTATIONS.values())
 
+    if owner:
+        plan_service = PlanService(current_org=owner)
+
+        if (
+            plan_service.trial_status == TrialStatus.ONGOING.value
+            or plan_service.trial_status == TrialStatus.EXPIRED.value
+            or plan_service.plan_name in LITE_PLAN_REPRESENTATIONS
+        ) and plan_service.plan_user_count <= 10:
```
JerrySentry, line 44: It would be better to put the `10` into constants file

graphql_api/types/owner/owner.graphql
```diff
@@ -20,6 +20,7 @@ type Owner {
   ownerid: Int!
   plan: Plan
   pretrialPlan: PlanRepresentation
+  availablePlans: [PlanRepresentation!]!
```
JerrySentry, line 23: What's the difference between `[PlanRepresentation!]!` and `[PlanRepresentation]!`?

plan/service.py
```diff
@@ -98,6 +100,29 @@ def monthly_uploads_limit(self) -> Optional[int]:
...
+            available_plans += SENTRY_PAID_USER_PLAN_REPRESENTATIONS.values()
+        else:
+            available_plans += PR_AUTHOR_PAID_USER_PLAN_REPRESENTATIONS.values()
+
+        # If you're trialing or have trialed and <= 10 users, or belong to the lite plan
+        if (
+            self.trial_status == TrialStatus.ONGOING.value
+            or self.trial_status == TrialStatus.EXPIRED.value
+            or self.plan_name in LITE_PLAN_REPRESENTATIONS
+        ) and self.plan_user_count <= 10:
```
JerrySentry, line 121: same here

PR #183, opened 2023-10-06
ref: Add new resolver for ImpactedFiles

JerrySentry:
### Purpose/Motivation

Make a new resolver for ImpactedFiles with a slight change in logic where if there are no coverage with list of flags provided for filtering it will return an error instead of show all files.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/641

### What does this PR do?
Flags refactoring

### Legal Boilerplate


graphql_api/types/comparison/comparison.py
```diff
@@ -45,6 +46,25 @@ def resolve_impacted_files(
...
+@comparison_bindable.field("impactedFilesNew")
+@convert_kwargs_to_snake_case
+@sync_to_async
+def resolve_impacted_files_new(
+    comparison_report: ComparisonReport, info, filters=None
+) -> List[ImpactedFile]:
+    command: CompareCommands = info.context["executor"].get_command("compare")
+    comparison: Comparison = info.context.get("comparison", None)
+
+    if filters:
```
adrian-codecov, line 58: I'd also add `if filters and comparison` cause there's some scenarios where the comparison isn't loaded into the context

graphql_api/types/comparison/comparison.graphql
```diff
@@ -14,6 +14,7 @@ type Comparison {
   hasDifferentNumberOfHeadAndBaseReports: Boolean!
   flagComparisonsCount: Int!
   componentComparisonsCount: Int!
+  impactedFilesNew(filters: ImpactedFilesFilters): ImpactedFilesResult!
```
adrian-codecov, line 17: this would be `impactedFilesNew(filters: ImpactedFilesFilters): [ImpactedFilesResult!]` since it's returning a list. 

Idk if this is in scope for this PR but we might want to also create a new version of a singular impacted file (see `impactedFile`)

graphql_api/tests/test_impacted_file_new.py
```diff
@@ -0,0 +1,846 @@
...
+
+        variables = {
+            "org": self.org.username,
+            "repo": self.repo.name,
+            "commit": self.commit.commitid,
+            "filters": {"flags": ["flag1"]},
+        }
+
+        data = self.gql_request(query_impacted_files, variables=variables)
+        assert (
```
adrian-codecov, line 336: Lines 328 onwards feels like another test, so like renaming this to "unknown flags" and the next one to "successful test", and assert exactly the response we get

graphql_api/tests/test_impacted_file_new.py
```diff
@@ -0,0 +1,846 @@
...
+            "owner": {
+                "repository": {
+                    "pull": {
+                        "compareWithBase": {
+                            "state": "pending",
+                            "impactedFile": {
+                                "headName": "fileB",
+                                "baseName": "fileB",
+                                "hashedPath": "eea3f37743bfd3409bec556ab26d4698",
+                                "baseCoverage": {"percentCovered": None},
```
adrian-codecov, line 518: Shouldn't be not None? Since the head/base reports have hits/misses

JerrySentry: This test doesn't have the percent covered setup so it should be as expected.

graphql_api/tests/test_impacted_file_new.py
```diff
@@ -0,0 +1,888 @@
...
+    @patch("services.comparison.Comparison.git_comparison")
+    def test_filtering_with_unknown_flags(
+        self, git_comparison_mock, read_file, build_report_from_commit
+    ):
+        git_comparison_mock.return_value = None
+        build_report_from_commit.return_value = sample_report()
+        read_file.return_value = mock_data_from_archive
+
+        query_impacted_files = """
+        query ImpactedFilesNew(
```
adrian-codecov, line 342: Optional: if the query is the same bw this and the other filter file, maybe add it to the setup to avoid DRY

adrian-codecov (approved):

PR #185, opened 2023-10-11
feat: Dump GraphQL schema in pre-commit hook

scott-codecov:
### Purpose/Motivation

We'd like a clear way to determine when the GraphQL schema is changed.  By making sure the full schema is part of every commit it will become clear in every commit what the schema diff is.

### Links to relevant tickets

N/A

### What does this PR do?

Adds a new Django management command to dump the GraphQL schema into a text file (`graphql_api/schema.graphql`).  This command is run as part of the pre-commit.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,3 +8,10 @@ repos:
     pass_filenames: false
     require_serial: true
     language: system
+  - id: graphql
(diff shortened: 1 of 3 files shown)


matt-codecov (approved):
schema file is 20kb, git should handle that fine

PR #187, opened 2023-10-12
ref: Accept service less requests gq

RulaKhaled:
### Purpose/Motivation
To be able to use TOS with service less requests, we need to update the mutation (graphql endpoints) to accept no service, and raise a proper exception otherwise. 

### What does this PR do?
- Add new MissingService exception 
- Allow non-service based TOS mutation 
- Add tests 

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/tests/test_owner.py
+++ b/graphql_api/tests/test_owner.py
@@ -2,6 +2,7 @@ import asyncio
 from datetime import timedelta
 from unittest.mock import patch
 
+import pytest
 from django.test import TransactionTestCase
 from django.utils import timezone
 from freezegun import freeze_time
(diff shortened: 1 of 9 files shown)


graphql_api/tests/test_owner.py
```diff
@@ -548,3 +549,23 @@ def test_owner_available_plans(self):
...
+        }
+        """ % (
+            current_org.username
+        )
+
+        try:
+            self.gql_request(query, provider="", owner=current_org)
+        except MissingService as e:
+            assert str(e) == "Missing service"
+            raise
```
scott-codecov, line 571: If this test is passing then maybe this error is not being rased as expected?

RulaKhaled: why?

scott-codecov: I think the test would error if this exception was indeed being raised

RulaKhaled: are you talking about raise? it is my understanding that the error will only be raised if the assert condition is not met

PR #202, opened 2023-10-19
feat: add prometheus instrumentation to requests, db accesses, and models

matt-codecov:
uses [`django-prometheus`](https://github.com/korfuri/django-prometheus) to instrument a bunch of metrics in django. there are quite a lot

if django is running on `localhost:8000`, metrics will be served at `localhost:8001/monitoring/metrics`. ~it will listen on a range from 8001-8020 for when `gunicorn` is running multiple worker processes that collect metrics separately. so i guess poll the whole range, though i think today prod only runs two workers~

when `gunicorn` runs django in multiple worker processes, we need to use prometheus's multiprocess mode. dev just runs a single worker, but prod runs 2 and enterprise runs a configurable number. so:
- in prod, assign a default value for the `$PROMETHEUS_MULTIPROC_DIR` env var (`$HOME/.prometheus`)
- in enterprise, assign a default value for the `$PROMETHEUS_MULTIPROC_DIR` env var _if using more than 1 worker process` (`$HOME/.prometheus`)
- in dev, prod, and enterprise, clear/recreate the `$PROMETHEUS_MULTIPROC_DIR` directory on startup if it's set

if `$PROMETHEUS_MULTIPROC_DIR` is set, `django-prometheus` will automatically use the multiprocessing registry and no extra configuration is needed.

some of the key metrics that this generates:
- API hits, labelled by view
- API responses, labeled by view, http method, and status code
- counters for insert/update/delete for models including commit, upload, report, repository...

### Legal Boilerplate


codecov/urls.py
```diff
@@ -25,4 +25,5 @@
     path("staticanalysis/", include("staticanalysis.urls")),
     path("labels/", include("labelanalysis.urls")),
     re_path(r"^redirect_app", views.redirect_app),
+    path("monitoring/", include("django_prometheus.urls")),
```
scott-codecov, line 28: Does this become publicly available?  Or is `django_prometheus` doing something internal such that this URL is only available on port 8001?

dev.sh
```diff
@@ -3,12 +3,18 @@
...
+    mkdir "$PROMETHEUS_MULTIPROC_DIR"
+fi
+
 export PYTHONWARNINGS=always
 suffix=""
 if [[ "$STATSD_HOST" ]]; then
   suffix="--statsd-host ${STATSD_HOST}:${STATSD_PORT}"
 fi
-if [[ "$RUN_ENV" == "enterprise" ]]; then
+if [[ "$RUN_ENV" == "enterprise" ]] || [[ "$RUN_ENV" == "DEV" ]]; then
```
scott-codecov, line 17: Good idea 👍

scott-codecov (approved):

codecov/urls.py
matt-codecov: good question! talking about those details with @rfox-codecov

as-is it would be public, and if we want to avoid leaking the metrics we would want to add a null-route rule to the URL mapping we do in kubernetes in production. i think prometheus requests the host directly, bypassing the null-route

if we don't like that solution, a couple other options which would require changes in this PR:
- create HTTP Basic Auth credentials for prometheus to supply and wrap this view to perform basic auth
- because prometheus requests the host directly, we could branch on `request.META["Host"]`. if it's an internal request we serve it, else 404

matt-codecov: i think we're going for the first which means no code change, but i'll add a comment disclaiming that without some higher-level configuration this route will be public

PR #203, opened 2023-10-20
ref: Handle current_user in base

RulaKhaled:
### Purpose/Motivation
passing current_user to use instead of relying on owner

### Links to relevant tickets
TBD

### What does this PR do?
- Handled current user in executor 
- Default to none user, so we don't go over 1000 tests :D 
- Refactor save terms interactor to handle user specific fields  

### Notes to Reviewer
..

### Legal Boilerplate

### DIFF
--- a/codecov/commands/base.py
+++ b/codecov/commands/base.py
@@ -42,5 +42,5 @@ class BaseInteractor:
         if not self.service and self.requires_service:
             raise MissingService()
 
-        if self.current_owner and not self.current_user.is_authenticated:
-            self.current_user = self.current_owner
+        if self.current_owner:
+            self.current_user = self.current_owner.user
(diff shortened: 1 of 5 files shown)


scott-codecov (approved):

codecov/commands/base.py
```diff
@@ -29,13 +34,13 @@ def get_command(self, namespace):
...
         self.service = service
 
         if not self.service and self.requires_service:
             raise MissingService()
 
-        self.current_user = AnonymousUser()
-        if self.current_owner:
-            self.current_user = self.current_owner.user
+        if self.current_owner and not self.current_user.is_authenticated:
+            self.current_user = self.current_owner
```
scott-codecov, line 46: This is mixing types.  I think this could remain as:
```python
if self.current_owner:
    self.current_user = self.current_owner.user
```

scott-codecov (changes requested):

codecov/commands/base.py
RulaKhaled: oh sorry lost with the new changes

scott-codecov (approved):

PR #208, opened 2023-10-25
fix: adjust existing code to fit team tier requirements

adrian-codecov:
### Purpose/Motivation
I'm adjusting some logic to better fit the team tier initiative.

### What does this PR do?
- Removes team tier related logic from the `available_plans` REST function and its tests
  - This is because the endpoint used the request user and not the owner, which is needed. Also a good excuse to use gql instead
- Provide the plan_service within gql context as it's needed in the benefits child resolver, which I didn't catch initially
- Expose `planUserCount` resolver
- Modify benefits list to tentative final values

### Legal Boilerplate


plan/constants.py
```diff
@@ -281,8 +282,9 @@ class PlanData:
         base_unit_price=PlanPrice.TEAM_MONTHLY.value,
         benefits=[
             "Up to 10 users",
-            "Unlimited public repositories",
-            "Unlimited private repositories",
+            "Unlimited repositories",
+            "2500 repositories",
+            "Patch coverage analysis",
```
JerrySentry, line 287: Probably a design/FE thing, but saying "Patch coverage analysis" here might give the assumption that other plans don't have patch coverage, maybe we should explicitly list all the features that each plan has. Or on the team plan say "No project coverage" or something

JerrySentry (approved):

plan/constants.py
adrian-codecov: Yeaah I see what you're saying, I can see the list getting exhaustive if we write all the features, but I can see the confusion of thinking this is the only plan that features patch coverage and not others, while that's not true. We can certainly followup on copy changes later if these are not final, but I'll rope @codecovdesign and @aj-codecov here for input

PR #212, opened 2023-10-25
feat: Add comparison endpoint for impacted files

JerrySentry:
### Purpose/Motivation
Adding two new endpoints to public v2 API
1. Ability to get comparisons with only the impacted files between the two commits (this similar to GraphQL API Comparison type)
2. Ability to get the Segment of diff for a particular file (this is similar to GraphQL API Segments type)

Overall these endpoints improve response times compared to the two equivalent public v2 endpoints since data retrieved are precomputed by worker and payload is much less since not _all_ files in the repo is returned in the response.

This also refactors some code with the GraphQL data loader to share the same query.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/654

### Legal Boilerplate


api/shared/compare/serializers.py
```diff
@@ -76,3 +90,172 @@ class FlagComparisonSerializer(serializers.Serializer):
...
+        # then will fall back to retrieving and generating all files on the fly
+        if not commit_comparison:
+            try:
+                new_comparison = CommitComparison(
+                    base_commit=comparison.base_commit.commitid,
+                    compare_commit=comparison.head_commit.commitid,
+                )
+                new_comparison.save()
+                TaskService().compute_comparison(new_comparison.pk)
+            except Exception as e:
```
scott-codecov, line 251: Avoid catchall exceptions like this if possible since it can easily hide bugs.  Is there a specific error that you expect to happen in this case?

api/shared/compare/serializers.py
```diff
@@ -76,3 +90,172 @@ class FlagComparisonSerializer(serializers.Serializer):
...
+
+
+class ImpactedFilesComparisonSerializer(ComparisonSerializer):
+
+    files = serializers.SerializerMethodField()
+
+    def get_files(self, comparison: Comparison) -> List[dict]:
+        comparison_table = CommitComparison._meta.db_table
+        commit_table = Commit._meta.db_table
+        commit_comparison = CommitComparison.objects.raw(
```
scott-codecov, line 220: Since this is a sort of complicated raw SQL query and could be error prone if updated maybe we pull this out somewhere and reuse it both in the dataloader and here.

api/shared/compare/serializers.py
```diff
@@ -76,3 +90,172 @@ class FlagComparisonSerializer(serializers.Serializer):
...
+    def get_misses_count(self, impacted_file: ImpactedFile) -> serializers.IntegerField:
+        return impacted_file.misses_count
+
+    def get_hashed_path(self, impacted_file: ImpactedFile) -> serializers.CharField:
+        path = impacted_file.head_name
+        encoded_path = path.encode()
+        md5_path = hashlib.md5(encoded_path)
+        return md5_path.hexdigest()
+
+    def get_segments(
```
scott-codecov, line 201: One note about the GraphQL API: we don't query for segments by default when querying the impacted files b/c there can be a lot of segments and we're fetching the source code for each file from GitHub.  Might be worth considering a similar design for this public API.  i.e. return a list of impacted files w/ metadata about them and then a separate endpoint where you can fetch the segments for a particular file

api/shared/compare/serializers.py
```diff
@@ -76,3 +90,172 @@ class FlagComparisonSerializer(serializers.Serializer):
...
+
+    def get_change_coverage(
+        self, impacted_file: ImpactedFile
+    ) -> serializers.FloatField:
+        return impacted_file.change_coverage
+
+    def get_misses_count(self, impacted_file: ImpactedFile) -> serializers.IntegerField:
+        return impacted_file.misses_count
+
+    def get_hashed_path(self, impacted_file: ImpactedFile) -> serializers.CharField:
```
scott-codecov, line 195: This might be unnecessary - I think this was for some niche use case in Gazebo.  Not sure if it would be generally useful.

api/shared/compare/serializers.py
JerrySentry: Moved to `services.CommitComparisonService`, both public and GQL APIs are getting the query from there.

api/shared/compare/serializers.py
JerrySentry: 👍 created a new endpoint: `/api/v2/{service}/{owner_username}/repos/{repo_name}/compare/segments/{file_path}` that fetches for a specific file. Removed getting segments for each impacted file in the impacted_files endpoint

api/shared/compare/mixins.py
```diff
@@ -100,3 +142,37 @@ def flags(self, request, *args, **kwargs):
             for flag_name in comparison.non_carried_forward_flags
         ]
         return Response(FlagComparisonSerializer(flags, many=True).data)
+
+    @action(detail=False, methods=["get"])
+    @torngit_safe
+    def impacted_files(self, request, *args, **kwargs):
+        comparison = self.get_object()
+
+        return Response(
```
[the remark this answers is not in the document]
scott-codecov: Alternatively maybe the serializer includes the state and we update the docs (the docstring comment in the child view) to explain that this state might change after the initial request.

api/shared/compare/mixins.py
```diff
@@ -100,3 +142,37 @@ def flags(self, request, *args, **kwargs):
...
+        )
+
+    @action(detail=False, methods=["get"])
+    @torngit_safe
+    def segments(self, request, *args, **kwargs):
+        file_path = file_path = kwargs.get("file_path")
+        comparison = self.get_object()
+        commit_comparison = self.get_or_create_commit_comparison(comparison)
+
+        if commit_comparison:
```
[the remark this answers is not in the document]
JerrySentry: Good catch!

api/shared/compare/mixins.py
```diff
@@ -100,3 +142,37 @@ def flags(self, request, *args, **kwargs):
...
         return Response(FlagComparisonSerializer(flags, many=True).data)
+
+    @action(detail=False, methods=["get"])
+    @torngit_safe
+    def impacted_files(self, request, *args, **kwargs):
+        comparison = self.get_object()
+
+        return Response(
+            ImpactedFilesComparisonSerializer(
+                comparison,
```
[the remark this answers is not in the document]
JerrySentry: The `comparison` object is still used in the `ComparisonSerializer` for displaying the head/base coverage and some other things.

api/shared/compare/mixins.py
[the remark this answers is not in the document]
JerrySentry: I see, right now if the comparison is still being computed it would return an empty object for the impacted files. I think the most consistent thing to do during pending state is to compute on-the-fly the files similar to what's already happening when the `CommitComparison` object doesn't exist yet.

scott-codecov (approved):

PR #218, opened 2023-10-31
Add test result ingestion in the API

joseph-sentry:
### Purpose/Motivation
This is being done to support failed test ingestion in the API.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/664

Depends on: https://github.com/codecov/worker/pull/197

### What does this PR do?
- Create Test and TestInstance models
- Create test_results endpoint


reports/models.py
```diff
@@ -190,3 +191,33 @@ class UploadLevelTotals(AbstractTotals):
...
+class TestSession(BaseCodecovModel):
+    report_session = models.OneToOneField(ReportSession, on_delete=models.CASCADE)
+
+    class Meta:
+        db_table = "reports_testsessions"
+
+
+class TestRun(BaseCodecovModel):
+    outcome = models.BooleanField()
+    duration = models.FloatField()  # number of seconds
```
giovanni-guidini, line 215: if you call the field `duration_seconds` you don't need the comment and it's easier to know that it's seconds when using the field

barnett-yuxiang (commented):
RFC

reports/models.py
```diff
@@ -216,3 +217,46 @@ class UploadLevelTotals(AbstractTotals):
...
+class TestInstance(BaseCodecovModel):
+    test = models.ForeignKey(
+        "Test",
+        db_column="test_id",
+        related_name="testinstances",
+        on_delete=models.CASCADE,
+    )
+    duration_seconds = models.FloatField()
+    outcome = models.IntegerField()
+    report = models.ForeignKey(
```
adrian-codecov, line 252: TODO: rename this to `upload`

reports/models.py
```diff
@@ -216,3 +217,46 @@ class UploadLevelTotals(AbstractTotals):
...
+        "ReportSession",
+        db_column="upload_id",
+        related_name="testinstances",
+        on_delete=models.CASCADE,
+    )
+    failure_message = models.TextField(null=True)
+    timestamp = models.TextField()
```
adrian-codecov, line 259: Food for thought: do we need a timezone? Would we want to save a timestamp or a datetime? Does it matter?

reports/models.py
```diff
@@ -216,3 +217,46 @@ class UploadLevelTotals(AbstractTotals):
...
+
+    repository = models.ForeignKey(
+        "core.Repository",
+        db_column="repoid",
+        related_name="tests",
+        on_delete=models.CASCADE,
+    )
+    name = models.TextField()
+    testsuite = models.TextField()
+    env = models.TextField(default="")
```
adrian-codecov, line 237: TODO: if you're gonna have a default, just set null=true, see this for reference https://www.notion.so/sentry/Database-Tips-and-Tricks-76df725ded264b2a8154c960d7ef3869?pvs=4#753bec6f64e0462c8f4d9379e79af9a2

reports/models.py
```diff
@@ -216,3 +217,46 @@ class UploadLevelTotals(AbstractTotals):
 
     class Meta:
         db_table = "reports_uploadleveltotals"
+
+
+class Test(models.Model):
+    id = models.TextField(primary_key=True)
```
adrian-codecov, line 223: TODO: leave comment w/ explanation of why this is a non-auto_now field

reports/migrations/0013_test_testinstance.py
```diff
@@ -0,0 +1,75 @@
+# Generated by Django 4.2.7 on 2024-01-17 17:56
+
+import uuid
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
```
adrian-codecov, line 10: TODO: add a sql migrate here, command `python manage.py sqlmigrate report 0013`, see sample of end result here https://github.com/codecov/codecov-api/blob/86431d3ac1d8d982590eede727bcb8b2afe982b0/codecov_auth/migrations/0045_remove_ownerprofile_terms_agreement.py#L8

upload/views/test_results.py
```diff
@@ -0,0 +1,116 @@
...
+                "branch": data.get("branch"),
+                "pullid": data.get("pr"),
+                "merged": False if data.get("pr") is not None else None,
+                "state": "pending",
+            },
+        )
+
+        upload_external_id = str(uuid.uuid4())
+
+        storage_path = f"v1/uploads/{upload_external_id}.json"
```
giovanni-guidini, line 79: [nit] I think there should be some indication in the path that this is a TestResult report. For bundle_analysis the indication is in the bucket (it's a [special bucket](https://github.com/codecov/shared/blob/main/shared/bundle_analysis/storage.py#L12)). But it seems that this one is the same as the coverage reports.

My suggestion is the obvious `test_results/v1/uploads/{upload_external_id}.json`

Personally I'd keep the coverage report's style of adding a repo hash and commit info as well, cause it makes it a tad easier to know that I'm looking at the right report if I have to debug something, but that is just my personal preference and maybe the extra effort is not worth it

upload/views/test_results.py
```diff
@@ -0,0 +1,116 @@
...
+
+        archive_service = ArchiveService(repo)
+        url = archive_service.create_presigned_put(storage_path)
+
+        task_arguments = {
+            # these are used in the upload task when saving an upload record
+            # and use some unfortunately named and confusing keys
+            # (eventual reports_upload columns indicated by comments)
```
giovanni-guidini, line 87: Excellent comments, thanks for that ❤️

upload/tests/views/test_test_results.py
```diff
@@ -0,0 +1,142 @@
...
+        {
+            "commit": "6fd5b89357fc8cdf34d6197549ac7c6d7e5977ef",
+            "slug": f"{repository.author.username}::::{repository.name}",
+        },
+        format="json",
+    )
+    assert res.status_code == 201
+
+
+def test_upload_bundle_analysis_missing_args(db, client, mocker, mock_redis):
```
giovanni-guidini, line 109: test name includes `bundle_analysis`

upload/views/test_results.py
```diff
@@ -0,0 +1,116 @@
...
+                repoid=repo.repoid,
+                task_arguments=task_arguments,
+            ),
+        )
+
+        dispatch_upload_task(
+            task_arguments,
+            repo,
+            get_redis_connection(),
+            report_type=CommitReport.ReportType.TEST_RESULTS,
```
giovanni-guidini, line 113: I noticed that in `dispatch_upload_task` the BUNDLE_ANALYSIS has a delay of 4 ([here](https://github.com/codecov/codecov-api/blob/a9a1f250afb24ed0900b26ff47da5fc5182ea854/upload/helpers.py#L674)), same for coverage reports with version v4.

I believe this delay tells celery "wait 4s to enqueue this task". This is because the user's CI needs time to upload the report to GCS before we can actually process it. Adding the delay (I think) reduces the chance of errors of the kind "we can't process because the upload was not found in storage".

I guess it seems prudent to follow suit and add this delay as well? I suppose it depends on how big the test result reports can be... totally up to you to add or not such delay, but I thought best to point this out anyway.

giovanni-guidini (approved):
LGTM. Left some thought, please address the ones you see fit before merging.

giovanni-guidini:
One more thought: it _might_ be a good idea to add a feature-flag (`shared.rollout.Feature`) so we can test it internally first. I'd ask what product has to say about this.

PR #219, opened 2023-10-31
fix: Disable early notify comment

scott-codecov:
### Purpose/Motivation

Seems there are cases where the pull does not have a `commentid`.  This is causing erroneous notifications that include the "we're processing your commit" message instead of the report info.

### Links to relevant tickets

N/A

### What does this PR do?

Disables the early notify for now.


giovanni-guidini (approved):
Sad sad. Hopefully we can get that rolling again soon

PR #220, opened 2023-10-31
fix: make GraphQL types matches nullability with django models

JerrySentry:
### Purpose/Motivation
Audit all the GraphQL API types to make sure that the nullabilitity (nullable or not nullable) matches that of the database columns (via Django ORM).

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/703

### Legal Boilerplate


graphql_api/types/commit/commit.graphql
```diff
@@ -1,8 +1,8 @@
 type Commit {
   state: String
   message: String
-  createdAt: DateTime
-  commitid: String
+  createdAt: DateTime!
```
adrian-codecov, line 4: So you know how this field doesn't have an explicit resolver? That's because ariadne somehow ends up in your DB model and if it finds a 1:1 mapping with the resolver name and it's field, or through an alias like in this case, it figures it out. I'd like to try and get rid of every implicit resolver (not in this ticket of course) the same way we came to return type parity with this ticket. Could you make an issue that expresses that? 🙏

graphql_api/types/owner/owner.graphql
```diff
@@ -1,5 +1,5 @@
 type Owner {
-  avatarUrl: String
+  avatarUrl: String!
```
adrian-codecov, line 2: I see a lot of logic going on in the avatar_url field, just confirming you went through it all and no children resort to None

graphql_api/types/segment_comparison/segment_comparison.graphql
```diff
@@ -1,5 +1,5 @@
 type SegmentComparison {
-  header: String
+  header: String!
```
adrian-codecov, line 2: Could you please adjust the return type of this resolver to reflect this? Still says optional 👌

adrian-codecov (commented):
Generally looks okay! Can you verify:
- The resolvers for which you made changes also match the graphql return type?
- For the entries you changed to not nullable, can you double check in Prod DB if there is no None values for that column?

graphql_api/types/commit/commit.graphql
JerrySentry: Yeah this `createdAt` field maps to the `timestamp` column in `commits` table which is a not null field hence adding the `!` here. Will make a ticket for implicit resolvers.
https://github.com/codecov/engineering-team/issues/745

graphql_api/types/owner/owner.graphql
JerrySentry: Yep tracing through the `Owner.avatar_url` function, every code path should result in a string being returned.

JerrySentry:
> Generally looks okay! Can you verify:
> 
> * The resolvers for which you made changes also match the graphql return type?
> * For the entries you changed to not nullable, can you double check in Prod DB if there is no None values for that column?

Done, though can not do a check in prod DB for null values for all types eg `select 1 from commits where timestamp is null;` because non-indexed searches on gigantic tables will take forever and might lock up other requests. Though since the prod DB schema for these columns say not-null then it shouldn't be possible to have null values anyways.

adrian-codecov (approved):

PR #222, opened 2023-10-31
Allow for Database options to be set via config

AndreasBergmeier6176:
### Purpose/Motivation
This is especially needed for Postgres 15 where by default users are no longer allowed to write to schema `public`. Being able to set options enables switching to another default schema (e.g. `codecov`)

### Legal Boilerplate

### DIFF
--- a/codecov/settings_base.py
+++ b/codecov/settings_base.py
@@ -1,3 +1,4 @@
+import ast
 import os
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


codecov/settings_base.py
```diff
@@ -217,6 +231,7 @@
         "HOST": DATABASE_READ_HOST,
         "PORT": DATABASE_READ_PORT,
         "CONN_MAX_AGE": CONN_MAX_AGE,
+        "OPTIONS": ast.literal_eval(DATABASE_READ_OPTIONS),
```
scott-codecov, line 234: Could this be a `json.loads` instead here instead?

AndreasBergmeier6176: Sure - would then be a more restricted format, though.

scott-codecov: In what sense?  Were you hoping to pass in Python literals that can't be encoded as JSON?

AndreasBergmeier6176: > Were you hoping to pass in Python literals that can't be encoded as JSON?

Not really. I think I can life with JSON just fine.

PR #224, opened 2023-11-01
feat: add a filter for public/private repo filter for viewableRepository

JerrySentry:
### Purpose/Motivation

- On GraphQL `viewableRepository` resolver, add ability to filter repo list by public/private.
- If `isPublic` is not provided it will return both public and private repos
- If `isPublic=true` it will only return public repos
- If `isPublic=false` it will only return private repos

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/724

### Legal Boilerplate


adrian-codecov (approved):
Lgtm! Thanks for the quick turn!

PR #230, opened 2023-11-03
fix: available plans for users while trialing (#731)

JerrySentry:
### Purpose/Motivation

Fix a bug where a user while trialing does not have Team plans available to upgrade to.
This happens because while trialing we set `plan_user_count` to 1000 while trialing, so we will check `plan_activated_users` while user is trialing. 

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/731

### Legal Boilerplate


plan/test_plan.py
```diff
@@ -611,11 +611,12 @@ def test_available_plans_for_users_trial_plan_ongoing_trial_more_than_10_users(
         expected_result = []
         expected_result.append(BASIC_PLAN)
         expected_result += PR_AUTHOR_PAID_USER_PLAN_REPRESENTATIONS.values()
+        expected_result += TEAM_PLAN_REPRESENTATIONS.values()
```
adrian-codecov, line 614: Oh this shouldn't happen actually, if someone is trialing, we want to only offer the team plan if there are LTE 10 plans. While they're trialling, we only show them the option for team bw 0 and 10 users

plan/test_plan.py
```diff
@@ -628,9 +629,37 @@ def test_available_plans_for_sentry_customer_users_trial_plan_ongoing_trial_more
         expected_result.append(BASIC_PLAN)
         expected_result += PR_AUTHOR_PAID_USER_PLAN_REPRESENTATIONS.values()
         expected_result += SENTRY_PAID_USER_PLAN_REPRESENTATIONS.values()
+        expected_result += TEAM_PLAN_REPRESENTATIONS.values()
```
adrian-codecov, line 632: Same here

plan/test_plan.py
```diff
@@ -628,9 +629,37 @@ def test_available_plans_for_sentry_customer_users_trial_plan_ongoing_trial_more
...
+        self.current_org.save()
+
+        # Can do Team plan when at 10 activated users
+        assert plan_service.available_plans(owner=self.owner) == expected_result
+
+        self.current_org.plan_activated_users = [i for i in range(11)]
+        self.current_org.save()
+
+        # Can not do Team plan when at 11 activated users
+        assert plan_service.available_plans(owner=self.owner) == expected_result[:3]
```
adrian-codecov, line 661: Could we adjust the `expected_result[:3]` to perhaps another list? `expected_result_without_team` or something? If we add another plan eventually, we'll have to refactor this test then

plan/test_plan.py
```diff
@@ -628,9 +629,37 @@ def test_available_plans_for_sentry_customer_users_trial_plan_ongoing_trial_more
         expected_result.append(BASIC_PLAN)
         expected_result += PR_AUTHOR_PAID_USER_PLAN_REPRESENTATIONS.values()
         expected_result += SENTRY_PAID_USER_PLAN_REPRESENTATIONS.values()
+        expected_result += TEAM_PLAN_REPRESENTATIONS.values()
 
         assert plan_service.available_plans(owner=self.owner) == expected_result
 
+    def test_available_plans_for_more_than_10_activated_users(self):
```
adrian-codecov, line 636: And can you indicate in the test title this is for trialing users too? `test_available_plans_for_trialing_users_with_more_than_10_activated_users` or something like that.

I'm okay with splitting this into two tests too, as this says more than 10 activated users, but you test a scenario with < 10 first, then > 10

plan/test_plan.py
```diff
@@ -585,98 +585,82 @@ def test_available_plans_for_sentry_plan_expired_trial_more_than_10_users(
...
     """
 
     def setUp(self):
         self.current_org = OwnerFactory(
+            plan=PlanName.TRIAL_PLAN_NAME.value,
             trial_start_date=datetime.utcnow(),
             trial_end_date=datetime.utcnow() + timedelta(days=14),
             trial_status=TrialStatus.ONGOING.value,
-            plan_user_count=13,
+            plan_user_count=1000,
```
adrian-codecov, line 604: nitpick: could we explicitly add plan_activated_users: None here? I like to initialize everything we use in our tests

PR #232, opened 2023-11-06
feat: Add Sentry user to admin panel (#702)

JerrySentry:
### Purpose/Motivation

Add Sentry user to Django admin dashboard, can search and update Sentry user by name/email and link to the associated User object. Hides access/refresh token fields.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/702

### Legal Boilerplate


codecov_auth/admin.py
```diff
@@ -89,6 +89,34 @@ def has_delete_permission(self, request, obj=None):
...
 
+@admin.register(SentryUser)
+class SentryUserAdmin(AdminMixin, admin.ModelAdmin):
+    list_display = (
+        "name",
+        "email",
+    )
+    search_fields = (
+        "name__iexact",
+        "email__iexact",
```
scott-codecov, line 100: This probably won't scale well since we don't have indexes on these fields but probably fine for now

scott-codecov (approved):

codecov_auth/admin.py
JerrySentry: Yea that's true. A scalability issue to query the sentry_user table would be a great problem to have!

PR #241, opened 2023-11-13
feat: Emit pubsub event when org token is updated

scott-codecov:
Similar to https://github.com/codecov/codecov-api/pull/215
Emit a pubsub event whenever an org token is saved.

Depends on https://github.com/codecov/shelter/pull/58


trent-codecov (approved):

barnett-yuxiang (commented):
LGTM

PR #249, opened 2023-11-15
feat(logging): change access logging format to JSON

rfox-codecov:
This is for https://github.com/codecov/infrastructure-team/issues/162 -- it enables GUnicorn access logs to be output as JSON instead of as HTTP access log format.


trent-codecov (approved):
Looks great!

PR #250, opened 2023-11-15
fix(logging): referer -> referrer

rfox-codecov:
historical nuances be damned, it says referrer in the dictionary!


utils/logging_configuration.py
```diff
@@ -45,7 +45,7 @@ class CustomGunicornLogFormatter(JsonFormatter):
         "levelname": "level",
         "r": "request",
         "a": "useragent",
-        "f": "referer",  # the biggest concern in my mind: should this be correctly spelled historically (one r) or correctly spelled linguistically (two r)
+        "f": "referrer",
```
JerrySentry, line 48: qq: does this break backwards compatibility with anything?

JerrySentry (approved):

utils/logging_configuration.py
rfox-codecov: nope -- this is only for the querying in our logs, which has been flowing for less than a day.

PR #260, opened 2023-11-21
feat: Add an endpoint to get all uploads of a given commit (#112 internal)

JerrySentry:
### Purpose/Motivation

Adds a new endpoint (`/api/v2/{service}/{owner_username}/repos/{repo_name}/commits/{commitsha}/uploads`) to return list of uploads for a given commit report by the commit SHA.

### Links to relevant tickets

https://github.com/codecov/internal-issues/issues/112

### Legal Boilerplate


codecov/settings_base.py
```diff
@@ -28,6 +28,7 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "django.contrib.postgres",
+    "django_extensions",
```
JerrySentry, line 31: Adding this little plugin as it has some useful django manage.py commands to make developing easy such as show_urls and graph_models

api/public/v2/commit/serializers.py
```diff
@@ -47,3 +53,39 @@ class CommitDetailSerializer(CommitSerializer):
     class Meta:
         model = Commit
         fields = CommitSerializer.Meta.fields + ("report",)
+
+
```
JerrySentry, line 57: A subset of fields of the `ReportSession` model, matching the contents of old v1 sessions endpoint, except keys are descriptive instead of single letters from before.

adrian-codecov (approved):

api/shared/commit/serializers.py
```diff
@@ -68,6 +71,10 @@ def get_complexity_ratio(self, totals) -> float:
         )
 
 
+class UploadTotalsSerializer(BaseTotalsSerializer):
```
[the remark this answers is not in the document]
JerrySentry: Technically no, but I just do this to be more explicit about it. And `BaseTotalsSerializer` acts as a ABC even though it is not

PR #265, opened 2023-11-22
feat: adjust schedules logic for upgrades to occur immediately

adrian-codecov:
### Purpose/Motivation
We're adjusting the logic that governs schedules to make changes effective immediate when the desired scheduled value is an upgrade from the original one, either seats or billing rate.

### What does this PR do?
- Adjust the billing file + tests to reflect ^ 

### Notes to Reviewer
[screen-capture (8).webm](https://github.com/codecov/codecov-api/assets/82913673/19a9f9e0-4f50-4959-b927-97ddbac0d78a)

### Legal Boilerplate

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -204,39 +204,42 @@ class StripeService(AbstractPaymentService):
 
         if is_upgrading:
             if subscription_schedule_id:
-                self._modify_subscription_schedule(
-                    owner, subscription, subscription_schedule_id, desired_plan
-                )
-            else:
                 log.info(
-                    f"Updating Stripe subscription for owner {owner.ownerid} to {desired_plan['value']} by user #{self.requesting_user.ownerid}"
+                    f"Releasing Stripe schedule for owner {owner.ownerid} to {desired_plan['value']} with {desired_plan['quantity']} seats by user #{self.requesting_user.ownerid}"
                 )
-                stripe.Subscription.modify(
-                    owner.stripe_subscription_id,
-                    cancel_at_period_end=False,
-                    items=[
-                        {
-                            "id": subscription["items"]["data"][0]["id"],
-                            "plan": settings.STRIPE_PLAN_IDS[desired_plan["value"]],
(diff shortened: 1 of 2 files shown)


services/billing.py
```diff
@@ -204,39 +204,42 @@ def modify_subscription(self, owner, desired_plan):
...
-                )
+            log.info(
+                f"Stripe subscription modified successfully for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
+            )
         else:
             if subscription_schedule_id:
+                new_plan = desired_plan["value"]
+                new_quantity = desired_plan["quantity"]
+
+                print("downgrade", new_plan, new_quantity)
```
JerrySentry, line 241: debugging

services/billing.py
```diff
@@ -204,39 +204,42 @@ def modify_subscription(self, owner, desired_plan):
...
+                name=desired_plan["value"], user_count=desired_plan["quantity"]
+            )
 
-                log.info(
-                    f"Stripe subscription modified successfully for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
-                )
+            log.info(
+                f"Stripe subscription modified successfully for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
+            )
         else:
```
JerrySentry, line 236: under this else block can be simplified to:
```
if not subscription_schedule_id:
    subscription_schedule_id = stripe.SubscriptionSchedule.create(
        from_subscription=owner.stripe_subscription_id
    ).id

self._modify_subscription_schedule(
    owner, subscription, subscription_schedule_id, desired_plan
)
```

JerrySentry (approved):
Pre-approved with one little thing

services/billing.py
adrian-codecov: tyty

services/billing.py
adrian-codecov: makes sense, will make the changes!

PR #267, opened 2023-11-24
feat: Support filters in componentComparisons

RulaKhaled:
### Purpose/Motivation
we need this to filter components tab **and** to filter components on search term 

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/838

### What does this PR do?
- Update comparison to handle the filter
- Tests 

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/comparison/comparison.py
+++ b/graphql_api/types/comparison/comparison.py
@@ -184,18 +184,30 @@ def resolve_flag_comparisons(
 @comparison_bindable.field("componentComparisons")
 @sync_to_async
 def resolve_component_comparisons(
-    comparison_report: ComparisonReport, info
+    comparison_report: ComparisonReport, info, filters=None
 ) -> List[ComponentComparison]:
     current_owner = info.context["request"].current_owner
     head_commit = comparison_report.commit_comparison.compare_commit
     components = components_service.commit_components(head_commit, current_owner)
+    list_components = comparison_report.commit_comparison.component_comparisons.all()
+
+    if filters and filters.get("components"):
+        components = [
+            component
+            for component in components
+            if component.name in filters["components"]
+        ]
+
+        list_components = list_components.filter(
+            component_id__in=[component.component_id for component in components]
+        )
 
     # store for child resolvers (needed to get the component name, for example)
     info.context["components"] = {
         component.component_id: component for component in components
     }
 
-    return list(comparison_report.commit_comparison.component_comparisons.all())
+    return list(list_components)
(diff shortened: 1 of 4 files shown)


JerrySentry (approved):
lgtm

graphql_api/types/comparison/comparison.py
```diff
@@ -184,18 +184,30 @@ def resolve_flag_comparisons(
...
     current_owner = info.context["request"].current_owner
     head_commit = comparison_report.commit_comparison.compare_commit
     components = components_service.commit_components(head_commit, current_owner)
+    list_components = comparison_report.commit_comparison.component_comparisons.all()
+
+    if filters and filters.get("components"):
+        components = [
+            component
+            for component in components
+            if component.name in filters["components"]
```
rohitvinnakota-codecov, line 198: Are component names guaranteed to be unique?

rohitvinnakota-codecov (approved):

graphql_api/types/comparison/comparison.py
```diff
@@ -184,18 +184,30 @@ def resolve_flag_comparisons(
...
+            component_id__in=[component.component_id for component in components]
+        )
 
     # store for child resolvers (needed to get the component name, for example)
     info.context["components"] = {
         component.component_id: component for component in components
     }
 
-    return list(comparison_report.commit_comparison.component_comparisons.all())
+    return list(list_components)
```
JerrySentry, line 210: Thoughts on keeping component filter consistent with flag filtering where in flag filtering if there are no results at the end of filtering we would return an error (UnknownFlags) instead of an empty array?

graphql_api/types/comparison/comparison.py
RulaKhaled: good question i'll follow up on that

RulaKhaled: Looking at https://docs.codecov.com/docs/components a component is unique by id, nevertheless i still think we want to filter by name, as the user is expecting name of components in the dashboard and they would expect the filter to match display names

that being said, i'll raise this with product/design to see if displaying the id in upcoming iteration would be helpful to have a more accurate components filter, thanks Rohit!

graphql_api/types/comparison/comparison.py
RulaKhaled: we huddled on this, the need to create a deprecate field isn't worth it so we're moving forward with this

PR #269, opened 2023-11-27
feat: Support components filter in pathContents (#825)

JerrySentry:
### Purpose/Motivation
Adds component filtering in pathContents query

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/825

### Legal Boilerplate

### DIFF
--- a/api/public/v2/report/views.py
+++ b/api/public/v2/report/views.py
@@ -23,7 +23,7 @@ from codecov_auth.authentication import (
     UserTokenAuthentication,
 )
 from core.models import Commit
-from services.components import commit_components, component_filtered_report
+from services.components import commit_components
 from services.path import ReportPaths, dashboard_commit_file_url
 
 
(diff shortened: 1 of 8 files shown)


RulaKhaled (approved):
Nice Jerry! 🚀

RulaKhaled (changes requested):
Before we merge this, can you push and test on staging please? 

something i want to confirm is that although commit_report is different than filtered commit report, it still behaves as intended

JerrySentry:
> Before we merge this, can you push and test on staging please?
> 
> something i want to confirm is that although commit_report is different than filtered commit report, it still behaves as intended
@RulaKhaled 
Did some testing with the two components on worker repo, seems fine to me. I was investigating into `FilteredReport` vs `Report` recently, afaik they share the same properties, and in FilteredReport it has an instance of the Report then when we access the properties in the FilteredReport it will do a bunch of flag filtering logic against the store Report instance.

RulaKhaled:
> > Before we merge this, can you push and test on staging please?
> > something i want to confirm is that although commit_report is different than filtered commit report, it still behaves as intended
> 
> @RulaKhaled Did some testing with the two components on worker repo, seems fine to me (I mostly used my local dev setup with better components setup prior to this). I was investigating into `FilteredReport` vs `Report` recently, afaik they share the same properties, and in FilteredReport it has an instance of the Report then when we access the properties in the FilteredReport it will do a bunch of flag filtering logic against the store Report instance.

Sounds good to me!

RulaKhaled (approved):

PR #274, opened 2023-11-29
fix: component with flag filtering

JerrySentry:
### Purpose/Motivation

Fix an issue with `pathContents` resolver not being able to handle component and flag filtering at the same time

### Legal Boilerplate


graphql_api/types/commit/commit.py
```diff
@@ -190,12 +191,29 @@ def resolve_path_contents(commit: Commit, info, path: str = None, filters=None):
                 f"missing coverage for report with components: {component_filter}"
             )
 
-        commit_report = components.component_filtered_report(
-            commit_report, filtered_components
-        )
+        for component in filtered_components:
+            component_paths.extend(component.paths)
+
+        if flags_filter:
```
RulaKhaled, line 197: what if there is no flags_filter but there are flags embedded in components? maybe you can do something like:

```
components_flags = []
 for component in filtered_components:
      components_flags.extend(
          component.get_matching_flags(commit_report.flags.keys())
      )
if components_flags:
  flags_filter = list(set(flags_filter) & set(component_flags))
```

just to make sure we're always handling the components flags

services/path.py
```diff
@@ -174,11 +178,16 @@ def __init__(
...
     @cached_property
     def files(self) -> List[str]:
+        files = []
+
         if self.filter_flags:
             files = report_service.files_belonging_to_flags(
                 commit_report=self.unfiltered_report, flags=self.filter_flags
             )
+            if self.filter_paths:
+                files = list(set(files) & set(self.report.files))
```
RulaKhaled, line 188: why are we getting the intersection between files and report.files here 🤔? 

Some thoughts:
- We are using `self.unfiltered_report` to get files that belong to flags from sessions, we might need to do the same thing for filtered report. unless report.files returns the filtered files expected then we might as well just use that.
- if it's any easier, we might need to filter `report.files` by paths and just return the set of the combined array

graphql_api/types/commit/commit.py
JerrySentry: Discussed this, we will be going with intersection query.

services/path.py
JerrySentry: Discussed over Slack.

graphql_api/types/commit/commit.py
RulaKhaled: The code i suggested doesn't disapprove the intersection query, we still need to take out the components_flags from the flags_filter condition block, this is only calling the components flags if filter_flags exist and we would want to avoid this

graphql_api/types/commit/commit.py
```diff
@@ -190,12 +189,31 @@ def resolve_path_contents(commit: Commit, info, path: str = None, filters=None):
...
+        flags_filter = set(flags_filter)
+        if not (flags_filter & set(commit_report.flags)):
+            return UnknownFlags()
+        if not (flags_filter & set(component_flags)):
+            return UnknownFlags(
+                f"unknown flags for report with components: {component_filter}"
+            )
+        flags_filter = list(flags_filter & set(component_flags))
+    else:
+        flags_filter = list(set(flags_filter) | set(component_flags))
```
RulaKhaled, line 209: if there's not flags_filter then we know it's def gonna be `component_flags` filter so no need for the OR here

PR #292, opened 2023-12-05
feat: Add component filtering for coverageFile type in GQL

JerrySentry:
### Purpose/Motivation

In GraphQL API when we query the `coverageFile` type add an option in the input to filter by list of component names, This will work in conjunction with filtering by list of flags. If no coverage for the given components filters will return 0s for totals similar to if no coverage for the given flags. When filtering by both flags and components the result will be an intersection between flags and components.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/826

### Legal Boilerplate


graphql_api/types/commit/commit.py
```diff
@@ -39,9 +41,22 @@
...
+            all_components, components
+        )
+        for fc in filtered_components:
+            paths.extend(fc.paths)
+        _else = FilteredReportFile(ReportFile(path), [])
+
+    commit_report = commit.full_report.filter(flags=flags, paths=paths)
+    file_report = commit_report.get(path, _else=_else)
 
     return {
```
rohitvinnakota-codecov, line 61: It might make sense to also return components here as part of the dict similar to how we are returning flags.

rohitvinnakota-codecov (approved):

PR #293, opened 2023-12-05
Auto activate owner when default org is updated or created 

rohitvinnakota-codecov:
### Purpose/Motivation

This PR fixes https://github.com/codecov/engineering-team/issues/304. When a user chooses their default org, they are not automatically added unless they open a PR in that org. With this change, they user is auto assigned a seat when they update their default org. 

- [x] Test and verify locally with umbrella-hat.

### Legal Boilerplate


codecov_auth/commands/owner/interactors/update_default_organization.py
```diff
@@ -30,7 +31,10 @@ def update_default_organization(self, default_org: Optional[Owner]):
             owner_id=self.current_owner.ownerid
         )
         owner_profile.default_org = default_org
-        return owner_profile.save()
+        saved_owner_profile = owner_profile.save()
+        if default_org:
```
adrian-codecov, line 35: We should be able to get rid of this line since it's accounted for above in the `validate` function

adrian-codecov (approved):
Lgtm, just that little comment I left

codecov_auth/commands/owner/interactors/update_default_organization.py
rohitvinnakota-codecov: We're not raising any exceptions in the validation function if the org is `None` so it looks like we can still execute this flow with a null org.

PR #298, opened 2023-12-07
Upgrade pytest-django

scott-codecov:
The old version was causing some conflicts with database access.  If the database connection was first setup through the pytest fixture then subsequent tests (that used the Django `TestCase` class) would fail.


giovanni-guidini (approved):
You are my hero 🥇

PR #303, opened 2023-12-12
fix: Report filtering by path pattern string (#internal_148)

JerrySentry:
### Purpose/Motivation

Fix the regex pattern matching for path filtering for Reports. `*` pattern will consume the previous pattern 0 or more times, in this case `{path}.*` means prefix matching. Before, `{path}*` will unintentionally consume the previous character of the path string.

### Links to relevant tickets

https://github.com/codecov/internal-issues/issues/148

### Legal Boilerplate

### DIFF
--- a/api/public/v2/report/views.py
+++ b/api/public/v2/report/views.py
@@ -120,7 +120,7 @@ class BaseReportViewSet(
         if path and flag:
             report = report.filter(flags=[flag], paths=[f"{path}*"])
         elif path:
-            report = report.filter(paths=[f"{path}*"])
+            report = report.filter(paths=[f"{path}.*"])
         elif flag:
             report = report.filter(flags=[flag])
         elif component_id:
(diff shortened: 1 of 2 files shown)


adrian-codecov (approved):

api/public/v2/report/views.py
```diff
@@ -118,9 +118,9 @@ def filter_report(
                 return Report()
 
         if path and flag:
-            report = report.filter(flags=[flag], paths=[f"{path}*"])
+            report = report.filter(flags=[flag], paths=[f"{path}.*"])
```
adrian-codecov, line 121: Could we add a test that makes the distinction bw the previous solution and this one? I'm assuming certain paths didn't use to work but now they do?

JerrySentry: Sure thing.
All paths used to work, except they may return more results than expected. For example if you have the folders:
`App/Test` and `App/TestTwo`, you search for path filter `"App/Test/"` it should return only the first folder, not both.

PR #304, opened 2023-12-12
feat: tokenless auth (part I)

giovanni-guidini:
context: codecov/engineering-team#736

We want to accept tokenless uploads for forks of public repos. This is a feature that is currently missing from the CLI and upload.

These changes are part I of that.
What it adds:
* `TokenlessAuthentication` - a new auth class that controls when a tokenless upload is accepted
* scaffolding `TokenlessAuth`
* `TOKENLESS_AUTH_BY_OWNER_SLUG` feature for internal testing
* The logic that will let us accept a tokenless upload

What is missing:
* Implementation of the methods in `TokenlessAuth`, so that it can actually be used
* Hooking up the auth to existing endpoints.

Conditions to accept tokenless upload:
1. Request has `X-Tokenless` header, value is the fork repo slug AND `X-Tokenless-PR` header, value is the PR number
2. Request is to one of the new upload endpoints
3. The encoded repo (in the request path) exists AND is public repo
4. (temp) the repo is from an owner that can accept tokenless
5. The git service provider reports that the PR in `X-Tokenless-PR` indeed exists and is from `X-Tokenless` repo to the encoded repo.

ps.: I decided to break the changes in more than 1 PR cause this one is getting pretty big already
pss.: also sneaking in an addition to .gitignore


giovanni-guidini:
@matt-codecov 
> the API already has everything it needs to fetch PR info without X-Tokenless and X-Tokenless-PR, right?

Can you elaborate on this?

Are you suggesting that for the `create-commit` we look in the request payload to make the verification, and for the other requests (`create-report` and `do-upload`) that don't contain the PR number not the fork name we pull the commit from the database and check again from that?

I would argue that this is more complicated, not the same for all requests, and actually also spoofable cause you can override the details on the `create-commit` command

giovanni-guidini:
> looking back at dana's PRs she made the CLI send the fork:branch branch name https://github.com/codecov/codecov-cli/pull/331 but tbh we should probably just compute it based on the fetched PR details and not rely on the CLI

That is fair, I can add that.

> logic should all happen in the auth controller

I don't think the auth controller should change the request itself. But we can add details in the serializer or the `perform_create` functions.

.gitignore
```diff
@@ -97,6 +97,7 @@ env.bak/
 venv.bak/
 .testenv
 .envrc
+.debug
```
adrian-codecov, line 100: also sneaky

rollouts/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+from codecov_auth.models import Owner
+
+
+def owner_slug(owner: Owner) -> str:
+    return f"{owner.service}/{owner.username}"
+
+
+# By default, features have one variant:
+#    { "enabled": FeatureVariant(True, 1.0) }
+TOKENLESS_AUTH_BY_OWNER_SLUG = Feature(
```
adrian-codecov, line 12: What, never heard of this

codecov_auth/authentication/repo_auth.py
```diff
@@ -174,3 +202,83 @@ def authenticate_credentials(self, key):
...
+    async def get_pull_request_info(self, repository_service, fork_pr: str):
+        try:
+            return await repository_service.get_pull_request(fork_pr)
+        except TorngitObjectNotFoundError:
+            raise exceptions.AuthenticationFailed(self.auth_failed_message)
+
+    def authenticate(self, request):
+        fork_slug = request.headers.get("X-Tokenless", None)
+        fork_pr = request.headers.get("X-Tokenless-PR", None)
+        if not fork_slug or not fork_pr:
```
adrian-codecov, line 255: You can have either one for it to work right? You don't need both?

adrian-codecov (approved):
Lgtm as well! Do we want to log this in any way?

giovanni-guidini:
This depends on https://github.com/codecov/shared/pull/101 now

PR #306, opened 2023-12-13
fix test that fails depending on run order

giovanni-guidini:
If you run
```
pytest \
 graphql_api/tests/test_impacted_file.py::TestImpactedFileFiltering::test_filtering_with_successful_flags \
 compare/commands/compare/interactors/tests/test_fetch_impacted_files.py::FetchImpactedFilesTest::test_impacted_files_filtered_by_flags_and_commit_comparison_for_pull
```

you'll see the 2nd test fails.

This is because of the PullID that apparently both tests create a pullID of 1.
It is strange because they are `TransactionTestCase` and that's supposed to reset the DB,
but God knowns what is going on.

Hard setting the pullID to different numbers solves the issue, but I'll be the 1st
to admit is not a pretty fix

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/compare/commands/compare/interactors/tests/test_fetch_impacted_files.py
+++ b/compare/commands/compare/interactors/tests/test_fetch_impacted_files.py
@@ -590,6 +590,7 @@ class FetchImpactedFilesTest(TransactionTestCase):
             CommitFactory(repository=repo),
         )
         pull = PullFactory(
+            pullid=256,
             repository=repo,
             base=base.commitid,
             head=head.commitid,


compare/commands/compare/interactors/tests/test_fetch_impacted_files.py
```diff
@@ -590,6 +590,7 @@ def test_impacted_files_filtered_by_flags_and_commit_comparison_for_pull(
             CommitFactory(repository=repo),
         )
         pull = PullFactory(
+            pullid=256,
```
JerrySentry, line 593: - Does this mean setting to any ID less than 256 would not work?
- If you add a test (that creates a Pull object in the DB) that runs before this one, does that mean this would fail? Since the next available unused ID is 257?

giovanni-guidini: > Does this mean setting to any ID less than 256 would not work?
No, it's just a random number.

> If you add a test (that creates a Pull object in the DB) that runs before this one, 
well the other test I mentioned that causes this issue creates another pull object in the database with ID 44. So, no? 

Where did you get that the next available ID is 257?...

JerrySentry: Oh ok I didn't know it was a random number, I though anything below 256 is used.

JerrySentry (approved):

PR #308, opened 2023-12-14
feat: tokenless auth (part II)

giovanni-guidini:
Implement `TokenlessAuth` details, so it actually
does things.

Hook it tokenless auth to the upload endpoints that are supposed to support it.


upload/views/commits.py
```diff
@@ -23,12 +26,41 @@ class CommitViews(ListCreateAPIView, GetterMixin):
...
+        # is the correct repo from the head of a PR to the upstream repo
+        # with the git provider
+        fork_slug = request.headers.get("X-Tokenless", None)
+        branch_info = request.data.get("branch")
+        if branch_info is None:
+            # There should always be a branch in the request
+            raise ValidationError("missing branch")
+        # The CLI might have pre-prended the branch with something already
+        if ":" in branch_info:
+            _, branch_info = branch_info.split(":")
```
adrian-codecov, line 55: Would branch_info always have 2 items if it had a ":"?

upload/tests/views/test_commits.py
```diff
@@ -165,3 +168,103 @@ def test_create_commit_already_exists(db, client, mocker):
...
+        "pullid": 4,
+        "state": None,
+        "timestamp": commit.timestamp.strftime("%Y-%m-%dT%H:%M:%S.%fZ"),
+    }
+    assert expected_response == response_json
+    mocked_call.assert_called_with(commitid="commit_sha", repoid=repository.repoid)
+    fake_provider_service.get_pull_request.assert_called_with("4")
+
+
+def test_commit_tokenless_missing_branch(db, client, mocker):
```
adrian-codecov, line 235: Is tokenless reliant on there being a branch?

codecov_auth/tests/unit/test_repo_authentication.py
```diff
@@ -341,3 +345,226 @@ def test_orgleveltoken_success_auth_enterprise(self, db, mocker):
...
+        request = APIRequestFactory().post(
+            "/upload/unknown_provider/owner::::repo/commits/commit_sha/reports/report_code/uploads",
+            headers={"X-Tokenless": "user-name/repo-forked", "X-Tokenless-PR": "15"},
+        )
+        authentication = TokenlessAuthentication()
+        with pytest.raises(exceptions.AuthenticationFailed) as exp:
+            authentication.authenticate(request)
+        assert str(exp.value) == "Not valid tokenless upload"
+
+    def test_tokenless_not_supported_services(self):
```
adrian-codecov, line 384: Does tokenless support `gh` and `github` and all there uppercase/lowercase variants?

upload/views/commits.py
giovanni-guidini: I think so. we do `fork_slug:branch`. And as mentioned by Dana [here](https://github.com/codecov/codecov-cli/pull/331#discussion_r1392259208) branches can't have `:` in them. I don't think repos can either.

upload/tests/views/test_commits.py
giovanni-guidini: only in this implementation, but it would be weird to not have a branch

codecov_auth/tests/unit/test_repo_authentication.py
giovanni-guidini: depends if the upload URL supports that or not. I don't think it does.
If you look in the `GetterMixin` ([here](https://github.com/codecov/codecov-api/blob/c69525cafd7d744437798990b845b10b2f19d74b/upload/views/base.py#L31)) you'll see that the `Service` has a limited number of options.

upload/tests/views/test_commits.py
adrian-codecov: yaaa makes sense, more so me learning how this works heh

codecov_auth/tests/unit/test_repo_authentication.py
adrian-codecov: Good to know in case we ever want to support/change for multiple services

PR #312, opened 2023-12-19
fix: allow self hosted admins to have parity with cloud admins

JerrySentry:
### Purpose/Motivation
See https://github.com/codecov/feedback/issues/181

- Allow erase repository coverage
- Allow flag deletion

### Legal Boilerplate

### DIFF
--- a/core/commands/flag/interactors/delete_flag.py
+++ b/core/commands/flag/interactors/delete_flag.py
@@ -24,8 +24,9 @@ class DeleteFlagInteractor(BaseInteractor):
         if not repo:
             raise ValidationError("Repo not found")
 
-        if settings.IS_ENTERPRISE and not self_hosted.is_admin_owner(self.current_user):
-            raise Unauthorized()
+        if settings.IS_ENTERPRISE:
+            if not self_hosted.is_admin_owner(self.current_user):
+                raise Unauthorized()
 
     def execute(self, owner_username: str, repo_name: str, flag_name: str):
         owner = Owner.objects.filter(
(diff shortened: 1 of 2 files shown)


scott-codecov (approved):

core/commands/flag/interactors/delete_flag.py
```diff
@@ -21,8 +24,9 @@ def validate(self, owner: Owner, repo: Repository):
         if not repo:
             raise ValidationError("Repo not found")
 
-        if not owner.is_admin(self.current_user):
```
scott-codecov, line 24: Should probably keep this part

core/commands/flag/interactors/delete_flag.py
```diff
@@ -21,8 +24,9 @@ def validate(self, owner: Owner, repo: Repository):
...
-            raise Unauthorized()
+        if settings.IS_ENTERPRISE:
+            if not self_hosted.is_admin_owner(self.current_user):
```
scott-codecov, line 28: I think this should take an `Owner` instance instead of a `User`

core/commands/flag/interactors/delete_flag.py
JerrySentry: thanks!

core/commands/flag/interactors/delete_flag.py
JerrySentry: `self.current_user` is an Owner object

scott-codecov (approved):

PR #317, opened 2023-12-21
feat: add resolver to expose private repos

adrian-codecov:
### Purpose/Motivation
We're adding a resolver to surface whether a repo has any private repos.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/939

### What does this PR do?
- Adds a resolver for an owner to surface if it has any private repos
- Adds a property to the owner model

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/tests/test_owner.py
+++ b/graphql_api/tests/test_owner.py
@@ -587,3 +587,34 @@ class TestOwnerType(GraphQLTestHelper, TransactionTestCase):
 
         data = self.gql_request(query, owner=current_org)
         assert data["owner"]["hasPrivateRepos"] == True
+
+    def test_owner_query_with_public_repos(self):
+        current_org = OwnerFactory(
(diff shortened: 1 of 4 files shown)


graphql_api/tests/test_owner.py
```diff
@@ -569,3 +569,21 @@ def test_owner_query_with_no_service(self):
 
         assert res["errors"][0]["message"] == MissingService.message
         assert res["data"]["owner"] is None
+
+    def test_owner_query_with_private_repos(self):
```
RulaKhaled, line 573: might be worth it to test if owner doesn't have any private repos

RulaKhaled (approved):

graphql_api/tests/test_owner.py
adrian-codecov: Done!

PR #321, opened 2023-12-27
Add an isATSConfigured field to the repository GQL type

rohitvinnakota-codecov:
### Purpose/Motivation
Adds a new gql field on a repo level to check if ATS is configured. This attribute will be consumed in gazebo to render a new ATS tab. 

### Links to relevant tickets

This PR relates to https://github.com/codecov/engineering-team/issues/930. A follow up PR will add the `language` field to the type.

Repos with ATS configured have certain values in their YAML file. We have a similar check in [worker](https://github.com/codecov/worker/blob/0d4ec536d5d384a75ff552f657411dfc472afc22/services/notification/notifiers/mixins/message/sections.py#L330). 

### Legal Boilerplate


graphql_api/types/repository/repository.py
```diff
@@ -283,6 +283,15 @@ def resolve_flags_measurements_backfilled(repository: Repository, info) -> bool:
     return dataset.is_backfilled()
 
 
+@repository_bindable.field("isATSConfigured")
+def resolve_is_ats_configured(repository: Repository, info) -> bool:
+    if not repository.yaml or "flag_management" not in repository.yaml:
```
adrian-codecov, line 288: If what you're checking is if repository.yaml exists/is not an empty dict, I'd do `if repository.yaml is not None` vs this ^. If you had an empty dict, this evaluates to true `if not {} // True`. Check this out https://stackoverflow.com/questions/3965104/not-none-test-in-python for general rules of thumb

graphql_api/types/repository/repository.py
```diff
@@ -283,6 +283,15 @@ def resolve_flags_measurements_backfilled(repository: Repository, info) -> bool:
...
+        return False
+
+    individual_flags = repository.yaml["flag_management"].get("individual_flags", {})
+    return individual_flags.get("carryforward_mode") == "labels"
```
adrian-codecov, line 292: Id add a default value for `carryforward_mode` here, so `individual_flags.get("carryforward_mode", '')`

graphql_api/tests/test_repository.py
```diff
@@ -424,3 +426,16 @@ def test_repository_not_found(self):
...
             "message": "Not found",
         }
+
+    def test_repository_has_ats_configured(self):
+        repo = RepositoryFactory(
+            author=self.owner,
+            active=True,
+            private=True,
+            yaml={
+                "flag_management": {"individual_flags": {"carryforward_mode": "labels"}}
```
adrian-codecov, line 436: I'd suggest adding tests for an empty yaml or when flag_management isn't in the yaml

graphql_api/types/repository/repository.py
adrian-codecov, line 288: Also, have you double checked if you want the repository.yaml or some other yaml value? Check this fn out, `UserYaml.get_final_yaml(...)`, the yaml can be from a commit, repo and/or org, so I'd double check on your end

graphql_api/types/repository/repository.py
adrian-codecov, line 292: Last but not least, I'd give a bit of an explanation why labels = ats configured. Will come in handy when someone, like myself, that doesn't know much about ATS sees this and needs the context for why this leads to ATS you know

graphql_api/types/repository/repository.py
rohitvinnakota-codecov: Do you feel strongly about this? It seems like it is a matter of preference and I can't see any logical inconsistency between the methods. It looks like `if not {}` returns [False](https://replit.com/@RohitVinnakota2/SelfreliantEmbellishedPolygon#main.py).

graphql_api/types/repository/repository.py
rohitvinnakota-codecov: `.get()` returns `None` by default so the code still executed as expected in this case.

graphql_api/tests/test_repository.py
rohitvinnakota-codecov: The other tests check for this situation and I have updated those to reflect that. See lines 126, 177.

graphql_api/types/repository/repository.py
rohitvinnakota-codecov: It looks like it is on the [repository](https://github.com/codecov/codecov-api/commit/6ad32c93f5a2afcca6221c98a3acbf31b92dd7b2#diff-3e878d9bdc47f29a0ab909a7db939acf08d2d2f1aaba81e6f897b32361fa40db). Thanks for the caution here.

adrian-codecov: So this suggestion is mostly since the values of `repository.yaml` can be null by default. Granted, doing `not` on None object will also yield True (cause not None ~ not False = True, see pic below), so I often like to check against the type itself, not strongly about it

Screenshot 2023-12-28 at 10 05 13 AM

graphql_api/types/repository/repository.py
adrian-codecov: True that, I mostly try to give an explicit default to be explicit 😂 but this is okay

adrian-codecov (approved):

Issue #297, opened 2023-12-07
Update Sentry and add in Spotlight


nicholas-codecov:
Wondering if we can update the Sentry dependency to the latest version and enable the spotlight option when running in development mode so that we can get full stack traces when running both the API and Gazebo locally.

Details for the required version and enabling Spotlight can be found [here](https://spotlightjs.com/setup/other/)

Feel free to poke me if you have any questions around this.

PR #322, opened 2023-12-27
feat: Add sentry spotlight (#297)

JerrySentry:
### Purpose/Motivation

Enable Sentry Spotlight when in development environment.

### Links to relevant tickets

https://github.com/codecov/codecov-api/issues/297

### Legal Boilerplate


codecov/settings_base.py
```diff
@@ -526,6 +526,10 @@
             ),
         },
     )
+elif IS_DEV:
+    sentry_sdk.init(
```
rohitvinnakota-codecov, line 530: Do we also want the `dsn` param here?

https://spotlightjs.com/setup/other/

JerrySentry: I don't think so, in the docs it says:
> If you want your Sentry SDKs to talk to the Sidecar, enable the spotlight setting. You don’t need a dsn for Spotlight to work.
and we don't actually have a `SENTRY_DSN` in dev.

rohitvinnakota-codecov (approved):

PR #323, opened 2023-12-29
[feat] Add primary language field to repository GQL

rohitvinnakota-codecov:
### Purpose/Motivation

Following up from #321, adds the language field to the Repository type. This will be consumed in gazebo to conditionally render an `ATS` tab. This PR closes https://github.com/codecov/engineering-team/issues/930. 

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/930

### Legal Boilerplate

### DIFF
--- a/graphql_api/tests/test_repository.py
+++ b/graphql_api/tests/test_repository.py
@@ -68,6 +68,7 @@ default_fields = """
     criticalFiles { name }
     graphToken
     yaml
+    isATSConfigured
     bot { username }
 """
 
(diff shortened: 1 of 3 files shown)


JerrySentry (approved):
Main discussion here: https://sentry.slack.com/archives/C04LW8R1M2N/p1704210521062999

Its fine to add this field to the GQL response, however beware what this language value represents and how it will be consumed. Its an arbitrary value retrieved from git providers' APIs, just because it shows one language doesn't mean that language exists in the repo nor does it mean it does not have any other languages.

PR #324, opened 2023-12-29
feat: expand message for gm marketplace customer when customer has previous plan

adrian-codecov:
### Purpose/Motivation
Closes https://github.com/codecov/internal-issues/issues/150

### What does this PR do?
Adds a quick extra params to ghm customers

### Legal Boilerplate

### DIFF
--- a/webhook_handlers/views/github.py
+++ b/webhook_handlers/views/github.py
@@ -507,7 +507,7 @@ class GithubWebhookHandler(APIView):
                     "GHM webhook - user purchasing but has a Stripe Subscription",
                     extra=dict(
                         username=username,
-                        plan_name=subscription.plan.get("name", None),
+                        old_plan_name=subscription.plan.get("name", None),
                         quantity=subscription.quantity,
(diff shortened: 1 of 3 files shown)


webhook_handlers/views/github.py
```diff
@@ -505,7 +505,11 @@ def _handle_marketplace_events(self, request, *args, **kwargs):
             if subscription.status == "active":
                 log.warning(
                     "GHM webhook - user purchasing but has a Stripe Subscription",
-                    extra=dict(username=username),
+                    extra=dict(
+                        username=username,
+                        plan_name=subscription.plan.get("name", None),
```
rohitvinnakota-codecov, line 510: Should we call this `old_plan_name` to be more explicit?

rohitvinnakota-codecov (approved):

webhook_handlers/views/github.py
adrian-codecov: Sounds good! Will do

PR #326, opened 2024-01-02
feat: add fn to update stripe email address

adrian-codecov:
### Purpose/Motivation
Customers want to update the stripe email address via Gazebo. This is the backend portion for this change.

### What does this PR do?
- Add `update_email_address` fn and tests

### Legal Boilerplate

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -1,4 +1,6 @@
+import email
 import logging
+import re
 from abc import ABC, abstractmethod
 
 import stripe
@@ -61,6 +63,10 @@ class AbstractPaymentService(ABC):
 
     @abstractmethod
     def update_payment_method(self, owner, payment_method):
+        pass
+
+    @abstractmethod
+    def update_email_address(self, owner, email_address):
         pass
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


services/billing.py
```diff
@@ -405,6 +411,22 @@ def update_payment_method(self, owner, payment_method):
...
+        if not re.fullmatch(r"[^@]+@[^@]+\.[^@]+", email_address):
+            return None
+
+        log.info(f"Stripe update email address owner {owner.ownerid}")
+        if owner.stripe_subscription_id is None:
+            log.info(
+                f"stripe_subscription_id is None, no updating stripe email for owner {owner.ownerid}"
+            )
+            return None
+        stripe.Customer.modify(owner.stripe_customer_id, email=email_address)
```
JerrySentry, line 425: Are there any error handling we need to do here? We should catch it and handle accordingly. Perhaps their call has validation for email regex, ideally we don't do the email validation if they have it.

services/billing.py
```diff
@@ -1,4 +1,6 @@
+import email
```
JerrySentry, line 1: unused import

JerrySentry (commented):
All the log messages has subtle grammatically errors 😂

services/billing.py
adrian-codecov: Their call meaning the client's call? If so, I think thinking of a service as standalone and say eventually using this in our open api, it should validate things don't you think?

I was debating error validation but honestly just followed what we had in other services. What were you thinking though, a good old `ValidationError`?

JerrySentry: I was wondering if `stripe.Customer.modify` API call does email validation, if so then we don't need to do it ourselves. But looking at the [doc](https://stripe.com/docs/api/customers/update) its unclear/doubtful that they do that, so forget what I said.

Also, a general Stripe billing question, how are we handling exceptions raised by the Stripe API calls? Seems like we have a `_log_stripe_error` wrapper the logs the exception and re-raises, but I don't see how we are catching it afterwards.

JerrySentry:
> @JerrySentry which sentences 👀? And how would you write them?

Eh its fine, I just got tripped up reading them the first time. English is my second language 😛 

> Stripe update email address _for_ owner

> stripe_subscription_id is None, ~no~ _not_ updating stripe email for owner

> Stripe ~success update~ _successfully updated_ email address for owner

services/billing.py
adrian-codecov: We basically have that decorator that acts as a fn's try/catch, so if we have a stripe specific error, we'll catch it and raise it in the return, otherwise we continue as normal. So say raising validation for the email could be a scenario to do our custom try/catch cause that could fail but not because it's a stripe api error. Seemingly, what we've historically cared here is to focus on the stripe error and return None otherwise, but I can see a use case here

JerrySentry (approved):
👍

PR #333, opened 2024-01-09
feat: Add languages and bundle_analysis_enabled columns to repos

JerrySentry:
### Purpose/Motivation

Add two columns: `languages` stores list of supported languages in a given repo, `bundle_analysis_enabled` is true when the user has uploaded bundle reports, ie they are using bundle analysis feature

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/963
https://github.com/codecov/engineering-team/issues/1005

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


core/models.py
```diff
@@ -127,6 +128,7 @@ class Languages(models.TextChoices):
     )
     activated = models.BooleanField(null=True, default=False)
     deleted = models.BooleanField(default=False)
+    bundle_analysis_enabled = models.BooleanField(null=True, default=False)
```
adrian-codecov, line 131: What would be the use case for a null value here?

JerrySentry: Good point, there is no case, removed it. I just copy pasted the coverage activated field.

adrian-codecov (approved):

PR #339, opened 2024-01-15
feat: Add bundle analysis report breakdown for a given commit

JerrySentry:
### Purpose/Motivation

GQL: Add a resolver in Commit type, that gets the bundle report of the commit, that displays the total size of all the bundles and size of each individual bundle

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/1027

### Legal Boilerplate


graphql_api/types/bundle_analysis_report/bundle_analysis_report.py
```diff
@@ -0,0 +1,46 @@
...
+@bundle_analysis_report_result_bindable.type_resolver
+def resolve_bundle_analysis_report_result_type(obj, *_):
+    if isinstance(obj, BundleAnalysisReport):
+        return "BundleAnalysisReport"
+    elif isinstance(obj, MissingHeadReport):
+        return "MissingHeadReport"
+
+
+@bundle_analysis_report_bindable.field("sizeTotal")
+def resolve_size_total(bundles_analysis_report: BundleAnalysisReport, info):
```
adrian-codecov, line 20: Could we add return types for these resolvers pls?

graphql_api/types/bundle_analysis_report/bundle_analysis_report.py
```diff
@@ -0,0 +1,46 @@
...
+def resolve_load_time_total(bundles_analysis_report: BundleAnalysisReport, info):
+    return bundles_analysis_report.load_time_total
+
+
+@bundle_analysis_report_bindable.field("bundles")
+def resolve_bundles(bundles_analysis_report: BundleAnalysisReport, info):
+    return bundles_analysis_report.bundles
+
+
+@bundle_report_bindable.field("name")
```
adrian-codecov, line 34: Wdyt about separating bundle analysis reports and bundles into their own folders? Rn they're similar but we could eventually make that difference more pronounced, so might make sense to separate that into different folders

graphql_api/types/commit/commit.py
```diff
@@ -160,6 +163,12 @@ def resolve_bundle_analysis_compare_with_parent(commit: Commit, info, **kwargs):
     return load_bundle_analysis_comparison(base_commit, commit)
 
 
+@commit_bindable.field("bundleAnalysisReport")
+@sync_to_async
+def resolve_bundle_analysis_report(commit: Commit, info, **kwargs):
```
adrian-codecov, line 168: Also return type here please 😬

graphql_api/dataloader/bundle_analysis.py
```diff
@@ -33,3 +33,17 @@ def load_bundle_analysis_comparison(
...
         head_report_key=head_report.external_id,
     )
+
+
+def load_bundle_analysis_report(commit: Commit) -> BundleAnalysisReport:
+    report = CommitReport.objects.filter(
+        report_type=CommitReport.ReportType.BUNDLE_ANALYSIS, commit=commit
+    ).first()
+    if report is None:
+        return MissingHeadReport()
```
adrian-codecov, line 43: Could this fn ever load a non-head report? Trying to see if it's better to say MissingReport as opposed of the head

services/bundle_analysis.py
```diff
@@ -40,6 +41,36 @@ def load_time_conversion(size):
     return round((8 * size) / (1024 * 1024 * 3), 1)
 
 
+@dataclass
+class BundleAnalysisReport(object):
+    def __init__(self, report: SharedBundleAnalysisReport):
+        self.report = report
+        self.report_bundles = []
+        self.total_size = 0
+        for bundle in self.report.bundle_reports():
```
adrian-codecov, line 50: TODO as we spoke previously:

1) add functionality to calculate total BundleAnalysisReport size in shared
2) rethink how to ensure we close/start all db connections in shared as well, fine here for now
3) replace BundleReport to use the SharedBundleReport to avoid disconnect bw data here and data from shared

(could you make an issue capturing this when you get a chance please?)

graphql_api/dataloader/bundle_analysis.py
JerrySentry: Doesn't appear like we have anything more suitable. FWIW we do use `MissingHeadReport` on non comparison like Commits.pathContents

graphql_api/types/bundle_analysis_report/bundle_analysis_report.py
JerrySentry: I don't see there being an entry point into bundles without going through bundle analysis though. If/when that does come up in the future, we can separate it, but for now I think the bundle analysis report and bundle analysis comparison folders separation is intuitive.

services/bundle_analysis.py
JerrySentry: https://github.com/codecov/engineering-team/issues/1067
https://github.com/codecov/engineering-team/issues/1068

Yep will refactor this code next when tackling these two tickets.

PR #340, opened 2024-01-15
feat: Add a coverage enabled column to repos

JerrySentry:
### Purpose/Motivation

Create a column to determine if a user has coverage running. This differentiates from `active` because that is if _any_ commit report has been uploaded, we need this to differentiate between coverage and bundle analysis.

This copies the current `active` values to the new column, can do this because there is no upload for bundle analysis yet.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/1033

### Legal Boilerplate


barnett-yuxiang (commented):
PTAL

graphql_api/types/repository/repository.py
```diff
@@ -338,6 +338,11 @@ def resolve_bundle_analysis_enabled(repository: Repository, info) -> bool:
     return repository.bundle_analysis_enabled
 
 
+@repository_bindable.field("coverageEnabled")
+def resolve_coverage_enabled(repository: Repository, info) -> bool:
```
adrian-codecov, line 342: Since the return type is optional cause it can be null, can we add Optional to the return type?

core/models.py
```diff
@@ -130,6 +130,7 @@ class Languages(models.TextChoices):
     activated = models.BooleanField(null=True, default=False)
     deleted = models.BooleanField(default=False)
     bundle_analysis_enabled = models.BooleanField(default=False, null=True)
+    coverage_enabled = models.BooleanField(default=False, null=True)
```
adrian-codecov, line 133: So this won't have a DB default right, is that what you intend?

core/migrations/0046_repository_coverage_enabled.py
```diff
@@ -0,0 +1,33 @@
...
+        ("core", "0045_repository_languages_last_updated"),
+    ]
+
+    operations = [
+        RiskyAddField(
+            model_name="repository",
+            name="coverage_enabled",
+            field=models.BooleanField(default=False, null=True),
+        ),
+        RiskyRunSQL("UPDATE repos SET coverage_enabled=true WHERE active=true;"),
```
adrian-codecov, line 32: Did you confirm these run sequentially? Would bork the DB if you rand the update before the field is in there

graphql_api/types/repository/repository.py
JerrySentry: 👍

core/models.py
JerrySentry: Yeah, only ORM default, just to keep it consistent with the related fields (bundle_analysis_enabled, active, activated).

core/migrations/0046_repository_coverage_enabled.py
JerrySentry: Yep it runs sequentially on my local migrations run, I hope the ordering is not non-deterministic, that would be crazy. Didn't find in the Django docs that mentioned ordering, but there's many instances in the code that would imply this runs sequentially.

adrian-codecov (approved):

PR #342, opened 2024-01-16
feat: allow nulls in non-required params for BA upload endpoint

JerrySentry:
### Purpose/Motivation

Allow clients to send null values for the params that are not required. Null values are appropriately handled by the API.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/1052

### Legal Boilerplate


adrian-codecov (approved):
Should we leave a note explaining why?

PR #345, opened 2024-01-18
fix: if upload moves commit to other branch, possibly update branch model head

matt-codecov:
related to https://github.com/codecov/codecov-api/pull/305

that PR updates `commit.branch` but it doesn't touch the `Branch` model. `Branch` has a `head` and `base` field, and we may want to update them when moving a commit from one branch to another

as-is this PR will recompute the head of the destination branch, so if the commit being moved is newer than other commits on that branch, it'll become the new head. we may similarly want to fetch any branches that the commit being moved was the head of and recompute heads for those as well.

### Legal Boilerplate


upload/helpers.py
```diff
@@ -440,19 +440,34 @@ def insert_commit(commitid, branch, pr, repository, owner, parent_commit_id=None
...
-        commit.save(update_fields=["parent_commit_id", "state", "branch"])
+        commit.save(update_fields=edited)
+
+    # Update Branch.head object for the new branch
+    # if the current commit is newer than the existing Branch.head commit
+    if "branch" in edited:
+        new_branch = Branch.objects.filter(
+            name=branch, repository_id=repository.repoid
+        ).first()
+        if new_branch:
```
rohitvinnakota-codecov, line 463: Is there a scenario where this returns `None`?

rohitvinnakota-codecov (approved):

upload/helpers.py
JerrySentry: If they specify a branch value that doesn't exist, then we don't update the Branch model

PR #346, opened 2024-01-18
feat: GithubAppInstallation model

giovanni-guidini:
Model definitions for GithubAppInstallation.
This will unlock multiple apps for an org.
It also will improve how we keep track of what repos
an installation can access.

context: codecov/engineering-team#969

this is PR 1 of 2


codecov_auth/models.py
```diff
@@ -471,6 +475,48 @@ def remove_admin(self, user):
         self.save()
 
 
+GITHUB_APP_INSTALLATION_DEFAULT_NAME = "codecov_app_installation"
+
+
+class GithubAppInstallation(
```
matt-codecov, line 481: do we also need an app id column? since we may use sentry's app or support multiple apps

codecov_auth/models.py
```diff
@@ -471,6 +475,48 @@ def remove_admin(self, user):
...
+    ExportModelOperationsMixin("codecov_auth.github_app_installation"), BaseCodecovModel
+):
+
+    # replacement for owner.integration_id
+    # installation id GitHub sends us in the installation-related webhook events
+    installation_id = models.IntegerField(null=False, blank=False)
+    name = models.TextField(default=GITHUB_APP_INSTALLATION_DEFAULT_NAME)
```
matt-codecov, line 488: what is this for?

codecov_auth/tests/unit/test_models.py
```diff
@@ -463,3 +463,50 @@ def test_token_is_deleted_when_changing_user_plan(
...
+        assert list(owner.github_app_installations.all()) == [installation_obj]
+        assert installation_obj.repository_queryset().exists()
+        assert set(installation_obj.repository_queryset().all()) == set(
+            [repo1, repo2, repo3]
+        )
+
+    def test_covers_some_repos(self):
+        owner = OwnerFactory()
+        repo = RepositoryFactory(author=owner)
+        other_repo_different_owner = RepositoryFactory()
```
matt-codecov, line 497: can you add a diff_repo_same_owner case too

matt-codecov (approved):
i like the helper methods, nice touch

pls address comments but this'll defo do the trick

codecov_auth/models.py
giovanni-guidini: the "installation_id" is the "app_id"

codecov_auth/models.py
giovanni-guidini: I think this is what will differentiate the type of apps. As you mentioned above, we might use sentry's app or support multiple apps. This is how we know which is which.

PR #347, opened 2024-01-18
feat: use GithubAppInstallation in api

giovanni-guidini:
Uses the gh app installation WITHOUT compromising
the old usage (of `owner.integration_id`).
Deprecated paths are marked so.

These changes don't backfill old data, but start using the new model with webhooks.

depends on: codecov/codecov-api#346

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


utils/github.py
```diff
@@ -4,5 +4,5 @@
 
 
 @cache.cache_function(ttl=480)
-def get_github_integration_token(service, integration_id=None):
-    return _get_github_integration_token(service, integration_id=integration_id)
+def get_github_integration_token(service, installation_id=None):
```
adrian-codecov, line 7: Ty for this renaming!

webhook_handlers/views/github.py
```diff
@@ -389,7 +397,44 @@ def pull_request(self, request, *args, **kwargs):
 
         return Response()
 
-    def _handle_installation_events(self, request, *args, **kwargs):
+    def _handle_installation_repository_events(self, request, *args, **kwargs):
```
adrian-codecov, line 400: Fn looks good, but wdyt about rearranging things a bit so they're used only when they're needed? Like this (same appliies for other fns)

``` python
def _handle_installation_repository_events(self, request, *args, **kwargs):
    # https://docs.github.com/en/webhooks/webhook-events-and-payloads#installation_repositories
    service_id = request.data["installation"]["account"]["id"]
    username = request.data["installation"]["account"]["login"]
    owner, _ = Owner.objects.get_or_create(
        service=self.service_name, service_id=service_id, username=username
    )
    ghapp_installation: Optional[
        GithubAppInstallation
    ] = owner.github_app_installations.filter(
        installation_id=installation_id
    ).first()
    
    if ghapp_installation is None:
        installation_id = request.data["installation"]["id"]
        ghapp_installation = GithubAppInstallation(
            owner=owner, installation_id=installation_id
        )
    
    all_repos_affected = request.data.get("repository_selection") == "all"
    if all_repos_affected:
        ghapp_installation.repository_service_ids = None
    else:
        repo_list_to_save = set(ghapp_installation.repository_service_ids or [])
        repositories_added_service_ids = set(
            map(lambda obj: obj["id"], request.data.get("repositories_added", []))
        )
        repositories_removed_service_ids = set(
            map(lambda obj: obj["id"], request.data.get("repositories_removed", []))
        )
        repo_list_to_save = repo_list_to_save.union(
            repositories_added_service_ids
        ).difference(repositories_removed_service_ids)
        ghapp_installation.repository_service_ids = list(repo_list_to_save)
    ghapp_installation.save()
```

adrian-codecov (commented):
Some minor comments, otherwise lgtm

webhook_handlers/views/github.py
giovanni-guidini: Ah yes that is better indeed. Thanks for this sugestion

upload/helpers.py
```diff
@@ -315,20 +322,45 @@ def determine_upload_pr_to_use(upload_params):
         return upload_params.get("pr")
 
 
+def ghapp_installation_id_to_use(repository: Repository) -> Optional[str]:
```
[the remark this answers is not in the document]
giovanni-guidini: I don't think this code should live as a method of `GithubAppInstallation`.

I think there's enough logic not related to `GithubAppInstallation` to leave it as is.
Plus in the future we might use this function to house logic to pick one `GithubAppInstallation` out of many that a user might have.

webhook_handlers/views/github.py
adrian-codecov: np!

upload/helpers.py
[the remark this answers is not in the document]
adrian-codecov: That makes sense to me! Tyty

adrian-codecov (approved):

PR #352, opened 2024-01-23
fix: fetch accurate branch head

joseph-sentry:
### Purpose/Motivation
This commit is a fix for the recent issue where
branch heads were all set to the same value each
time they were updated due to a bug in the Branch
django model.

This fix creates a helper function that checks for the 2 remaining innacurate commit shas that a large number of branches were set to. If the current branch head is set to one of those shas, we update the branch head by fetching the latest commit on that branch from the db.

Note that we must use a raw sql query here to update the branch object because using the django ORM to update it would lead to the same issue we are trying to fix.

### Incident to link to
https://status.codecov.com/incidents/z10fykg3c9w4

### What does this PR do?
- create `get_or_update_branch_head`
- use `get_or_update_branch_head` function in many endpoints most notably: graphs (badges) and the graphql branch endpoint

### DIFF
--- a/api/public/v2/branch/serializers.py
+++ b/api/public/v2/branch/serializers.py
@@ -2,6 +2,7 @@ from rest_framework import serializers
 
 from api.public.v2.commit.serializers import CommitDetailSerializer
 from core.models import Branch, Commit
+from utils.temp_branch_fix import get_or_update_branch_head
 
 
 class BranchSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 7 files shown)


api/public/v2/branch/serializers.py
```diff
@@ -19,6 +20,8 @@ class BranchDetailSerializer(BranchSerializer):
     )
 
     def get_head_commit(self, branch: Branch) -> CommitDetailSerializer:
+        _ = get_or_update_branch_head(Commit.objects, branch, branch.repository_id)
```
thomasrockhu-codecov, line 23:
```suggestion
        get_or_update_branch_head(Commit.objects, branch, branch.repository_id)
```

giovanni-guidini (approved):
please create an issue to revisit this implementation and remove if we don't need it anymore so we don't forget about this.

PR #357, opened 2024-01-24
fix: add unique constraint on test model

joseph-sentry:
### Purpose/Motivation
The outcome field should be an enum stored as a string.

The env field should renamed to flags_hash for clarity.

The uniqueconstraint was added to ensure that Test objects are unique according to repo, name, testsuite and flags_hash.

### What does this PR do?
- Add unique constraint on aforementioned fields on Test model
- Create Outcome enum for TestInstance outcome field
- Change outcome to string and use enum
- Change env field to flags_hash field
- Create Django migration


reports/migrations/0014_rename_env_test_flags_hash_and_more.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 4.2.7 on 2024-01-24 22:47
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
```
adrian-codecov, line 7: Could you pls ad the SQL code for this migration here? 🙇

reports/models.py
```diff
@@ -238,10 +239,20 @@ class Test(models.Model):
     )
     name = models.TextField()
     testsuite = models.TextField()
-    env = models.TextField()
+    # this is a hash of the flags associated with this test
+    # users will use flags to distinguish the same test being run
+    # in a different environment
+    # for example: the same test being run on windows vs. mac
+    flags_hash = models.TextField()
```
adrian-codecov, line 246: Oh interesting, is this the same flag as a codecov flag?

reports/models.py
```diff
@@ -224,6 +224,7 @@ class Test(models.Model):
     # in this case is because we want to be able to compute/predict
     # the primary key of a Test object ourselves in the processor
     # so we can easily do concurrent writes to the database
+    # this is a hash of the repoid, name, testsuite and env
```
adrian-codecov, line 227: Interesting, so the id itself will account for the env now 👀

reports/models.py
joseph-sentry: This is a hash of the flags associated with the uploads that map to this test.

The idea is that users will use flags with their uploads to differentiate between the same test being run in a different environment (this used to be the env field but now it's the flags_hash field).

reports/models.py
joseph-sentry: i should rename it to `flags_hash` in the comment here as well but yes, because a test is unique on the flags_hash as well, we will include it in the hash

reports/models.py
adrian-codecov: Sounds good. I'd expect this field to be nullable too right?

adrian-codecov (approved):

barnett-yuxiang (commented):
LGTM

PR #358, opened 2024-01-24
fix: branch switch on upload

JerrySentry:
### Purpose/Motivation

related to https://github.com/codecov/codecov-api/pull/305

that PR updates commit.branch but it doesn't touch the Branch model. Branch has a head and base field, and we may want to update them when moving a commit from one branch to another

as-is this PR will recompute the head of the destination branch, so if the commit being moved is newer than other commits on that branch, it'll become the new head. we may similarly want to fetch any branches that the commit being moved was the head of and recompute heads for those as well.

### Verification

Done locally:
1. Create a commit on a feature branch `jan_24_zz` with a commit hash `6fcbe9..`
Screenshot 2024-01-24 at 4 43 23 PM

2. Make a request to upload endpoint with branch switch to `main` on the same commit
Screenshot 2024-01-24 at 4 44 23 PM

3. Branch table is correctly updated, while the other repo's (`codecov-api`) `main` branch remains unchanged.
Screenshot 2024-01-24 at 4 44 55 PM

4. The Commit table is correctly updated for the commit to move to `main`
Screenshot 2024-01-24 at 4 49 26 PM

### Legal Boilerplate


upload/helpers.py
```diff
@@ -440,19 +441,42 @@ def insert_commit(commitid, branch, pr, repository, owner, parent_commit_id=None
...
+        commit.save(update_fields=edited)
+
+    # Update Branch.head object for the new branch
+    # if the current commit is newer than the existing Branch.head commit
+    if "branch" in edited:
+        new_branch = Branch.objects.filter(
+            name=branch, repository_id=repository.repoid
+        ).first()
+        if new_branch:
+            new_branch_head = Commit.objects.filter(
```
joseph-sentry, line 465: i would just rename this as new_branch_curr_head for clarity

upload/helpers.py
```diff
@@ -440,19 +441,45 @@ def insert_commit(commitid, branch, pr, repository, owner, parent_commit_id=None
...
+                repository_id=repository.repoid,
+            ).first()
+            if (
+                new_branch_curr_head is None
+                or new_branch_curr_head.timestamp < commit.timestamp
+            ):
+                # using this raw sql because the current branches table does not allow for updating based on repoid
+                # it only updates based on branch name which means if we were to use the django orm to update this
+                # branch.head, all branches with the same name as the one we are trying to update would have their head
+                # updated to the value of this ones head
```
matt-codecov, line 476: nit, but it's not the table that blocks this, it's the ORM that doesn't support composite primary keys

matt-codecov (approved):

PR #360, opened 2024-01-25
reactivate ghapp usage in api

giovanni-guidini:
reactivate ghapp usage in api and fix bug when getting file contents.

This bug caused a little incident in Codecov.

The actual issue is that we can't make DB queries syncronously from an
async context.

When getting the file content we have to get an adapter,
and that - since the GithubAppInstallation work - required a query to
the database to get the default GithubAppInstallation for an `owner`.
That was the actual error that caused the incident.

Error surfacing was terrible in that endpoint so I changed the `info` log
from an `exception` one so we can have a stacktrace.

To actually fix the error I actually created an async version of `get_adapter`.
I scanned the code, but it seems that this instance is the only place where we
actually use this.

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


services/repo_providers.py
```diff
@@ -86,8 +95,60 @@ def get_provider(service, adapter_params):
...
+        Service.GITHUB.value,
+        Service.GITHUB_ENTERPRISE.value,
+    ]:
+        return None
+    return owner.github_app_installations.filter(
+        name=GITHUB_APP_INSTALLATION_DEFAULT_NAME
+    ).first()
+
+
+async def async_get_ghapp_default_installation(
```
adrian-codecov, line 111: We often fix async/sync issues with the `@sync_to_async` or `@async_to_sync` decorators. Did you give that a shot as well? Curious if that would have fixed it

giovanni-guidini: Making `get_adapter` async didn't work cause it's the same issue (sync only op running in async loop)
Making `is_using_integration` async also didn't work (same as above)

I didn't try to make the interactor sync. I assumed it was async for a reason and decided not to touch it.

I did try some other helper functions with those wrappers but couldn't find a combination that worked. So I gave up and decided to split the function into two 🤷

matt-codecov (approved):

PR #362, opened 2024-01-26
feat: support force flag in empty upload endpoint:

joseph-sentry:
### Purpose/Motivation
Users may want to bypass the files check of the empty upload endpoint and force a passing notification

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1015
CLI change: https://github.com/codecov/codecov-cli/pull/362

### What does this PR do?
Accepts a should_force boolean field in the data for this endpoint, if it's true then we send a passing notification else we do the regular behaviour of this endpoint

### DIFF
--- a/upload/views/empty_upload.py
+++ b/upload/views/empty_upload.py
@@ -75,13 +75,23 @@ class EmptyUploadView(CreateAPIView, GetterMixin):
         if not serializer.is_valid():
             return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
         data = serializer.validated_data
-        if "should_force" in data:
-            should_force = data["should_force"]
-        else:
-            should_force = False
+        should_force = data.get("should_force", False)
 
         repo = self.get_repo()
         commit = self.get_commit(repo)
+
+        if should_force is True:
+            TaskService().notify(
+                repoid=repo.repoid, commitid=commit.commitid, empty_upload="pass"
+            )
+            return Response(
+                data={
+                    "result": "Force option was enabled. Triggering passing notifications.",
+                    "non_ignored_files": [],
+                },
+                status=status.HTTP_200_OK,
+            )
+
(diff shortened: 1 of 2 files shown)


upload/views/empty_upload.py
```diff
@@ -91,7 +104,18 @@ def post(self, request, *args, **kwargs):
             if any(map(lambda regex: regex.match(file), compiled_files_to_ignore))
         ]
 
-        if set(changed_files) == set(ignored_changed_files):
+        if should_force is True:
```
giovanni-guidini, line 107: Maybe move this before you do the `changed_files` scan? No point in doing that if you're going to force the notif regardless

upload/views/empty_upload.py
```diff
@@ -67,6 +71,15 @@ class EmptyUploadView(CreateAPIView, GetterMixin):
     ]
 
     def post(self, request, *args, **kwargs):
+        serializer = EmptyUploadSerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+        data = serializer.validated_data
+        if "should_force" in data:
```
giovanni-guidini, line 78: This is VERY nit, but doesn't `should_force = data.get("should_force", False)` accomplishes the same?

upload/views/empty_upload.py
```diff
@@ -67,8 +71,27 @@ class EmptyUploadView(CreateAPIView, GetterMixin):
     ]
 
     def post(self, request, *args, **kwargs):
+        serializer = EmptyUploadSerializer(data=request.data)
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
```
giovanni-guidini, line 76: It would be a good idea to add a test for this case even if it's simple...

giovanni-guidini (approved):

PR #379, opened 2024-02-07
fix: Update available plan to remove trial condition

RulaKhaled:
### Purpose/Motivation
We want to roll out the team plan to all orgs that has less than 11 activated users.  

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1122

### What does this PR do?
- Remove trial conditions to serve this plan to basic and free users 
- Fix tests 

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/plan/service.py
+++ b/plan/service.py
@@ -126,18 +126,12 @@ class PlanService:
         if owner and sentry.is_sentry_user(owner=owner):
             available_plans += SENTRY_PAID_USER_PLAN_REPRESENTATIONS.values()
 
-        # If user is already in team plan or is/have trialed
-        if (
-            self.plan_name in TEAM_PLAN_REPRESENTATIONS
-            or self.trial_status == TrialStatus.EXPIRED.value
-            or self.trial_status == TrialStatus.ONGOING.value
-        ):
             # If number of activated users is less than or equal to TEAM_PLAN_MAX_USERS
-            if (
-                self.plan_activated_users is None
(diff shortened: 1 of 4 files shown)


plan/service.py
```diff
@@ -126,18 +126,12 @@ def available_plans(self, owner: Owner) -> List[PlanData]:
         if owner and sentry.is_sentry_user(owner=owner):
             available_plans += SENTRY_PAID_USER_PLAN_REPRESENTATIONS.values()
```
JerrySentry, line 128: Should we be using `plan_activated_users` or `plan_user_count`? Say they have purchased 100 seats but only 10 have been activated, do we want to show Team plan as an option?

We had to switch to `plan_activated_users` from `plan_user_count` because trial plan sets the `plan_user_count` to some fixed large number, so we just used `plan_activated_users`. Now that we want to surface this plan to non-trial plans, how do we want to handle this?

RulaKhaled: We still wanna go with activated users, here's what AJ said:

> I believe we want to stick with plan activated users - my thought process on this is that this is who is actually using the product, if you only have 10 users I’d want them to be on the most economically sensible plan for them, even if it means a downgrade from pro to team - same logic applies when canceling, even if they reduce # of users AND downgrade to Team that’s still preferable to the alternative of giving them the less activated save offer (30% off pro) and having them maybe choose Team

JerrySentry (approved):

PR #388, opened 2024-02-09
feat: include repos affected when sync_repos from installation

giovanni-guidini:
Does nothing in practice until https://github.com/codecov/worker/pull/259
(but does not depend on it for deploy)

These changes pass the repos covered by an installation to sync_repos.
This can speed up the sync_repos time because we know what repos to sync.

ps.: changes to `uploads_used.py` were because my tests were failing


nora-codecov (approved):

barnett-yuxiang (commented):
Is it possible to configure static code rule checking?

PR #393, opened 2024-02-12
Add customer intent to Users type and SaveTermsAgreement mutation

rohitvinnakota-codecov:
### Purpose/Motivation

Following up from https://github.com/codecov/codecov-api/pull/390, updates the SaveTermsAgreement to read the customer intent value from the front end. It should always be provided. 

Additionally, add a resolver to fetch the intent on the User object. 

### Links to relevant tickets

This closes https://github.com/codecov/engineering-team/issues/1090

### Legal Boilerplate


codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -30,6 +32,13 @@ def update_terms_agreement(self, input: TermsAgreementInput):
         self.current_user.terms_agreement_at = timezone.now()
         self.current_user.save()
 
+        OwnerProfile.objects.update_or_create(
```
adrian-codecov, line 35: After our conversation in Slack, we wouldn't need this and instead would be `self.current_user.customer_intent = input.customer_intent`

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -48,6 +57,7 @@ def execute(self, input):
             business_email=input.get("businessEmail"),
             terms_agreement=input.get("termsAgreement"),
             marketing_consent=input.get("marketingConsent"),
```
adrian-codecov, line 59: Yeah this one is weird 😂 we're not using it, @RulaKhaled pinging you since you were the last to work on this, do we know what's the intended use for this?

RulaKhaled: We used to use this to decide if user signs up for marketing consent or not, to decide if we should opt user in for emails

RulaKhaled: im not sure if we use it anymore though

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -29,6 +30,7 @@ def update_terms_agreement(self, input: TermsAgreementInput):
         self.current_user.terms_agreement = input.terms_agreement
         self.current_user.terms_agreement_at = timezone.now()
         self.current_user.save()
+        self.current_user.customer_intent = input.customer_intent
```
JerrySentry, line 33: `customer_intent` can only be business or personal right? should validate that.

graphql_api/types/mutation/save_terms_agreement/save_terms_agreement.graphql
```diff
@@ -8,4 +8,5 @@ input SaveTermsAgreementInput {
   businessEmail: String
   termsAgreement: Boolean!
   marketingConsent: Boolean
+  customerIntent: String!
```
JerrySentry, line 11: so when they call this mutation, they have to provide `customerIntent`, they can call this to update other things right?

graphql_api/types/user/user.py
```diff
@@ -39,3 +39,9 @@ def resolve_student_created_at(user: Owner, info) -> Optional[datetime]:
 @user_bindable.field("studentUpdatedAt")
 def resolve_student_updated_at(user: Owner, info) -> Optional[datetime]:
     return user.student_updated_at
+
+
+@user_bindable.field("customerIntent")
+def resolve_customer_intent(user: Owner, info) -> Optional[datetime]:
```
JerrySentry, line 45: return typing

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -14,6 +14,7 @@ class TermsAgreementInput:
     business_email: Optional[str] = None
     terms_agreement: bool = False
     marketing_consent: bool = False
```
ajay-sentry, line 16: Just curious (and also a Python noob here), but on line 16 isn't the bool implied through the default assignment?
@JerrySentry @rohitvinnakota-codecov

codecov_auth/commands/owner/interactors/save_terms_agreement.py
adrian-codecov: Good catch!

codecov_auth/commands/owner/interactors/save_terms_agreement.py
adrian-codecov: Might be worth following up with them, not a blocker for this PR though

graphql_api/types/mutation/save_terms_agreement/save_terms_agreement.graphql
adrian-codecov: This form has 2 required things: termsAgreement + customerIntent, other fields are optional. We should validate we have all required fields to call this mutation successful + validate values of those if necessary, like with `customerIntent`

graphql_api/types/user/user.py
```diff
@@ -39,3 +39,9 @@ def resolve_student_created_at(user: Owner, info) -> Optional[datetime]:
...
+    owner = user
+    return owner.user.customer_intent
```
adrian-codecov, line 47: Wait this is odd, why is this file called user.py, all users are `Owners`, the `me` resolver creates a custom `User` type which is certain fields from the Owner object? Pretty odd, probably some legacy thing we want wanted to differentiate different use cases for it.

Do we not have a type that properly maps to the current_user/`User` class?

codecov_auth/commands/owner/interactors/save_terms_agreement.py
ajay-sentry: Got it, that totally makes sense. Only follow up I have here is do we have a stance between when we're looking to add explicit typing vs. not?

adrian-codecov: I would say rule of thumb is always try and add types to your fn params + return values.

I personally like to type my variables if I'm using a class, like with Person above, and create dataclasses in scenarios where I need to create a helper dictionary instead

codecov_auth/commands/owner/interactors/save_terms_agreement.py
rohitvinnakota-codecov: Will there be any implicit validation provided by `models.TextChoices`?

graphql_api/types/user/user.py
rohitvinnakota-codecov: I thought it was odd as well. We changed our owner/user model a while back so I'm not sure if that is related.

codecov_auth/commands/owner/interactors/save_terms_agreement.py
ajay-sentry: Sounds good! Will keep that in mind.

codecov_auth/commands/owner/interactors/save_terms_agreement.py
JerrySentry: Wait the `customer_intent` is not being saved

graphql_api/types/user/user.py
adrian-codecov: Mmmm, don't want to bloat your PR but I think there's a legit use case to have a User resolver. We can do it now or later, I'll leave that up to you 👌

codecov_auth/commands/owner/interactors/save_terms_agreement.py
adrian-codecov: Also true. I would have to test it myself, but I suspect when you try to do safe with a choice not belonging to text choices, we'd get an error

graphql_api/types/user/user.py
rohitvinnakota-codecov: Created an issue here
https://github.com/codecov/engineering-team/issues/1220

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -22,12 +23,20 @@ class SaveTermsAgreementInteractor(BaseInteractor):
     def validate(self, input: TermsAgreementInput):
         if input.terms_agreement is None:
             raise ValidationError("Terms of agreement cannot be null")
+        if input.customer_intent not in [
```
RulaKhaled, line 26: Let's do:
`if input.customer_intent and input.customer_intent not in [...`

to enforce a default to the field so it does not break for current users.

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -14,6 +14,7 @@ class TermsAgreementInput:
     business_email: Optional[str] = None
     terms_agreement: bool = False
     marketing_consent: bool = False
+    customer_intent: str = None
```
RulaKhaled, line 17: let's add optional

To be fair, i don't think this should ever be a blocker for onboarding, the page is a blocker for users from accessing the dashboard.

graphql_api/types/mutation/save_terms_agreement/save_terms_agreement.graphql
RulaKhaled, line 11: Let's remove `!`

RulaKhaled: @adrian-codecov I think this should be optional, it makes sense not to allow users form accessing dashboard if TOS is not approved, but even though we opt users in to choose customer intent, it should never be a blocker from accessing the dashboard / should not cause server error if it is missing, the mutation will handle this successfully in Gazebo as expected, let's allow client to control this field more and default things to `None`. This also means that we don't have to do risky deployments with gaps between client and API.

codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -22,12 +23,20 @@ class SaveTermsAgreementInteractor(BaseInteractor):
     def validate(self, input: TermsAgreementInput):
         if input.terms_agreement is None:
             raise ValidationError("Terms of agreement cannot be null")
+        if input.customer_intent and input.customer_intent not in [
+            "Business",
```
adrian-codecov, line 27: Any reason why we don't lowercase/uppercase the input to follow what we have in the DB?

adrian-codecov (approved):

codecov_auth/commands/owner/interactors/save_terms_agreement.py
rohitvinnakota-codecov: I am validating to keep parity with the [choices](https://github.com/codecov/codecov-api/blob/main/codecov_auth/migrations/0051_user_customer_intent.py) here.

PR #403, opened 2024-02-16
chore: remove rollout check on test result endpoint

joseph-sentry:


barnett-yuxiang (commented):
LGTM

utils/rollouts.py
```diff
@@ -5,18 +5,3 @@
 
 def owner_slug(owner: Owner) -> str:
```
daniel-codecov, line 6: should this get deleted too?

joseph-sentry:
> Just curious, why are we removing this rollout? Do we just want it enabled for everyone now?

yeah, we were running into problems where people would want to try it out but they would get blocked on this

utils/rollouts.py
joseph-sentry: I think it's fine to leave this here, it might end up being useful next time someone wants to use a rollout

giovanni-guidini: I don't think so... that is a helper function, not directly related to this rollout. We can reuse that for future rollouts.
(even if there aren't none right now 😅)

giovanni-guidini (approved):

PR #404, opened 2024-02-16
feat: add user agent to prom metrics

daniel-codecov:
### Purpose/Motivation

This change adds the User-Agent http header to our prometheus metrics that we expose so that we can have a better understanding of where traffic comes from for our API. We can visualize the data via Grafana by the different User-Agents and it could give us insights in how our users use the codecov cli, uploader, etc.. 

### What does this PR do?
- subclass the provided django_prometheus middlewares, but add the additional label when appropriate
- add unit tests to ensure functionality

### Notes to Reviewer

You can view the exposed metrics at http://localhost:8000/monitoring/metrics, but I wasn't able to try querying in Grafana. Do we need to deploy to staging to test that? 

### Legal Boilerplate

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -11,7 +11,7 @@ from django_prometheus.middleware import (
 
 from utils.services import get_long_service_name
 
-# Prometheus metrics that will additional be annotated with User-Agent http header as label
+# Prometheus metrics that will be annotated with User-Agent http header as label
 USER_AGENT_METRICS = [
     "django_http_requests_unknown_latency_including_middlewares_total",
     "django_http_requests_latency_seconds_by_view_method",
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -11,25 +11,7 @@ from django_prometheus.middleware import (
 
 from utils.services import get_long_service_name
 
-
-def get_service_long_name(request: HttpRequest) -> Optional[str]:
-    resolver_match = resolve(request.path_info)
-    service = resolver_match.kwargs.get("service")
-    if service is not None:
-        resolver_match.kwargs["service"] = get_long_service_name(service.lower())
-        service = get_long_service_name(service.lower())
-        return service
-    return None
-
-
-class ServiceMiddleware(MiddlewareMixin):
-    def process_view(self, request, view_func, view_args, view_kwargs):
-        service = get_service_long_name(request)
-        if service:
-            view_kwargs["service"] = service
-        return None
-
-
+# Prometheus metrics that will additional be annotated with User-Agent http header as label
 USER_AGENT_METRICS = [
     "django_http_requests_unknown_latency_including_middlewares_total",
     "django_http_requests_latency_seconds_by_view_method",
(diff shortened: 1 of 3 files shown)


core/middleware.py
```diff
@@ -3,9 +3,34 @@
...
+from django_prometheus.middleware import (
+    Metrics,
+    PrometheusAfterMiddleware,
+    PrometheusBeforeMiddleware,
+)
 
 from utils.services import get_long_service_name
 
+# Prometheus metrics that will additional be annotated with User-Agent http header as label
+USER_AGENT_METRICS = [
```
daniel-codecov, line 15: not sure what is the best way to choose these metrics is. I went through the list here, and chose the ones that didn't get an error: link[https://github.com/korfuri/django-prometheus/blob/e96708873c7aaf49f5b571cb08804aaecb45cc9d/django_prometheus/middleware.py#L24](url)

daniel-codecov: django_prometheus likes to error when you assign a label to a metric and the corresponding label value does not get populated

core/middleware.py
```diff
@@ -23,3 +48,39 @@ def process_view(self, request, view_func, view_args, view_kwargs):
...
+    """
+    django_prometheus Metrics class but with extra user_agent label for applicable metrics
+    """
+
+    def register_metric(self, metric_cls, name, documentation, labelnames=(), **kwargs):
+        if name in USER_AGENT_METRICS:
+            labelnames = list(labelnames) + ["user_agent"]
+        return super().register_metric(
+            metric_cls, name, documentation, labelnames=labelnames, **kwargs
+        )
```
matt-codecov, line 63: i think your code is right, but it made me notice a couple interesting things so indulge me

compare your approach here to the approach taken in the django-prometheus test app:
```
def register_metric(..., labelnames=(), ...):
    if name in EXTENDED_METRICS:
        labelnames.extend(("view_type", "user_agent_type"))
    return return super().register_metric(...)
```
their approach modifies the passed-in `labelnames` value so whoever calls `register_metric()` would be able check the list they passed in after the function returns and see which labels were added. your approach reassigns `labelnames` to a new list and leaves the passed-in value alone. see what i mean: https://www.online-python.com/lOkDiMp4Kj

looking at how `register_metric()` is used ([example](https://github.com/korfuri/django-prometheus/blob/e96708873c7aaf49f5b571cb08804aaecb45cc9d/django_prometheus/middleware.py#L56)), it doesn't look like anything relies on the passed-in list being updated to contain the new labels. so your approach is fine

the other thing is, it's weird that the default value for `labelnames` is an empty tuple. the passed-in values are all lists so i'd expect the default to be an empty list. also tuples don't have an `.extend()` method so their test app actually requires `labelnames` to be a list :v but it's like that in `django-prometheus` so following suit here is reasonable

core/tests/test_middleware.py
```diff
@@ -0,0 +1,131 @@
+from django.test import TestCase
+from prometheus_client import REGISTRY
+
+from core.middleware import USER_AGENT_METRICS
+
+
+# TODO: consolidate with worker/helpers/tests/unit/test_checkpoint_logger.py into shared repo
```
matt-codecov, line 7: 👍  nice

matt-codecov (approved):
awesome work, thanks a ton!

`codecov-api` automatically deploys when you merge to `main` so we should wait until monday to merge this in case something goes wrong. wouldn't want to make anyone work late on a friday

also might be a good idea to push to staging first and click around to see if anything crashes. and if there's a staging instance of grafana we can test it end-to-end with a dashboard too

core/middleware.py
daniel-codecov: Interesting, yeah I didn't give the modification vs new list much thought tbh

I'm not actually sure if changing default value of param will mess up the override declaration, but I feel like tuple as default value is just a mistake on their end lol

PR #406, opened 2024-02-17
feat: measure uploads using sentry metrics

joseph-sentry:
### Purpose/Motivation
Adding Sentry DDM to the various upload endpoints

### Links to relevant tickets
Fixes: https://github.com/codecov/engineering-team/issues/1178

### What does this PR do?
- `generate_upload_sentry_metrics_tags` in upload helpers to generate tags for metrics, tags are:
  - agent: uploader or cli
  - version: which version of the agent
  - action: coverage vs test results
  - private: is the repo the upload is being made to private?
  - is_shelter_request: is the request coming from shelter
- `get_agent_from_headers` and `get_version_from_headers` to parse `User-Agent` header to get agent and version, if it fails to parse it will return `unsupported-user-agent`

Note: My auto formatter went off while saving some of the files and the linter did not undo them, so this PR comes with some formatting changes to some of the files that have not been visited recently.

### DIFF
--- a/upload/helpers.py
+++ b/upload/helpers.py
@@ -175,9 +175,7 @@ def parse_params(data):
                     else (
                         value[7:]
                         if value[:7] == "origin/"
-                        else value[11:]
-                        if value[:11] == "refs/heads/"
-                        else value
+                        else value[11:] if value[:11] == "refs/heads/" else value
                     )
                 ),
             ),
@@ -753,6 +751,7 @@ def validate_activated_repo(repository):
         )
 
 
+# headers["User-Agent"] should look something like this: codecov-cli/0.4.7 or codecov-uploader/0.7.1
 def get_agent_from_headers(headers):
     try:
         return headers["User-Agent"].split("/")[0].split("-")[1]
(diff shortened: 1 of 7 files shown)


upload/helpers.py
```diff
@@ -739,3 +751,42 @@ def validate_activated_repo(repository):
...
+    try:
+        return headers["User-Agent"].split("/")[0].split("-")[1]
+    except Exception as e:
+        log.warning(
+            "Error getting agent from user agent header",
+            extra=dict(
+                err=str(e),
+            ),
+        )
+        return "unsupported-user-agent"
```
giovanni-guidini, line 766: [nit] maybe "unknown" is better than "unsupported" here IF (and only if, I actually don't know) we actually process the upload

upload/helpers.py
```diff
@@ -739,3 +751,42 @@ def validate_activated_repo(repository):
...
+    try:
+        return headers["User-Agent"].split("/")[1]
+    except Exception as e:
+        log.warning(
+            "Error getting version from user agent header",
+            extra=dict(
+                err=str(e),
+            ),
+        )
+        return "unsupported-user-agent"
```
giovanni-guidini, line 780: Same nit suggestion as function above

giovanni-guidini (approved):
some nit suggestions but looks good 👍

barnett-yuxiang (commented):
Is this version management more dangerous?

PR #408, opened 2024-02-21
Update Bundle analysis Schema and add GQL cleanup step

JerrySentry:
### Purpose/Motivation

Make a teardown step for GQL request to clean up `/tmp` directory. This is needed for BA because we load the SQLite file (used as a storage for bundle stats) from GCS into the containers file system in the `/tmp` dir, we need to delete that file at the end when the entire GQL request is resolved.

Note: Update Shared commit once that is push https://github.com/codecov/shared/pull/129

Relevant ticket: https://github.com/codecov/engineering-team/issues/1068

### Legal Boilerplate


services/bundle_analysis.py
```diff
@@ -109,8 +109,8 @@ def __init__(self, report: SharedBundleAnalysisReport):
         self.cleanup()
 
     def cleanup(self) -> None:
```
JerrySentry, line 111: Don't do the full `cleanup` since that will delete the DB temp file prematurely.

adrian-codecov: Do we still need the cleanup at this level? I feel having a cleanup here might be more confusing than not, since all connection related stuff should be done at the parent (shared) level

graphql_api/views.py
```diff
@@ -50,7 +93,8 @@ async def post(self, request, *args, **kwargs):
         log.info("GraphQL Request", extra=log_data)
 
         # request.user = await get_user(request) or AnonymousUser()
-        return await super().post(request, *args, **kwargs)
+        with RequestFinalizer():
+            return await super().post(request, *args, **kwargs)
```
adrian-codecov, line 97: What's a way for us to see/measure a connection being closed through this code?

services/bundle_analysis.py
JerrySentry: We still need this for now, but you're right it shouldn't have to do this on this level. Reason being on the `shared.BundleAnalysisReport` it is still relying on the implementation of manually closing connections. I didn't want to touch that in the current shared PR because `BundleAnalysisReport` is too intertwined with worker as well. But yeah next step I need to make some changes to `BundleAnalysisReport` class

graphql_api/views.py
JerrySentry: What do you mean? All requests comes to this code.

barnett-yuxiang (commented):
LGTM

graphql_api/views.py
adrian-codecov: Like how do we know this code indeed closed a connection after the fact, is there a utility or metric that shows live connection before, during and after?

PR #410, opened 2024-02-22
Add user check to customer intent resolver

rohitvinnakota-codecov:
### Purpose/Motivation

Fixes these issues on [Sentry](https://codecov.sentry.io/issues/?alert_rule_id=1071529&alert_timestamp=1708624187882&alert_type=email&environment=production&notification_uuid=6be186ed-b075-4864-8d15-3eabfcc007b7&project=5215654&query=is%3Aunresolved+%27customer_intent%27&referrer=issue-list&statsPeriod=30d&utc=true). The `user` subtype on the owner may be undefined so we will return `None` for the customer intent in that case. 

### Legal Boilerplate


JerrySentry (approved):

graphql_api/types/user/user.py
```diff
@@ -44,4 +44,6 @@ def resolve_student_updated_at(user: Owner, info) -> Optional[datetime]:
 @user_bindable.field("customerIntent")
 def resolve_customer_intent(user: Owner, info) -> str:
     owner = user
+    if not owner.user:
```
thomasrockhu-codecov, line 47: curious, why the misdirect `owner = user` above? Seems unnecessary

RulaKhaled: I'm guessing because he want to showcase that the user is of type owner and we're fetching current from there?

@rohitvinnakota-codecov you can also do:
```
def resolve_customer_intent(current_owner: Owner, info) -> str:
    return current_owner.user.customer_intent
```

rohitvinnakota-codecov: Yes, this is mostly for readability. Otherwise it would read `user.user...`.

We have an issue to improve [this](https://github.com/codecov/engineering-team/issues/1220) so hopefully things will be more clear in the future.

thomasrockhu-codecov: Got it, thanks!

RulaAbuHasna (approved):

PR #412, opened 2024-02-23
Bundle Analysis v0.5 - GQL refresh

JerrySentry:
Note: Update Shared commit once https://github.com/codecov/shared/pull/132 is pushed.

Add module and asset reporting for bundle analysis.

### Legal Boilerplate


services/bundle_analysis.py
```diff
@@ -42,6 +46,26 @@ def load_time_conversion(size):
     return round((8 * size) / (1024 * 1024 * 3), 1)
 
 
+def get_extension(filename: str) -> str:
```
adrian-codecov, line 49: Is this something that could be in shared? I'm trying to think what's a good etiquette we can adopt for knowing what should/shouldn't be in shared. 

For instance, I can see this being shared logic, but there's not an apparent usecase for it atm, but in the future it would be 2x work to get move this you know

JerrySentry: Yeah I agree it is hard to determine if it should be in shared or not. My thinking (and current implementation) is that any data fetching and data aggregation should be in shared, whereas any form of data manipulation/transformation for the purpose of displaying the values to end users should be on their own codebases.

I suppose whenever the use case for PR comments to display the same info as the UI comes, then it may make more sense at that time to migrate this to shared.

PR #413, opened 2024-02-23
feat: hookup feature models to django admin

daniel-codecov:
### Purpose/Motivation

https://github.com/codecov/codecov-api/assets/159859649/dfdb68ec-1f94-4221-be74-290045e51928

Adds the new feature models from [here](https://github.com/codecov/shared/pull/130) to django admin so that we can assign feature variants to features dynamically

### DIFF
--- a/codecov/urls.py
+++ b/codecov/urls.py
@@ -24,6 +24,16 @@ urlpatterns = [
     path(f"{settings.DJANGO_ADMIN_URL}/", admin.site.urls),
     path("staticanalysis/", include("staticanalysis.urls")),
     path("labels/", include("labelanalysis.urls")),
+    path(
+        "admin-repository-autocomplete/",
+        views.RepositoryAutoCompleteSearch.as_view(),
+        name="admin-repository-autocomplete",
+    ),
+    path(
+        "admin-owner-autocomplete/",
+        views.OwnerAutoCompleteSearch.as_view(),
+        name="admin-owner-autocomplete",
+    ),
     re_path(r"^redirect_app", views.redirect_app),
     # /monitoring/metrics will be a public route unless you take steps at a
     # higher level to null-route or redirect it.
--- a/codecov/views.py
+++ b/codecov/views.py
@@ -1,8 +1,10 @@
+from dal import autocomplete
 from django.conf import settings
 from django.db import connection
 from django.http import HttpResponse, HttpResponseRedirect
 
-from core.models import Constants
+from codecov_auth.models import Owner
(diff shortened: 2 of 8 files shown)


matt-codecov (commented):
could you upload a screenshot of what the webform looks like or something? very curious if it is actually usable OOTB or if we have to do some custom webform work to allow users to input things like "github/codecov" and have that translate to the database id for codecov's github org

codecov/urls.py
```diff
@@ -24,6 +24,16 @@
     path(f"{settings.DJANGO_ADMIN_URL}/", admin.site.urls),
     path("staticanalysis/", include("staticanalysis.urls")),
     path("labels/", include("labelanalysis.urls")),
+    path(
+        "admin-repository-autocomplete/",
+        views.RepositoryAutoCompleteSearch.as_view(),
+        name="admin-repository-autocomplete",
+    ),
+    path(
```
daniel-codecov, line 32: these endpoints will be exposed on the API but I think its fine cause of [this](https://github.com/codecov/codecov-api/blob/ca8de7b5d4334c1dd0122efab355d0ccfb91b328/codecov/views.py#L37)

codecov/views.py
```diff
@@ -28,3 +30,31 @@ def redirect_app(request):
     """
     app_domain = settings.CODECOV_DASHBOARD_URL
     return HttpResponseRedirect(app_domain + request.path.replace("/redirect_app", ""))
+
+
+class RepositoryAutoCompleteSearch(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        # must be authenticated to query
+        if not self.request.user.is_authenticated:
```
matt-codecov, line 38: i think if you do `self.request.user.is_staff` instead of `is_authenticated` it's just users who can access the admin interface

codecov/urls.py
matt-codecov: thanks for calling this out!

codecov/views.py
```diff
@@ -28,3 +30,31 @@ def redirect_app(request):
...
+        if self.q:
+            repos = repos.filter(name__icontains=self.q)
+
+        return repos
+
+
+class OwnerAutoCompleteSearch(autocomplete.Select2QuerySetView):
+    def get_queryset(self):
+        # must be authenticated to query
+        if not self.request.user.is_authenticated:
```
matt-codecov, line 52: same `is_staff` comment as above

matt-codecov (approved):

PR #416, opened 2024-02-24
chore: use debian base images instead of alpine

matt-codecov:
fixes https://github.com/codecov/engineering-team/issues/1242

with an alpine/musl image we have to build more packages from source. grpcio takes an eternity to build for example

also apparently alpine/musl images have DNS resolution problems

this uses an official "slim" python image based on debian stable. from a quick glance at uncompressed image sizes on my local machine, this is actually 300mb smaller:
Screenshot 2024-02-23 at 7 48 56 PM
Screenshot 2024-02-23 at 7 48 34 PM

i have not rigorously tested this but it seems okay

### Legal Boilerplate

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,7 +188,7 @@ test_env.install_cli:
 	pip install codecov-cli==$(CODECOV_CLI_VERSION)
 
 test_env.container_prepare:
-	apk add -U curl git build-base jq
+	apt-get -y install git build-essential netcat-traditional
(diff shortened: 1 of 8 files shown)


trent-codecov (changes requested):
This generally looks good. Can you change the shebang for prod.sh etc? Then lets merge this to staging and test it out further.

trent-codecov (approved):

PR #418, opened 2024-02-26
fix: Handle file contents string type

JerrySentry:
### Purpose/Motivation

When retrieving a large (>1 MB) file from GH, the contents of the response is not encoded, so don't decode it when we handle it.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/1013

### Legal Boilerplate


core/commands/commit/interactors/get_file_content.py
```diff
@@ -13,6 +13,9 @@ async def get_file_from_service(self, commit, path):
                 owner=self.current_owner, repo=commit.repository
             )
             content = await repository_service.get_source(path, commit.commitid)
+            # Sometimes the content may be of type str, so don't decode it
```
ajay-sentry, line 16: super nit: this comment could be reworded to explain some of the cases where the content might be a str. Currently idk how much value the comment is bringing

ajay-sentry (approved):

PR #420, opened 2024-02-26
Fix: #251 - Update regex library to prevent excessive backtracking

ajay-sentry:
### Purpose/Motivation

- Aims to fix both places in the Codecov codebase that evaluate a user-supplied regex using Python's re.match. 
- Python's regex implementation allows backtracking
- ~~Replaces the areas using regex.match for user supplied regexes with [google's re2](https://github.com/google/re2)~~
- Replaces the areas using regex.match for user supplied regexes with a new [regex](https://pypi.org/project/regex/) library

### Links to relevant tickets

[#251](https://github.com/codecov/internal-issues/issues/251)

### What does this PR do?

- ~~re2 is a "drop-in" replacement for Python's re package, so I simply changed the two instances we have user supplied regexes with re2 from re.~~
- ~~In order to get the docker image to BUILD though, I had to do some "hacks" which were laid out in https://github.com/google/re2/issues/456. Explicitly, they were the following:~~
  - ~~abseil-cpp-dev~~
  - ~~py3-pybind11-dev~~
  - ~~re2-dev~~
- ~~Additionally, as called out in the above link, I had to use v1.0 instead of v1.1 of the library to get it to work.~~

- **UPDATE (2/29/24):** Once we got the re2 library working, we realized that there were two "big" issues with the re2 library
  - First, re2 doesn't support forward looking regex (ending wildcards) out the box, which for a few file types in empty_upload.py we had ending wildcards and I didn't want to mess with "removing" functionality that we may have had prior; though it _may_ have been a worthy tradeoff if the benefit is to prevent DDOSers
  - Second, the escape character for re2 is \\z vs. \\Z which required a little bit of funkiness throughout the codebase requiring "one-off" polyfills today, _or_ a mass conversion to re2. Which seemed a little anti-patterny and like scope creep we didn't need
  - After conversing with @JerrySentry he pointed out another regex library that has a built in timeout feature for pattern matching which seemed like a decent compromise. Plugging in a 5 second timeout for now, I think this solved the "root" issue of these potential long running regex backtracks, while at the same time not being _so_ involved that we needed to change a lot more things than necessary (I'm looking at you docker files!)

### Legal Boilerplate


JerrySentry (approved):
lgtm

adrian-codecov (approved):

PR #421, opened 2024-02-27
feat: allow config for multi-apps

giovanni-guidini:
These changes introduce `OwnerInstallationNameToUseForTask` model.

This model allows an `Owner` to configure a `installation_name` to be
used for a given `task_name`. This can only be configured manually for now.

That is on purpose as we want to test these changes first.

context: codecov/engineering-team#1146

> :warning: **These changes include a migration**
>
> Migration creates table for `OwnerInstallationNameToUseForTask`.
> Should not be disruptive when deployed.


codecov_auth/migrations/0053_ownerinstallationnametousefortask_and_more.py
```diff
@@ -0,0 +1,63 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name="OwnerInstallationNameToUseForTask",
+            fields=[
+                ("id", models.BigAutoField(primary_key=True, serialize=False)),
+                ("external_id", models.UUIDField(default=uuid.uuid4, editable=False)),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                ("installation_name", models.TextField()),
```
trent-codecov, line 37: This just uses the name out of the app installations table? Any reason not to use a fk id?

trent-codecov: image
This is what I was describing just to clarify. The way I understand it, a given task for an owner will only ever be assigned one GithubAppInstallation, either the default or a specific entry in your new table. 

> But there can be many OwnerInstallationName... that use the same installation_name and one installation_name can be used in multiple GithubAppInstallation.

I don't quite follow this statement. Happy to slack huddle too if you want to talk it out.

giovanni-guidini: Yes that makes perfect sense. That's not exactly the model I was working on though.

The difference is subtle, but the implication is essentially that it allows us to have [this piece of code](https://github.com/codecov/worker/pull/279/files#diff-ef1268deade32b7db2724bca634c4712d318fb3b2957cc2d7bf99b6124319541R33-R42) 

It might be an overkill. But then you have "classes of apps" essentially (under a same name).
And they are all considered pretty equally when you select them for some task.

matt-codecov: am i understanding your idea right?
- partner creates a GitHub app named `"Codecov Notifications"`
- partner creates an `OwnerInstallationNameToUseForTask` which says `tasks.Notify` should use the app called `"Codecov Notifications"`
- they wind up hitting the rate limit even on their single notifications app
- they create a distinct app, but they give it the same exact name `"Codecov Notifications"`
- now we will automatically choose from 2 `"Codecov Notifications"` apps without them having to configure another `OwnerInstallationNameToUseForTask`

if that's right, it does make the UX smoother for users when they are so big that a dedicated app for a task is still not enough and they need to add a second, third, fourth app

i don't have a strong opinion on whether the slightly-weird design here is worth it for the improvement. the alternative UX is not that bad (same steps + configuring another `OwnerInstallationToUseForTask`), and afaik the need for this should be pretty rare, but if we add comments explaining this then the design isn't that strange

giovanni-guidini: Your understanding is correct Matt

codecov_auth/migrations/0053_ownerinstallationnametousefortask_and_more.py
```diff
@@ -0,0 +1,63 @@
...
+    #     ALTER TABLE "codecov_auth_ownerinstallationnametousefortask" ADD CONSTRAINT "codecov_auth_ownerin_owner_id_8bf0ce9b_fk_owners_ow" FOREIGN KEY ("owner_id") REFERENCES "owners" ("ownerid") DEFERRABLE INITIALLY DEFERRED;
+    #     CREATE INDEX "codecov_auth_ownerinstalla_owner_id_8bf0ce9b" ON "codecov_auth_ownerinstallationnametousefortask" ("owner_id");
+    # COMMIT;
+
+    operations = [
+        migrations.CreateModel(
+            name="OwnerInstallationNameToUseForTask",
+            fields=[
+                ("id", models.BigAutoField(primary_key=True, serialize=False)),
+                ("external_id", models.UUIDField(default=uuid.uuid4, editable=False)),
```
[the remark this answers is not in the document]
giovanni-guidini: All codecov base models include `id`, `external_id`, `created_at` and `updated_at`.
It is basically an ID for the object that we can share with users. A way to reference it that is not the actual ID

trent-codecov (approved):

PR #425, opened 2024-02-28
feat: add methods to query + insert coverage measurements

adrian-codecov:
### Purpose/Motivation
We're adding a partitioned with the UserMeasurements models. We want to add some functions to add/query for monthly uploads. This PR adds those + tests

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1174

### What does this PR do?
- adds a query + insert functions using the user measurements model
- adds tests to those

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/utils/uploads_used.py
+++ b/utils/uploads_used.py
@@ -5,8 +5,12 @@ from django.conf import settings
 from django.db.models import Q
 from django.utils import timezone
 
+from codecov_auth.models import Owner
+from core.models import Commit, Repository
 from plan.constants import TrialStatus
-from reports.models import ReportSession
+from plan.service import PlanService
+from reports.models import ReportSession, ReportType
+from user_measurements.models import UserMeasurement
 from utils.config import get_config
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


utils/uploads_used.py
```diff
@@ -80,3 +84,43 @@ def query_uploads_used(plan_service, limit, owner):
         )
 
     return queryset[:limit].count()
+
+
+def query_monthly_coverage_measurements(plan_service: PlanService) -> int:
```
JerrySentry, line 89: How will this be used? How come it only does private repo?

utils/uploads_used.py
```diff
@@ -80,3 +84,43 @@ def query_uploads_used(plan_service, limit, owner):
...
+        report_type="coverage",
+    )
+    if (
+        plan_service.trial_status == TrialStatus.EXPIRED.value
+        and plan_service.has_trial_dates
+    ):
+        queryset = queryset.filter(
+            Q(created_at__gte=plan_service.trial_end_date)
+            | Q(created_at__lte=plan_service.trial_start_date)
+        )
```
JerrySentry, line 104: Do we need to filter out `ONGOING` rows?

utils/uploads_used.py
adrian-codecov: This is currently not in use but will replace the `query_uploads_used` fn. It's mimicking the logic there, excluding the `uploaded` filter (there's 2 types so far, uploaded and carriedforward; this new user_measurements table will only insert uploaded entries, so no need for it anymore) and the `...timedelta(days=60)` as that was a stab at performance, which this whole approach will take care of now

utils/uploads_used.py
adrian-codecov: When you're ongoing, you're in the `users-trial` plan, so there's no monthly_limit associated with it, so it automatically takes care of this

JerrySentry (approved):

PR #426, opened 2024-02-28
feat: check authorization for GQL owners

JerrySentry:
### Purpose/Motivation

Add a authentication and authorization for a few sensitive fields in the Owners query, we don't want to expose these information to users that don't belong in this org.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/1082

### Legal Boilerplate


graphql_api/types/owner/owner.graphql
```diff
@@ -15,21 +15,21 @@ type Owner {
   repository(name: String!): RepositoryResult!
```
JerrySentry, line 15: These fields can be null now, when the user is not part of the the field will be null and will also have error in the response with the appropriate message.

graphql_api/helpers/mutation.py
```diff
@@ -57,5 +58,20 @@ def authenticated_resolver(instance, info, *args, **kwargs):
     return authenticated_resolver
 
 
+def require_part_of_org(resolver):
+    def authenticated_resolver(queried_owner, info, *args, **kwargs):
+        current_user = info.context["request"].user
+        current_owner = info.context["request"].current_owner
+        if not current_user.is_authenticated or not current_owner:
```
ajay-sentry, line 65: will current_user always be non-null? Wonder if we should use .get() here instead

graphql_api/types/owner/owner.py
```diff
@@ -81,6 +82,7 @@ def resolve_plan_representation(owner: Owner, info) -> PlanData:
 
 @owner_bindable.field("availablePlans")
 @convert_kwargs_to_snake_case
+@require_part_of_org
```
ajay-sentry, line 85: what was the determining factor in picking which functions to add the decorator to? I see there are certain functions where we didn't add it so was just curious

graphql_api/helpers/mutation.py
JerrySentry: Updated to use .get

graphql_api/types/owner/owner.py
JerrySentry: https://www.notion.so/sentry/GraphQL-Owner-Type-Sensitive-Fields-f78cb53e77e4428aa6e04b2819d67884
and yes, good catch, added in all the fields

graphql_api/helpers/mutation.py
```diff
@@ -57,5 +58,20 @@ def authenticated_resolver(instance, info, *args, **kwargs):
     return authenticated_resolver
 
 
+def require_part_of_org(resolver):
```
ajay-sentry, line 61: is it easy to add a UT for this guy?

ajay-sentry (approved):
just a small comment about UT but otherwise looks good

graphql_api/helpers/mutation.py
JerrySentry: this function should be fully covered with the test_owner.py tests

barnett-yuxiang (commented):
cool

PR #428, opened 2024-02-29
fix: turn Commits endpoint into APIView instead of ListCreateAPIView

joseph-sentry:
### Purpose/Motivation
In the previous version of the endpoint's definition, there was a possibility of an IntegrityError. This change is meant to move the get_or_create into CommitViews endpoint.

### What does this PR do?
- Implement the CommitViews endpoint as an APIView instead of a ListCreateAPIView
- Create a new CommitSerializer
- add a get handler that returns the list of matching Commit objects
- add a post handler that possibly creates a new Commit object in the database

### DIFF
--- a/upload/tests/views/test_commits.py
+++ b/upload/tests/views/test_commits.py
@@ -43,22 +43,6 @@ def test_get_repo_not_found(db):
     assert exp.match("Repository not found")
 
 
-def test_get_queryset(db):
-    target_repo = RepositoryFactory(
-        name="the_repo", author__username="codecov", author__service="github"
-    )
-    random_repo = RepositoryFactory()
-    target_commit_1 = CommitFactory(repository=target_repo)
-    target_commit_2 = CommitFactory(repository=target_repo)
-    random_commit = CommitFactory(repository=random_repo)
-    upload_views = CommitViews()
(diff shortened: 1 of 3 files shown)


giovanni-guidini (commented):
The code itself seems fine and well-written.
But it seems a big change for a bug that you don't even show a test for... is there more about the bug?

Also looking at https://www.django-rest-framework.org/api-guide/serializers/#saving-instances maybe there's a REST framework way of using `get_or_create` without changing the endpoint so much?

```
Calling .save() will either create a new instance, or update an existing instance, depending on if an existing instance was passed when instantiating the serializer class:

# .save() will create a new instance.
serializer = CommentSerializer(data=data)

# .save() will update the existing `comment` instance.
serializer = CommentSerializer(comment, data=data)
```

Have you tried this approach?

Ideally though the SQL emitted would have some form on "on conflict do nothing" so the DB deals with the conflict... that seems a step further tho, trying to get an existing instance first would be ok

joseph-sentry:
going to close this for now because I made #437 to address your feedback

I think for this specific bug #437 is a better solution but I also think we should consider making the upload endpoint use the same format as the bundle analysis endpoint

PR #429, opened 2024-03-01
fix: store login state in a session cookie

matt-codecov:
fixes https://github.com/codecov/internal-issues/issues/270

### Legal Boilerplate


giovanni-guidini (approved):
LGTM. Great work 👍

codecov_auth/views/base.py
```diff
@@ -100,19 +103,48 @@ def generate_state(self) -> str:
...
         redirection_url = self._generate_redirection_url()
         self.redis.setex(self._get_key_redis(state), 500, redirection_url)
+
+        # By saving the state in a session cookie, we can ensure that the user
+        # following the redirection URL after OAuth authorization is the same
+        # as the user who initiated it. Otherwise, a trickster could generate
+        # a final redirect URL to log into their account, send it to some
+        # victim, and trick the victim into linking their account with the
+        # trickster's.
+        self.request.session[self._session_key()] = state
```
RulaKhaled, line 113: Can we add some sort of a verifier for extra layer of security e.g https://github.com/pennersr/django-allauth/blob/627e40c4babd5bb35d72d4c1bdf0d19def7a95ec/allauth/socialaccount/models.py#L363-L368?

matt-codecov: i think what we call state is what they call verifier, and what they call state is what we put in redis

RulaKhaled: verifier is a random string added to the state so it's harder to reduce state predictability

PR #430, opened 2024-03-01
feat: improve querying capabilities

daniel-codecov:
### Purpose/Motivation

Adds more natural querying to django admin for feature flags overrides. Should also speed up the query since no longer using `icontains`. Closes points 5 & 6 from here: https://github.com/codecov/engineering-team/issues/1315

https://github.com/codecov/codecov-api/assets/159859649/e448ee47-9047-4b10-ab8b-080fa0ad584e

### Legal Boilerplate

### DIFF
--- a/codecov/views.py
+++ b/codecov/views.py
@@ -3,7 +3,7 @@ from django.conf import settings
 from django.db import connection
 from django.http import HttpResponse, HttpResponseRedirect
 
-from codecov_auth.models import Owner
+from codecov_auth.models import Owner, Service
 from core.models import Constants, Repository
 
 _version = None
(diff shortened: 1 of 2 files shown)


codecov/views.py
```diff
@@ -40,8 +40,33 @@ def get_queryset(self):
 
         repos = Repository.objects.all()
 
-        if self.q:
-            repos = repos.filter(name__icontains=self.q)
+        terms = self.q.split("/") if self.q else []
```
daniel-codecov, line 43: is it fine to ignore these warnings from codecov?

codecov/views.py
```diff
@@ -54,7 +79,18 @@ def get_queryset(self):
 
         owners = Owner.objects.all()
 
-        if self.q:
-            owners = owners.filter(username__icontains=self.q)
```
daniel-codecov, line 58: swapping from `icontains` to `startswith` should help with speed

PR #433, opened 2024-03-04
fix: handle rate limit error in get pr info fn

joseph-sentry:
We want to handle receiving a 429 or 403
status code response from gh that is related
to hitting the gh rate limit.

now we do that when we try to get PR info while
trying to auth with tokenless

we will return a `exceptions.Throttled` which
corresponds to a 429 status code

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1247

### What does this PR do?
- Handle `TorngitRateLimitError` in `get_pull_request_info`
- add tests for `get_pull_request_info` to test it can handle receiving a 429 status code

### DIFF
--- a/codecov_auth/authentication/repo_auth.py
+++ b/codecov_auth/authentication/repo_auth.py
@@ -1,4 +1,5 @@
 import re
+from datetime import datetime
 from typing import List
 from uuid import UUID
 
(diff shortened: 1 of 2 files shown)


codecov_auth/authentication/repo_auth.py
```diff
@@ -211,6 +214,7 @@ class TokenlessAuthentication(authentication.TokenAuthentication):
     """
 
     auth_failed_message = "Not valid tokenless upload"
+    rate_limit_failed_message = "Tokenless has reached GitHub rate limit. Please consider uploading using a token: https://docs.codecov.com/docs/adding-the-codecov-token."
```
thomasrockhu-codecov, line 217:
```suggestion
    rate_limit_failed_message = "Tokenless has reached GitHub rate limit. Please upload using a token: https://docs.codecov.com/docs/adding-the-codecov-token."
```

adrian-codecov (approved):

PR #434, opened 2024-03-05
fix: use state for Sentry/Okta oAuth flow

RulaKhaled:
### Purpose/Motivation
fixes https://github.com/codecov/internal-issues/issues/270

### Legal Boilerplate

### DIFF
--- a/codecov_auth/views/okta.py
+++ b/codecov_auth/views/okta.py
@@ -14,7 +14,7 @@ from django.views import View
 from requests.auth import HTTPBasicAuth
 
 from codecov_auth.models import OktaUser, User
-from codecov_auth.views.base import LoginMixin
+from codecov_auth.views.base import LoginMixin, StateMixin
 from utils.services import get_short_service_name
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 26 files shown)


codecov_auth/views/okta.py
```diff
@@ -48,31 +48,38 @@ def validate_id_token(iss: str, id_token: str) -> dict:
...
                 "code": code,
                 "redirect_uri": settings.OKTA_OAUTH_REDIRECT_URL,
+                "state": state,
             },
         )
 
+        if not self.verify_state(state):
+            log.warning("Invalid state during Okta OAuth")
+            return None
+
```
matt-codecov, line 69: we already verify the state in `_perform_login()` before we call `_fetch_user_data()`. do we also need to verify it here? i suppose it wouldn't hurt

codecov_auth/views/okta.py
```diff
@@ -96,12 +103,18 @@ def _store_iss_cookie(self, iss: str, response: HttpResponse):
 
     def _perform_login(self, request: HttpRequest) -> HttpResponse:
```
matt-codecov, line 104: i think after we've successfully logged in we have to call `self.remove_state(state)` ([see the github view for reference](https://github.com/codecov/codecov-api/blob/f85b724284038271b3b68d8d968b633fd29eca5c/codecov_auth/views/github.py#L127))

matt-codecov (approved):
same comments apply for both sentry and okta. i think it looks good! accepting so you don't have to wait on a review after adding `self.remove_state(state)` to the views

codecov_auth/views/okta.py
RulaKhaled: Not if it is triggered from perform login, but i thought it would be good to verify on fetch user if the flow is ever initiated from there.

PR #435, opened 2024-03-05
Bug: Restore Github OIDC Auth

nora-codecov:
### Purpose/Motivation
Codecov CLI needs to be able to use Github OIDC for Authentication on upload endpoints.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1217

### What does this PR do?
Adds new authentication class `GitHubOIDCTokenAuthentication` to upload views

### Notes to Reviewer
Added one question in test_uploads, line 554, about which `token` to track


giovanni-guidini (approved):
LGTM

upload/views/uploads.py
```diff
@@ -53,6 +53,7 @@ class UploadViews(ListCreateAPIView, GetterMixin):
     authentication_classes = [
         GlobalTokenAuthentication,
         OrgLevelTokenAuthentication,
+        GitHubOIDCTokenAuthentication,
```
adrian-codecov, line 56: I think the order of these matter in terms of which one gets called first. If so, is this in the order you intend it to be in?

adrian-codecov (approved):

upload/views/uploads.py
nora-codecov: yes, I believe this is the correct order - the order was suggested by @trent-codecov on the [issue ticket](https://github.com/codecov/engineering-team/issues/1217)

PR #437, opened 2024-03-06
fix: use get_or_create in CommitSerializer.create

joseph-sentry:
### Purpose/Motivation
there is a possibility of creating commits concurrently if that happens there is a possibility of getting an IntegrityError when trying to create the Commit object

this commit solves this problem by using get_or_create to create the Commit object in the serializer

### What does this PR do?
- remove repository and commitid from validated data so its not included in the defaults
- use `get_or_create` instead of `filter` then `super().create`

### DIFF
--- a/upload/serializers.py
+++ b/upload/serializers.py
@@ -136,13 +136,12 @@ class CommitSerializer(serializers.ModelSerializer):
         )
 
     def create(self, validated_data):
-        commit = Commit.objects.filter(
-            repository=validated_data.get("repository"),
-            commitid=validated_data.get("commitid"),
-        ).first()
-        if commit:
-            return commit
-        return super().create(validated_data)
+        repo = validated_data.pop("repository", None)
+        commitid = validated_data.pop("commitid", None)
+        commit, _ = Commit.objects.get_or_create(
+            repository=repo, commitid=commitid, defaults=validated_data
+        )
+        return commit
 
 
 class CommitReportSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


upload/serializers.py
```diff
@@ -133,13 +133,12 @@ class Meta:
...
 
     def create(self, validated_data):
-        commit = Commit.objects.filter(
-            repository=validated_data.get("repository"),
-            commitid=validated_data.get("commitid"),
-        ).first()
-        if commit:
-            return commit
-        return super().create(validated_data)
+        repo = validated_data.pop("repository", None)
```
giovanni-guidini, line 136: Are you sure this is a `Repository` and not `RepositorySerializer`?

giovanni-guidini (commented):
I was looking at mentions of `CommitSerializer` in the unit tests, but could only find tests in which the `create` method are not called.

In `tests/views/test_commits.py::test_create_commit_already_exists` I don't see an assert to make sure that a duplicated Commit was not created.

Can you add a test that calls the `CommitSerializer.create` method twice and/or asserts that no duplicates happen?

upload/serializers.py
joseph-sentry: It's because we pass `repository` to the `serializer.save` call [here](https://github.com/codecov/codecov-api/blob/b99b186333870d5ac8add76c7abe28708f3b87c0/upload/views/commits.py#L66) which will then get put into the dict that is passed to this `create` [here](https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py#L746-L775)

giovanni-guidini: perfection, thank you for checking that

giovanni-guidini (approved):

PR #438, opened 2024-03-06
[admin] allow trial extension for orgs currently on trial

JerrySentry:
### Purpose/Motivation

Currently the admin app can extend a org trial, but only after they have expired and turned back to basic plan.
This change allows the orgs' trials to be extended while they're still trialing.

### Legal Boilerplate

### DIFF
--- a/codecov_auth/tests/test_admin.py
+++ b/codecov_auth/tests/test_admin.py
@@ -14,7 +14,11 @@ from codecov_auth.tests.factories import (
     SentryUserFactory,
     UserFactory,
 )
-from plan.constants import ENTERPRISE_CLOUD_USER_PLAN_REPRESENTATIONS, TrialStatus
+from plan.constants import (
+    ENTERPRISE_CLOUD_USER_PLAN_REPRESENTATIONS,
+    PlanName,
+    TrialStatus,
+)
 from utils.test_utils import APIClient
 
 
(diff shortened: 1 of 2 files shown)


codecov_auth/tests/test_admin.py
```diff
@@ -293,6 +297,26 @@ def test_start_trial_action(self, mock_start_trial_service):
...
 
+    @patch("plan.service.PlanService._start_trial_helper")
+    def test_extend_trial_action(self, mock_start_trial_service):
+        mock_start_trial_service.return_value = None
+        org_to_be_trialed = OwnerFactory()
+        org_to_be_trialed.plan = PlanName.TRIAL_PLAN_NAME.value
+        org_to_be_trialed.save()
+
+        res = self.client.post(
+            reverse("admin:codecov_auth_owner_changelist"),
```
ajay-sentry, line 308: looked this up, honestly a pretty interesting django feature lol

plan/service.py
```diff
@@ -139,21 +140,28 @@ def available_plans(self, owner: Owner) -> List[PlanData]:
         return available_plans
 
     def _start_trial_helper(
-        self, current_owner: Owner, end_date: datetime = None
+        self,
+        current_owner: Owner,
+        end_date: datetime = None,
```
ajay-sentry, line 145: This should be datetime | None for the type right?

ajay-sentry (approved):
makes sense, just 1 small nit

plan/service.py
JerrySentry: Yes, updated to `end_date: Optional[datetime] = None`

PR #440, opened 2024-03-07
GQL Owners: return null for unauthorized fields

JerrySentry:
### Purpose/Motivation

Currently a GQL error will appear for Unauthenticated/Unauthorized requests, this change makes it so that those fields would return null instead. Making this change because the UI calls for unauthorized fields would not spam Sentry errors.

### Legal Boilerplate


rohitvinnakota-codecov (approved):

barnett-yuxiang (commented):
cool

PR #442, opened 2024-03-11
test: fix query measurements after 30 days test

joseph-sentry:
### Purpose/Motivation
This test started failing ater 2024-03-10 because
the time at the query was not frozen to a specific date so it was using the system's actual time and
the measurements were not a part of the last 30 days.

### What does this PR do?
Freezes time around monthly coverage measurements query in test

### DIFF
--- a/utils/tests/unit/test_coverage_measurements.py
+++ b/utils/tests/unit/test_coverage_measurements.py
@@ -94,9 +94,12 @@ class CoverageMeasurement(TestCase):
         assert len(all_measurements) == 8
 
         plan_service = PlanService(current_org=owner)
+        freezer = freeze_time("2024-03-05T00:00:00")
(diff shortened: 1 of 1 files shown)


utils/tests/unit/test_coverage_measurements.py
```diff
@@ -94,9 +94,12 @@ def test_query_monthly_coverage_measurements_after_30_days(self):
         assert len(all_measurements) == 8
 
         plan_service = PlanService(current_org=owner)
+        freezer = freeze_time("2024-03-05T00:00:00")
```
ajay-sentry, line 97: nit: could you add a comment here for what this time signifies? similar to what we have on 81 and 87

ajay-sentry (approved):

PR #456, opened 2024-03-15
1427 use shared models for foundational apps

adrian-codecov:
### Purpose/Motivation
We're moving core, codecov_auth + reports models into shared. This is a PR to use the model related code there, + other things

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1427

### What does this PR do?
- Replaces model definitions w/ models in shared
- Deletes tests that were moved to shared
- Adjusts test imports in places where services were moved to shared

### Legal Boilerplate


codecov/settings_base.py
```diff
@@ -549,3 +549,10 @@
 
 SHELTER_PUBSUB_PROJECT_ID = get_config("setup", "shelter", "pubsub_project_id")
 SHELTER_PUBSUB_SYNC_REPO_TOPIC_ID = get_config("setup", "shelter", "sync_repo_topic_id")
+
+# Allows to do migrations from another module
+MIGRATION_MODULES = {
+    "codecov_auth": "shared.django_apps.codecov_auth.migrations",
+    "core": "shared.django_apps.core.migrations",
+    "reports": "shared.django_apps.reports.migrations",
+}
```
matt-codecov, line 558: not necessary for legacy_migrations?

core/migrations/__init__.py
```diff
@@ -0,0 +1 @@
+# from shared.django_apps.core.migrations import *
```
matt-codecov, line 1: delete pls

codecov/settings_base.py
adrian-codecov: Good catch, I left it out weeks back when doing this and forgot to put it in. 

In theory, we shouldn't add any new migration here, but it's the correct thing to do. I'll add it in

core/migrations/__init__.py
adrian-codecov: Thank you!

matt-codecov (approved):

trent-codecov (approved):

JerrySentry (approved):

PR #417, opened 2024-02-26
feat: Update to block users from accessing app if guest set to false for enterprise  

RulaKhaled:
### Purpose/Motivation
What is the feature? Why is this being done?
This change affects enterprise users only. We need a setting to disable guest access where it is enabled by default. We are looking to block guest users from accessing ALL enterprise pages and redirect them back to the login page, as we technically cannot block them there

### Links to relevant tickets
https://github.com/codecov/feedback/issues/252

closes: https://github.com/codecov/feedback/issues/297

### What does this PR do?
Include a brief description of the changes in this PR. Bullet points are your friend.

### Notes to Reviewer
Anything to note to the team? Any tips on how to review, or where to start?

### Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. In 2022 this entity acquired Codecov and as result Sentry is going to need some rights from me in order to utilize my contributions in this PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #458, opened 2024-03-18
feat: Guest feature attempt 2

RulaKhaled:
### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


JerrySentry:
What's the tldr on what happened with attempt 1?

RulaKhaled:
> What's the tldr on what happened with attempt 1?

**What is guest feature?** 
guest feature it meant to ensure that only authenticated users can access enterprise pages, any unauthenticated requests should be redirected to /login page. read more about it in this thread: https://sentry.slack.com/archives/C04MJP65Z2M/p1708605980807579?thread_ts=1706626113.674569&cid=C04MJP65Z2M 

**What was the first attempt?** 
Initially, I tried using a middleware: https://github.com/codecov/codecov-api/pull/417 it worked, but it worked too well. It wasn't sufficient because it blocked all endpoints except health and login. and apparently we need others like /upload and /commit to always be publicly accessed (some are used in CI build), and the second issue was the inability to fetch internal user to check if current user is logged in.

**what am i doing here**:
I'm tying this feature to the owner query to block access to all dashboard pages. As you know, all of our dashboard pages depend on this query, if you don't have access to owner, you don't have access to anything. This resolves both of issues mentioned in the first attempt.

JerrySentry (approved):
@RulaKhaled Thanks for the explanation. Approved. Is accessing v2 public REST API for these enterprise builds not an issue in this case?

RulaKhaled:
> @RulaKhaled Thanks for the explanation. Approved. Is accessing v2 public REST API for these enterprise builds not an issue in this case?

@JerrySentry I'm planning to follow up on the expected for the public API, but i'm not prioritizing it in this PR because most public endpoints require auth token anyway. 

@thomasrockhu-codecov hehe quoted Trent on that one

PR #467, opened 2024-03-22
fix: carefully assign ghapp names

giovanni-guidini:
The configuration for ghapps need to be done carefully before they can
be used. And that process is _manual_.

However it's very simple for a user to simply go and install an app to
their organization and point it's webhooks to us. In this case we would
think it's the default app and break the app.

So now we update the app's names everytime we get a new webhook for it.
Only the configured default app may use the default name, all others,
unless they manually receive a custom name, will be 'unconfigured_app'.

Closes: https://github.com/codecov/engineering-team/issues/1445

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/codecov_auth/models.py
+++ b/codecov_auth/models.py
@@ -510,10 +510,11 @@ class GithubAppInstallation(
 
     def is_configured(self) -> bool:
         """Returns whether this installation is properly configured and can be used"""
-        if self.name == GITHUB_APP_INSTALLATION_DEFAULT_NAME:
-            # The default app is configured in the installation YAML
+        if self.app_id is not None and self.pem_path is not None:
             return True
-        return self.app_id is not None and self.pem_path is not None
+        # The default app is configured in the installation YAML
+        installation_default_app_id = get_config("github", "integration", "id")
+        return self.app_id == installation_default_app_id
 
     def repository_queryset(self) -> BaseManager[Repository]:
         """Returns a QuerySet of repositories covered by this installation"""
--- a/webhook_handlers/views/github.py
+++ b/webhook_handlers/views/github.py
@@ -3,7 +3,7 @@ import logging
 import re
 from contextlib import suppress
 from hashlib import sha1, sha256
-from typing import Optional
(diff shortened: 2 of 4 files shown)


webhook_handlers/views/github.py
```diff
@@ -411,6 +411,29 @@ def pull_request(self, request, *args, **kwargs):
...
+            * none of the above (use 'unconfigured_app');
+
+        Returns the app name that should be used
+        """
+        if ghapp.name != GITHUB_APP_INSTALLATION_DEFAULT_NAME:
+            return ghapp.name
+        installation_default_app_id = get_config("github", "integration", "id")
+        # `app_id` and `installation_default_app_id` should both be ints
+        # But just to avoid differences parsing either the YAML or the request data
+        # Casting them to str for the comparison (str has less change of failing)
```
montross50, line 432: We probably want to check if pem_path is set here. If it is, we can also not change the name. Otherwise we are enforcing only one default app.

codecov_auth/models.py
```diff
@@ -510,10 +510,11 @@ class GithubAppInstallation(
 
     def is_configured(self) -> bool:
         """Returns whether this installation is properly configured and can be used"""
-        if self.name == GITHUB_APP_INSTALLATION_DEFAULT_NAME:
-            # The default app is configured in the installation YAML
+        if self.app_id is not None and self.pem_path is not None:
```
adrian-codecov, line 513: Is there a way to tell if it's configured properly? Idk if they could end up with a self.app_id that isn't related to their PEM

adrian-codecov (approved):

codecov_auth/models.py
giovanni-guidini: There is no way to tell that, I don't think.
If it's configured improperly we'd see the app being unable to be used, and errors everywhere.

But the configuration needs to be done manually by us. Low chance of someone that doesn't know what their doing making this config.

PR #469, opened 2024-03-22
Add components measurements retrieval

JerrySentry:
### Purpose/Motivation
Add a GQL resolver to retrieve measurements of components for a given repo.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1360

### Legal Boilerplate


graphql_api/actions/flags.py
```diff
@@ -48,25 +48,11 @@ def flag_measurements(
     after: datetime,
     before: datetime,
 ) -> Mapping[int, Iterable[dict]]:
```
JerrySentry, line 50: Move this to a shared generic function to be used by both flag and components measurements

graphql_api/actions/components.py
```diff
@@ -0,0 +1,25 @@
...
+    component_ids: Iterable[int],
+    interval: Interval,
+    after: datetime,
+    before: datetime,
+    branch: Optional[str] = None,
+) -> Mapping[int, Iterable[dict]]:
+    return measurements_by_ids(
+        repository=repository,
+        measurable_name=MeasurementName.COMPONENT_COVERAGE.value,
+        measurable_ids=component_ids,
```
ajay-sentry, line 20: Did we need to do a similar cast to str here like we did with flags? 
`measurable_ids=[str(flag_id) for flag_id in flag_ids],`

graphql_api/types/repository/repository.py
```diff
@@ -396,3 +398,53 @@ def resolve_repository_result_type(obj, *_):
...
+@sync_to_async
+def resolve_component_measurements(
+    repository: Repository,
+    info,
+    interval: Interval,
+    before: datetime,
+    after: datetime,
+    filters: Mapping = None,
+    ordering_direction: OrderingDirection = OrderingDirection.ASC,
+    branch: str = None,
```
ajay-sentry, line 414: nit: This should be Optional[str]

graphql_api/types/repository/repository.py
```diff
@@ -396,3 +398,53 @@ def resolve_repository_result_type(obj, *_):
...
+@repository_bindable.field("componentMeasurements")
+@convert_kwargs_to_snake_case
+@sync_to_async
+def resolve_component_measurements(
+    repository: Repository,
+    info,
+    interval: Interval,
+    before: datetime,
+    after: datetime,
+    filters: Mapping = None,
```
ajay-sentry, line 412: nit: This should be Optional[Mapping]

services/components.py
```diff
@@ -72,3 +75,41 @@ def head_totals(self) -> ReportTotals:
...
+        self.after = after
+        self.before = before
+
+    @cached_property
+    def name(self):
+        return self.component_id
+
+    @cached_property
+    def percent_covered(self):
+        if len(self.raw_measurements) > 1:
```
ajay-sentry, line 101: do we need to worry about return values for percent_covered or percent_changed if raw_measurements is 1 or 0?

JerrySentry: Yes, good catch, it should be ` if raw_measurements > 0`

graphql_api/types/repository/repository.py
JerrySentry: Done

graphql_api/types/repository/repository.py
JerrySentry: Done

graphql_api/actions/components.py
JerrySentry: No `measurable_ids` takes Iterable[str] which is what the type of component_ids is. Updated the `component_ids` param type to correct value.

ajay-sentry (approved):
lgtm!

Issue #441, opened 2024-03-07
Collect metrics for VS Code Extension usage via validate API endpoint


JerrySentry:
1. [API] Update the validate YAML endpoint to take in a new argument (`source`?) that indicates which app the call came from. Then add a Sentry metrics for this with label VSCode.
2. [VS Code Extension] Make the request to validate endpoint with `source=vs_code_extension`
3. Make Sentry Metrics dashboard to look at how many calls are coming from VS Code Extension

PR #472, opened 2024-03-28
feat: add source query param to validate v2 endpoint

spalmurray-codecov:
### Purpose/Motivation

We currently have no visibility into how much our vscode extension is being used outside of install counts. This PR, paired with the corresponding change to our vscode extension, will allow us to see how frequently the validate YAML endpoint is being hit from the extension.

Part of https://github.com/codecov/codecov-api/issues/441

### What does this PR do?
- Add a 'source' query param to the validate v2 endpoint


validate/views.py
```diff
@@ -77,6 +78,10 @@ class V2ValidateYamlHandler(V1ValidateYamlHandler):
     permission_classes = [AllowAny]
 
     def post(self, request, *args, **kwargs):
+        source = self.request.query_params.get("source", None)
+        if source is not None:
+            sentry_metrics.set("validate_yaml_source", source)
```
spalmurray-codecov, line 83: lmk if I'm doing this wrong - first time using sentry metrics

JerrySentry: I believe this would be using `incr`, double check here https://develop.sentry.dev/sdk/metrics/
something like:
```
source = self.request.query_params.get("source", "n/a")
sentry_metrics.incr("validate_v2", tags={"source": source})
```

JerrySentry (approved):

PR #473, opened 2024-03-28
[ezpz]: fix: uppercase enterprise env var name

matt-codecov:
### Legal Boilerplate

### DIFF
--- a/dev.sh
+++ b/dev.sh
@@ -15,7 +15,7 @@ _start_gunicorn() {
   if [[ "$STATSD_HOST" ]]; then
     suffix="--statsd-host ${STATSD_HOST}:${STATSD_PORT}"
   fi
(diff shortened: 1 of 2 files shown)


giovanni-guidini (approved):

dev.sh
```diff
@@ -15,7 +15,7 @@ _start_gunicorn() {
   if [[ "$STATSD_HOST" ]]; then
     suffix="--statsd-host ${STATSD_HOST}:${STATSD_PORT}"
   fi
-  if [[ "$RUN_ENV" == "enterprise" ]] || [[ "$RUN_ENV" == "DEV" ]]; then
+  if [[ "$RUN_ENV" == "ENTERPRISE" ]] || [[ "$RUN_ENV" == "DEV" ]]; then
```
trent-codecov, line 18: I would make this either case insensitive compare or add another if. Just in case users are already accommodating for this being lower case.

PR #495, opened 2024-04-08
add getting components from yaml GQL

JerrySentry:
Add a new GQL field in Repository to get all components in the YAML, returning the `id` and `name`, allow searching by a term
### Legal Boilerplate

### DIFF
--- a/graphql_api/types/repository/repository.graphql
+++ b/graphql_api/types/repository/repository.graphql
@@ -79,6 +79,9 @@ type Repository {
     filters: ComponentMeasurementsSetFilters
     orderingDirection: OrderingDirection
   ): [ComponentMeasurements!]!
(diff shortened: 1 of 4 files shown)


graphql_api/types/repository/repository.graphql
```diff
@@ -79,6 +79,9 @@ type Repository {
     filters: ComponentMeasurementsSetFilters
     orderingDirection: OrderingDirection
   ): [ComponentMeasurements!]!
+  componentsYaml(
+    term_id: String
```
RulaKhaled, line 83: let's use `termID` instead of snake case as this deals directly with client

JerrySentry: updated

PR #503, opened 2024-04-10
Update coverage/tree route to parse flags and components filters 

rohitvinnakota-codecov:
API changes to propagate totals data to the sunburst chart on the front end. 

The chart calls the `/coverage/tree endpoint`, which does not currently filter on flags or components. This change makes it possible for it to do so. We do something similar in [commit.py](https://github.com/codecov/codecov-api/blob/main/graphql_api/types/commit/commit.py).

There will be a follow up PR on gazebo that passes flags and components as query params. 

Example call on staging 
Screenshot 2024-04-15 at 2 46 20 PM

### Legal Boilerplate


barnett-yuxiang (commented):
LGTM

api/internal/tests/views/test_coverage_viewset.py
```diff
@@ -286,3 +296,39 @@ def test_tree_missing_report(self, build_report_from_commit):
...
+                {
+                    "component_id": "c1",
+                    "name": "ComponentOne",
+                    "paths": ["dne.py"],
+                }
+            ),
+        ]
+        build_report_from_commit.return_value = sample_report()
+        res = self._tree(components="Does_not_exist")
+        assert res.status_code == 404
```
ajay-sentry, line 334: Should we also add 1 more test for the flag filtering?

ajay-sentry (approved):
test comment, logic looks good

barnett-yuxiang (approved):
good

PR #507, opened 2024-04-15
fix: populate sessions when logging in

JerrySentry:
### Purpose/Motivation

Create a Session object that stores user info like IP and user agent when logging in. All sessions will be displayed on the UI in the settings/access page, can also click on revoke to delete the sessions which will logout the corresponding sessions. The Session is created whenever we create a DjangoSession object, for deletion the DjangoSession object is deleted which will cascade delete the Session object

### Links to relevant tickets

fixes: https://github.com/codecov/engineering-team/issues/1278

### Legal Boilerplate


barnett-yuxiang (commented):
LGTM

ajay-sentry (dismissed):
Discussed with Jerry on a huddle, lgtm!

ajay-sentry (approved):

PR #509, opened 2024-04-15
sec: Integrate Ariadne's query cost validator

ajay-sentry:
## Purpose/Motivation

Our goal with this PR is to mitigate the possibility of a DoS attack with users making calls to our GQL API. We aim to tackle this by using Ariadne's built in query cost validator, assigning a default cost of 1 to each field being requested, a multiplier of x amount (between 3 and 25) * number of edge resources requested (e.g. first:20 repos would add 20 * 25 to the cost, not 20 * number of fields requested for repo, but that is something we can update later on if we want additional granularity in the future)

This first pass PR is meant to lay the infrastructure and initial guidelines that I hope we can tinker and update as we better understand what cost "makes sense" for us long term. **For now,** I have set the maximum cost to the value mentioned on slack. This seems to be reasonable since we're fetching _at most_ "first: 20" in gazebo, and the "lengthiest" query looks to have a size of ~40 fields. Refering to: `src/services/user/useUser.js`

**Dev Notes:**
Below is the thought process that got me to this point and why I chose to add the query complexity logic server side, and using ariadne.

Initially started search looking at what we are currently using for sending HTTP requests on the frontend. GraphQL api looks to be a wrapper function around the JS native Fetch API. Didn’t want to rip out existing infrastructure and put in something like Apollo Client, though that does look to have native capabilities around validating query complexity or looking at query depth, etc. Fetch API also doesn’t have much “built in,” but didn’t rule out the possibility of making creating some simple middleware for doing some type of complexity analysis natively. But there’s probably still a much lower friction way to do this, so started looking at ways we can maybe accomplish something similar server side at the resolver level. 

It looks like Ariadne has some query cost validator built in: https://ariadnegraphql.org/docs/0.19/query-validators

**Helpful resources / Further Reading:**
- https://medium.com/workflowgen/graphql-query-timeout-and-complexity-management-fab4d7315d8d
- Fetch API Docs: https://javascript.info/fetch-api
- Other companies with these rules in place: https://aps.autodesk.com/en/docs/mfgdataapibeta/v2/developers_guide/query-complexity/
- NPM package if we used Apollo: https://github.com/slicknode/graphql-query-complexity

## Links to relevant tickets

Closes https://github.com/codecov/internal-issues/issues/385

## What does this PR do?

- Integrates Ariadne's cost complexity validator in our API
- Assigns cost of "1" for related fetches
- Assigns cost of "1" for each field requested
- Assigns maximum query cost to be a number defined via env variable (prevent bad actors from knowing our limit and abusing right below that value)

## Screenshots from Testing

**Query cost massively ballooning when I set query cost to be 20 on the multiplier.**

Screenshot 2024-04-15 at 2 31 15 PM

**Query cost prior to setting 20 on multiplier**
Screenshot 2024-04-15 at 2 04 54 PM

**Fuzzy find for "first" in Gazebo**
Screenshot 2024-04-15 at 2 38 01 PM

### UPDATE FROM 4/18/2024
**Testing out nesting pulls on the ReposForOwner query**
Screenshot 2024-04-18 at 1 49 18 PM

**Without nesting pulls on the ReposForOwner query (Original)**
Screenshot 2024-04-18 at 1 50 06 PM

Query for Reference:
`query ReposForOwner(
  $filters: RepositorySetFilters!
  $owner: String!
  $ordering: RepositoryOrdering!
  $direction: OrderingDirection!
  $after: String
  $first: Int
) {
  owner(username: $owner) {
    repositories(
      filters: $filters
      ordering: $ordering
      orderingDirection: $direction
      first: $first
      after: $after
    ) {
      edges {
        node {
          name
          active
          activated
          private
          coverage
          updatedAt
          latestCommitAt
          lines
          author {
            username
          }
          coverageEnabled
          bundleAnalysisEnabled
          repositoryConfig {
            indicationRange {
              upperRange
              lowerRange
            }
          }
          pulls(
            orderingDirection: $direction
            first: 20
            after: $after
          ){
            edges {
              node {
                title
              }
            }
          }
        }
      }
      pageInfo {
        hasNextPage
        endCursor
      }
    }
  }
}`

### Legal Boilerplate


graphql_api/views.py
```diff
@@ -87,12 +107,34 @@ async def post(self, request, *args, **kwargs):
...
+                    if data["errors"][0]["extensions"]["cost"]:
+                        costs = data["errors"][0]["extensions"]["cost"]
+                        log.error(
+                            "Query Cost Exceeded",
+                            extra=dict(
+                                requested_cost=costs.get("requestedQueryCost"),
+                                maximum_cost=costs.get("maximumAvailable"),
+                            ),
+                        )
+                        return HttpResponseBadRequest(
```
ajay-sentry, line 132: we don't want to expose the max cost to the client

graphql_api/views.py
```diff
@@ -64,6 +68,22 @@ class AsyncGraphqlView(GraphQLAsyncView):
...
 
+    def get_validation_rules(
+        self,
+        context_value: Optional[Any],
+        document: DocumentNode,
+        data: dict,
+    ) -> Optional[Collection]:
+        return [
+            cost_validator(
+                maximum_cost=settings.GRAPHQL_QUERY_COST_THRESHOLD,
```
ajay-sentry, line 79: ingest our env variable when deriving the max cost

JerrySentry (approved):

JerrySentry (approved):

joseph-sentry (approved):

PR #513, opened 2024-04-18
fix: use proper measurement last uploaded

JerrySentry:
### Purpose/Motivation

Don't use the aggregated timestamp from the summaries table, instead use the timestamp from the measurements table, this returns the true timestamp of when the last upload was inserted into the timeseries DB.

### Legal Boilerplate

### DIFF
--- a/graphql_api/actions/components.py
+++ b/graphql_api/actions/components.py
@@ -1,6 +1,8 @@
 from datetime import datetime
 from typing import Iterable, Mapping, Optional
 
+from django.db.models import QuerySet
+
 from codecov_auth.models import Owner
 from core.models import Repository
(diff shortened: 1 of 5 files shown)


graphql_api/actions/components.py
```diff
@@ -23,3 +27,18 @@ def component_measurements(
         before=before,
         branch=branch,
     )
+
+
+def component_measurements_last_uploaded(
```
ajay-sentry, line 32: nit: thoughts on adding return types to these fns? would make it a little easier to review for me (as someone not super versed in python syntax lol)

ajay-sentry (approved):
lgtm! Just the nit

ajay-sentry (approved):

PR #516, opened 2024-04-19
 Skip OwnerNotActivated error on unconfigured repos

rohitvinnakota-codecov:
This is a follow up to https://github.com/codecov/engineering-team/issues/1253. 

On the front end, we want to show an error inside a page instead of letting `NetworkErrorBoundary` overwrite the whole page. Furthermore, we need repository information to conditionally render Coverage and Bundle Analysis setup tabs regardless of the owner's activation status. 

With this change, we do not throw an `OwnerNotActivated` error if the repo is not configured, enabling us to display onboarding tabs on the front end without throwing a page level error. 

More information about the project [here](https://www.notion.so/sentry/Seats-Allocations-Epic-733b98e82b2345c59e830e65d70ea8b1). 

Please note that this must be merged before https://github.com/codecov/gazebo/pull/2794.

### Legal Boilerplate

### DIFF
--- a/graphql_api/tests/test_repository.py
+++ b/graphql_api/tests/test_repository.py
@@ -460,7 +460,7 @@ class TestFetchRepository(GraphQLTestHelper, TransactionTestCase):
 
     @patch("services.activation.is_activated")
     @patch("services.activation.try_auto_activate")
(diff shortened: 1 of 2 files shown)


RulaKhaled (approved):
Let's gooo

PR #521, opened 2024-04-23
disable user-agent prometheus label to decrease load

matt-codecov:
re-enable in https://github.com/codecov/engineering-team/issues/1654

apparently this label was bringing down prometheus so we are nuking it for now. there are so many distinct UAs that prometheus suffers from a combinatorial explosion of possible label combinations

the skeleton daniel set up to add/test this custom label is still good, just gotta tweak the actual value of the label to scrub some of the minutiae from it

### Legal Boilerplate


rfox-codecov (approved):
looks largely appropriate to me

jason-ford-codecov (approved):

PR #523, opened 2024-04-25
Add "token" variables to Sentry's EventScrubber denylist

michelletran-codecov:
### Purpose/Motivation
We want to ensure that `token`s don't get sent to Sentry when errors occur.

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/419

### What does this PR do?
This change will filter out the `_header` and `token_to_use` variables in a stacktrace frame so that the event doesn't push PII to Sentry. I'm currently filtering the whole `_header`, rather than just the `authorization` inside the `_header`. To enable that, I would need to enable `recursive` in the `EventScrubber`, which [might have potential performance implications](https://docs.sentry.io/platforms/python/data-management/sensitive-data/#event_scrubber). 

### Notes to Reviewer
We might want to actually store the `denylist` in the `shared` repo (so that we can reuse this same list in `worker`).

### Legal Boilerplate

### DIFF
--- a/codecov/settings_base.py
+++ b/codecov/settings_base.py
@@ -1,15 +1,13 @@
 import os
 from urllib.parse import urlparse
 
-import asgiref.sync as sync
 import sentry_sdk
-from asgiref.sync import SyncToAsync
 from corsheaders.defaults import default_headers
-from django.db import close_old_connections
 from sentry_sdk.integrations.celery import CeleryIntegration
 from sentry_sdk.integrations.django import DjangoIntegration
 from sentry_sdk.integrations.httpx import HttpxIntegration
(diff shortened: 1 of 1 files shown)


codecov/settings_base.py
```diff
@@ -533,10 +531,13 @@
 
 SENTRY_ENV = os.environ.get("CODECOV_ENV", False)
 SENTRY_DSN = os.environ.get("SERVICES__SENTRY__SERVER_DSN", None)
+sentry_deny_list = DEFAULT_DENYLIST + ["_headers", "token_to_use"]
```
giovanni-guidini, line 534: [very nit] SENTRY_DENY_LIST for consistency

giovanni-guidini (approved):

PR #526, opened 2024-04-26
Update sqlparse to 0.5.0

michelletran-codecov:
### Purpose/Motivation
This is to fix GHSA-2m57-hf25-phgg(https://github.com/advisories/GHSA-2m57-hf25-phgg) vulnerability.

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/427

### What does this PR do?
Update `sqlparse` to a non-vulnerable version.

### Legal Boilerplate

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -423,7 +423,7 @@ sortedcontainers==2.4.0
     #   intervaltree
 sqlalchemy==1.4.50
     # via shared
-sqlparse==0.4.4
+sqlparse==0.5.0
     # via django
 starlette==0.36.2
     # via ariadne


requirements.txt
```diff
@@ -423,7 +423,7 @@ sortedcontainers==2.4.0
     #   intervaltree
 sqlalchemy==1.4.50
     # via shared
-sqlparse==0.4.4
+sqlparse==0.5.0
```
adrian-codecov, line 426: Hmm, so the way we normally update our requirements is on the `requirements.in` file, which we then `pip-compile` to make it work. That also updates the `requirement.txt`, so if we were to update it this way, next time someone updates dependencies with `pip-compile` this will get overwritten since it depends on the django version (unless the thing you're upgrading is django).

I think the solution here is to add sqlparse as an independent dependency in the `requirements.in`, have it be at least 0.5.0, and then `pip-compile`. We'd just need to ensure our current django version isn't strictly tethered to 0.4.4 or 0.4.x

adrian-codecov (approved):
Spoke separately and we confirmed Django isn't strictly expecting 0.4.4

PR #527, opened 2024-04-29
fix: Remove duplicate names for resolvers and tests

ajay-sentry:
### Purpose/Motivation

When investigating another ticket, I saw that in our repo we sometimes used the same function name for multiple resolvers. This seemed like a clear code error, but then I realized that the resolver "bindable_field" decorator was added to things to maintain that uniqueness.

**However,** one place this wasn't accounted for was in tests. And I found some surprising results post fixing the duplicate test names. **That the tests with duplicate names were not running.** This seemed pretty concerning, since after fixing all the duplicate test names I saw there were 8 failing tests; ranging from a few months to a few years old.

This also made me realize that we don't have a linter on API (or worker / shared if I was looking at the code correctly), and might be something we want to look into adding sooner rather than later.

Anyway, this PR aims to fix all the duplicate function name errors in API, though I may need a bit of help with the remaining failing tests to see if they're still needed or not. My thinking is 3 of the 5 aren't needed anymore? In test_branch.py, and test_pull_comparison.py

## Screenshots

**AFTER Removing the duplicate function names**
Screenshot 2024-04-26 at 4 35 33 PM

**BEFORE**
Screenshot 2024-04-26 at 4 42 41 PM

### Legal Boilerplate

### DIFF
--- a/api/internal/tests/views/test_coverage_viewset.py
+++ b/api/internal/tests/views/test_coverage_viewset.py
@@ -334,24 +334,6 @@ class CoverageViewSetTests(APITestCase):
         assert res.status_code == 404
 
     @patch("shared.reports.api_report_service.build_report_from_commit")
-    @patch("services.components.commit_components")
-    def test_tree_not_found_for_components(
-        self, commit_components_mock, build_report_from_commit
-    ):
-        commit_components_mock.return_value = [
-            Component.from_dict(
-                {
-                    "component_id": "c1",
-                    "name": "ComponentOne",
-                    "paths": ["dne.py"],
-                }
-            ),
-        ]
-        build_report_from_commit.return_value = sample_report()
-        res = self._tree(components="Does_not_exist")
-        assert res.status_code == 404
-
-    @patch("shared.reports.api_report_service.build_report_from_commit")
     def test_tree_no_data_for_flags(self, build_report_from_commit):
         build_report_from_commit.return_value = sample_report()
         res = self._tree(flags="Does_not_exist")
--- a/codecov_auth/commands/owner/tests/test_owner.py
+++ b/codecov_auth/commands/owner/tests/test_owner.py
@@ -56,7 +56,7 @@ class OwnerCommandsTest(TransactionTestCase):
         interactor_mock.assert_called_once_with(input_dict)
 
     @patch("codecov_auth.commands.owner.owner.StartTrialInteractor.execute")
-    def test_cancel_trial_delegate_to_interactor(self, interactor_mock):
+    def test_start_trial_delegate_to_interactor(self, interactor_mock):
         org_username = "random_org"
         self.command.start_trial(org_username=org_username)
         interactor_mock.assert_called_once_with(org_username=org_username)
--- a/graphql_api/tests/test_branch.py
+++ b/graphql_api/tests/test_branch.py
@@ -628,7 +628,7 @@ class TestBranch(GraphQLTestHelper, TransactionTestCase):
         }
 
     @patch("services.report.build_report_from_commit")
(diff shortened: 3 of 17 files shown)


api/internal/tests/views/test_coverage_viewset.py
```diff
@@ -333,24 +333,6 @@ def test_tree_not_found_for_components(
         res = self._tree(components="Does_not_exist")
         assert res.status_code == 404
 
-    @patch("shared.reports.api_report_service.build_report_from_commit")
-    @patch("services.components.commit_components")
-    def test_tree_not_found_for_components(
```
ajay-sentry, line 338: This test looked to be an exact copy of the one above it

codecov_auth/commands/owner/tests/test_owner.py
```diff
@@ -56,7 +56,7 @@ def test_save_terms_agreement_delegate_to_interactor(self, interactor_mock):
         interactor_mock.assert_called_once_with(input_dict)
 
     @patch("codecov_auth.commands.owner.owner.StartTrialInteractor.execute")
-    def test_cancel_trial_delegate_to_interactor(self, interactor_mock):
```
ajay-sentry, line 59: For all the resolvers, I just named them what the property was

ajay-sentry: These changes weren't necessary though; just a cleanliness thing

graphql_api/tests/test_branch.py
```diff
@@ -628,36 +628,7 @@ def test_fetch_path_contents_unknown_path(
         }
 
     @patch("services.report.build_report_from_commit")
-    def test_fetch_path_contents_unknown_flags(self, report_mock):
```
ajay-sentry, line 631: this looked to be a duplicate of the below test that was never updated

graphql_api/tests/test_pull_comparison.py
```diff
@@ -979,68 +979,6 @@ def test_pull_comparison_no_comparison(self, compute_comparisons_mock):
 
         compute_comparisons_mock.assert_called_once
 
-    def test_pull_comparison_missing_head_report(self):
```
ajay-sentry, line 982: This and the following test are using deprecated patterns

upload/tests/test_upload.py
```diff
@@ -2629,7 +2629,10 @@ def test_circleci_invalid_commit(self, mock_get):
 
     @patch.object(requests, "get")
     def test_circleci_invalid_stop_time(self, mock_get):
-        expected_response = {"vcs_revision": "c739768fcac68144a3a6d82305b9c4106934d31a"}
+        expected_response = {
+            "vcs_revision": "c739768fcac68144a3a6d82305b9c4106934d31a",
+            "stop_time": "",
```
ajay-sentry, line 2634: test was missing stop_time

services/tests/test_components.py
```diff
@@ -153,9 +153,9 @@ def test_head_report(self, base_report_mock):
             }
         )
         component_comparison = ComponentComparison(self.comparison, component_go)
-        assert component_comparison.head_report.files == ["file_1.go"]
+        assert component_comparison.base_report.files == ["file_1.go"]
```
ajay-sentry, line 156: test wasn't actually looking at base_report

api/internal/tests/views/test_coverage_viewset.py
nora-codecov: good catch

nora-codecov (approved):
I agree with the linter issue you raised in the description of this pr - it's standard to have one, we should have one.

PR #530, opened 2024-04-30
Add CSP for not allow frame-ancestors

JerrySentry:
Testing confirmed that the platform lacks the Content-Type-Security header in HTTP responses. This does not directly lead to a security issue, yet it might aid attackers in their efforts to exploit other areas of weakness.

The X-Frame-Options header is present and this item specifies whether the web page is allowed to be framed. Although this header is known to prevent clickjacking attacks, a plethora of alternative breach strategies are achievable when a web page is frameable. The Content-Security-Policy (CSP) framework offers similar protection to X-Frame-Options via methods that overcome some shortcomings of the aforementioned header.

To optimally protect users of older browsers and modern browsers simultaneously, Cure53 recommends deploying the Content-Security-Policy: frame-ancestors 'none'; header in addition to the already implemented solutions.

### Legal Boilerplate


codecov/settings_base.py
```diff
@@ -304,16 +305,9 @@
     "REDOC_DIST": "SIDECAR",  # serve Redoc from Django (not CDN)
 }
```
JerrySentry, line 307: The reason I removed these existing policies is because it wasn't doing anything before this commit. These `CSP_*`  configurations are consumed by the `CSPMiddleware` module, so we probably don't need these as they didn't do anything this whole time.

ajay-sentry: This make sense, I'm wondering if maybe these were meant to be on though at some point but implementation was just never finished 🤔 

Or conversely were never cleaned up

ajay-sentry: Looks like they were added purposefully by Scott ~13 months ago, maybe we should actually keep?

JerrySentry: Yeah it was purposefully added, but the thing is those CSP headers aren't even in the response because the app doesn't use CSP headers right now. And with this change it would add those CSP headers, so I kind of don't want to introduce all those policies now, not knowing what they would impact.

ajay-sentry: Yeah that's a fair point, we can always add them back too if we want

ajay-sentry (approved):

PR #533, opened 2024-05-01
Tokenless V3

joseph-sentry:
### Purpose/Motivation
We don't want to make GH API calls when doing tokenless anymore

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1574

### What does this PR do?
- Remove GitHub API calls from TokenlessAuthentication

### DIFF
--- a/codecov_auth/authentication/repo_auth.py
+++ b/codecov_auth/authentication/repo_auth.py
@@ -255,70 +255,10 @@ class GitHubOIDCTokenAuthentication(authentication.TokenAuthentication):
         )
 
 
-tokenless_auth_failed_message = "Not valid tokenless upload"
-
-
-class TokenlessCommitAuthentication(authentication.TokenAuthentication):
-    def _get_info_from_request_path(self, request: HttpRequest) -> Repository:
-        path_info = request.get_full_path_info()
-        # The repo part comes from https://stackoverflow.com/a/22312124
-        upload_views_prefix_regex = r"\/upload\/(\w+)\/([\w\.@:_/\-~]+)\/commits"
-        match = re.search(upload_views_prefix_regex, path_info)
-
-        if match is None:
-            raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
-        # Validate provider
-        service = match.group(1)
-        try:
-            service_enum = Service(service)
-            # Currently only Github is supported
-            # TODO [codecov/engineering-team#914]: Extend tokenless support to other providers
-            if service_enum != Service.GITHUB:
-                raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
-        except ValueError:
-            raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
-
-        # Validate that next group exists and decode slug
-        encoded_slug = match.group(2)
-        repo = get_repository_from_string(service_enum, encoded_slug)
-        if repo is None:
-            # Purposefully using the generic message so that we don't tell that
-            # we don't have a certain repo
-            raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
-
-        return repo
-
-    def authenticate(self, request: HttpRequest):
-        # Get the repo
-        repository = self._get_info_from_request_path(request)
-
-        # Tokenless is only for public repos
-        if repository.private is True:
-            raise NotAuthenticated()
-
-        if request.method == "GET":
-            return (
-                RepositoryAsUser(repository),
-                TokenlessAuth(repository),
-            )
-
-        try:
-            body = json.loads(str(request.body, "utf8"))
-            branch = body.get("branch", "")
-        except json.JSONDecodeError:
-            return None
-        else:
-            if ":" in branch:
-                return (
-                    RepositoryAsUser(repository),
-                    TokenlessAuth(repository),
-                )
-            else:
-                return None
-
-
 class TokenlessAuthentication(authentication.TokenAuthentication):
-    def _get_info_from_request_path(self, request) -> Tuple[Repository, Commit]:
+    def _get_info_from_request_path(
+        self, request
+    ) -> tuple[Repository, str | None] | None:
         path_info = request.get_full_path_info()
         # The repo part comes from https://stackoverflow.com/a/22312124
         upload_views_prefix_regex = (
@@ -327,17 +267,14 @@ class TokenlessAuthentication(authentication.TokenAuthentication):
         match = re.search(upload_views_prefix_regex, path_info)
 
         if match is None:
-            raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
+            return None, None
+
         # Validate provider
         service = match.group(1)
         try:
             service_enum = Service(service)
-            # Currently only Github is supported
-            # TODO [codecov/engineering-team#914]: Extend tokenless support to other providers
-            if service_enum != Service.GITHUB:
-                raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
         except ValueError:
-            raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
+            return None, None
 
         # Validate that next group exists and decode slug
         encoded_slug = match.group(2)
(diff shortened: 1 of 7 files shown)


giovanni-guidini:
We need to be careful merging these changes to avoid re-introducing the bug of having an upload from a fork branch accidentally overwrite coverage in the upstream branch.

Considering the upload endpoints used by the CLI either we change the CLI to send the branch info every time (maybe through headers) OR we use the info we have to make the validation that the branch name is in the format `fork:branch`. If that is not the case, reject the upload.

Currently I believe it should be possible to make this check:
* For the commit creation the branch should be part of the body
* For the subsequent requests (create report and send the upload) we can pull the commit from the database and check its branch information.

Personally I feel that we should be able to accept / reject a request without having to look at the request body. So I'd opt to make the CLI send specific headers with the information we need.
(Also by removing the current checks maybe X-Tokenless-PR header is useless?)

joseph-sentry:
> So I'd opt to make the CLI send specific headers with the information we need.
(Also by removing the current checks maybe X-Tokenless-PR header is useless?)

We could just validate that the x-tokenless-pr header is in the format we expect and has the correct repo name?

giovanni-guidini:
X-Tokenless-PR is a number [1] that we used to get the correct PR from the provider. That does little for us considering the changes in this PR. If you just change the value the header name will be misleading.

X-Tokenless should _not_ match the repo name in the URL if it comes from a fork [1], cause it's supposed to be the fork slug, while the slug in the URL should be the upstream's.

[1] https://github.com/codecov/codecov-cli/blob/7b028499a521f8df3cf3bb6b642246a19e07bee9/codecov_cli/services/report/__init__.py#L57-L58

codecov_auth/authentication/repo_auth.py
```diff
@@ -243,87 +244,131 @@ def authenticate_credentials(self, token):
...
+def check_tokenless_header(request, branch=None):
+    tokenless = request.headers.get("X-Tokenless", "")
+
+    if not tokenless:
+        return False
+
+    if ":" not in tokenless:
+        raise exceptions.AuthenticationFailed(tokenless_auth_failed_message)
+
+    # make sure it's backwards compatible with the old way that
```
giovanni-guidini, line 259: It can't be compatible if validation fails (in the line above) because "the old way" doesn't include a `:` in the `X-Tokenless` header, ...right?

matt-codecov (commented):
this probably belongs in another PR, but in addition to getting rid of the `get_pull_request_info()` check like you do here, we also want to get rid of the "recent CI run" check. [that happens in `upload/helpers.py`](https://github.com/codecov/codecov-api/blob/e8cbe70c654074d2b4f1ed1adb59a94527d79420/upload/helpers.py#L271-L275). with that gone, i think we can delete [the whole `upload/tokenless` directory](https://github.com/codecov/codecov-api/tree/66f41fc2aeed3e588a12007d3d3b9203bc42c6a2/upload/tokenless)

also, the other day you made a great point about the report/upload endpoints being able to get the branch from the commit rather than needing it to be sent by the CLI. i think we can take advantage of that to simplify this logic across the API, CLI, and CI actions

on the CLI side, we already have to send the branch name in the request body for the create-commit step, and we already modify the branch name for forks there. the other two steps include which commit they are working with, so the API will be able to look up the commit and get the branch from there. so none of the commands need to set `X-Tokenless`, and the action doesn't have to set any extra env vars. (i think this also avoids an edge case where an attacker passes authentication by passing `X-Tokenless: fork:hahaha` to the do-upload endpoint but the commit SHA they passed is actually a commit on `main`)

then on the API side, i think the only difference between the commit endpoint and the others is where we get the branch name from? if so, i think we can merge back into one `TokenAuthentication` class, and if the commit sha is part of the URL we get the branch from the commit DB object, otherwise we look for the branch in the request body because we're creating a new commit

my brain is a little fried right now so this comment might not be very useful. i'll do another pass tomorrow, or maybe we could hop on a call and talk about it

upload/tests/views/test_reports.py
```diff
@@ -100,6 +114,7 @@ def test_reports_post_no_auth(db, mocker):
     repository = RepositoryFactory(
         name="the_repo", author__username="codecov", author__service="github"
     )
+    repository.private = False
     token = "BAD"
     commit = CommitFactory(repository=repository)
```
matt-codecov, line 119: the test never calls `commit.save()`. is it failing for the reason we expect, which is an invalid token? or is it failing because the commit isn't in the DB?

matt-codecov (changes requested):
i think you can do this with one `TokenAuthentication` class. the regex with the commit sha already makes that part an optional group so the same regex should work for all three endpoints. the logic could be something like:
```
def _get_branch(self, request, commit_from_url):
    if commit_from_url:
        return commit_from_url.branch
    else:
        body = json.loads(str(request.body, "utf8"))
        return body.get("branch")

def authenticate(self, request):
    repo, commit_from_url = self._get_info_from_request_path(request)
    branch = self._get_branch(request, commit_from_url)

    if not branch:
        raise Whatever()
    ...
```

also, i left a couple comments in test files themselves but as a general note, can you make sure there are test cases that ensure a tokenless upload for a unprefixed branch name fails? the test cases covering tokenless rejections are all due to the repo being private, but we also want to reject based on the branch name

codecov_auth/authentication/repo_auth.py
```diff
@@ -253,86 +256,64 @@ def authenticate_credentials(self, token):
...
-
-    def _get_repo_info_from_request_path(self, request) -> Repository:
+    def _get_info_from_request_path(
+        self, request
+    ) -> tuple[Repository, str | None] | None:
         path_info = request.get_full_path_info()
         # The repo part comes from https://stackoverflow.com/a/22312124
-        upload_views_prefix_regex = r"\/upload\/(\w+)\/([\w\.@:_/\-~]+)\/commits"
+        upload_views_prefix_regex = (
+            r"\/upload\/(\w+)\/([\w\.@:_/\-~]+)\/commits(?:\/([a-f0-9]{40}))?"
```
thomasrockhu-codecov, line 265: did this regex change?

codecov_auth/authentication/repo_auth.py
```diff
@@ -253,86 +256,64 @@ def authenticate_credentials(self, token):
...
-            sentry_metrics.incr("auth.get_pr_info.rate_limit_hit")
-            if e.reset:
-                now_timestamp = datetime.now().timestamp()
-                retry_after = int(e.reset) - int(now_timestamp)
-            elif e.retry_after:
-                retry_after = int(e.retry_after)
-            raise exceptions.Throttled(
-                wait=retry_after, detail=self.rate_limit_failed_message
-            )
+        commitid = match.group(3)
```
thomasrockhu-codecov, line 288: I wonder if it would be better to aggregate the `match.group(x)` calls into one paragraph so it's easier to see

codecov_auth/authentication/repo_auth.py
```diff
@@ -253,86 +256,64 @@ def authenticate_credentials(self, token):
...
+
+        return repo, commitid
+
+    def get_branch(self, request, commitid=None):
+        if commitid:
+            commit = Commit.objects.get(commitid=commitid)
```
thomasrockhu-codecov, line 294: I think `.get` will throw an error if this doesn't exist instead of `None`

codecov_auth/authentication/repo_auth.py
joseph-sentry: yes, there's an optional commit sha group at the end now

matt-codecov (commented):
is there a reason you got rid of all of the exceptions that we used to throw? i think returning `None` means something different from raising an exception to Django. if we want to stop throwing we should probably do it as a separate change

aside from that, the implementation looks good! now just drilling into the tests because this area keeps biting us (and our users) and i want to make sure this is rock-solid so we can be free of it for a while haha. the tests generally exercise all the important behaviors, but with some tweaks we can be more confident that the system works the way we think and that the tests will start failing if any part of it breaks
- logically, we have two inputs: `private` and `prefixed_branch`, and we want to test all combinations of those two inputs. all of them are tested at one point or another in your PR currently, but IMO we should have versions of all combos in all four test files:
  - `private=False`, `prefixed_branch=True`, auth succeeds
  - `private=False`, `prefixed_branch=False`, auth fails
  - `private=True`, `prefixed_branch=True`, auth fails
  - `private=True`, `prefixed_branch=False`, auth fails (we can actually skip this one; both of the failure reasons are already covered by other tests)
- when deciding whether a branch is prefixed or not, we can get the branch from two different places. what should happen if the branch is found in both places? our tests should address that one way or the other:
  - write tests for `get_branch()` directly to guarantee that the request body's branch is ignored if the commit sha is passed in, and/or
  - in test cases that put the commit ID in the URL, put an "opposite" branch in the request body. in other words, if the URL commit's branch is prefixed, put an unprefixed branch in the request body. if the URL commit's branch is unprefixed, put a prefixed branch in the request body. that way, the test will fail if we ever start reading the branch name from the wrong place.

matt-codecov (approved):
great work!

PR #542, opened 2024-05-06
fix: update tokenless string

thomasrockhu-codecov:
### Purpose/Motivation
Showing the wrong string message before. This will push the retry timing to the user as to not block CIs while still displaying the correct timing.

### Links to relevant tickets
fixes https://github.com/codecov/infrastructure-team/issues/346

### What does this PR do?
- Sets wait to `None`, this will push the behavior to the user. Bash and uploader (I believe) currently do an exp backoff up to 10 seconds. Is this enough time? Absolutely not, but it will not run for 6 hours.

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/upload/tokenless/github_actions.py
+++ b/upload/tokenless/github_actions.py
@@ -52,9 +52,12 @@ class TokenlessGithubActionsHandler(BaseTokenlessUploadHandler):
                 retry_after = int(e.retry_after)
             else:
                 retry_after = None
+            time_to_available_str = ''
+            if retry_after is not None:
+                time_to_available_str = f" Expected time to availability: {retry_after}s."
             raise exceptions.Throttled(
-                wait=min(10, retry_after) if type(retry_after) == int else retry_after,
-                detail="Rate limit reached. Please upload with the Codecov repository upload token to resolve issue.",
+                wait=None,
+                detail=f"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue.{time_to_available_str}",
             )
         except TorngitClientError as e:
             self.log_warning(message=f"Request client error {e}")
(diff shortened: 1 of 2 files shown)


upload/tokenless/github_actions.py
```diff
@@ -52,9 +52,14 @@ def get_build(self):
...
+            time_to_available_str = ""
+            if retry_after is not None:
+                time_to_available_str = (
+                    f" Expected time to availability: {retry_after}s."
+                )
             raise exceptions.Throttled(
-                wait=min(10, retry_after) if type(retry_after) == int else retry_after,
-                detail="Rate limit reached. Please upload with the Codecov repository upload token to resolve issue.",
+                wait=None,
+                detail=f"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue.{time_to_available_str}",
```
joseph-sentry, line 62: [todo] taking a look at this "Please upload with the Codecov repository upload token to resolve issue" message, I think it's inaccurate, but I'm not sure how else to phrase this. Perhaps it would make sense to add a page to the docs that explain _why_ using the repo upload token may help because it means we can decide to use a github installation if it exists for that user.

joseph-sentry (approved):
just left a todo comment for something that we should consider doing as part of another change

upload/tokenless/github_actions.py
thomasrockhu-codecov: @joseph-sentry oh that's a really good point, let me noodle on this for a bit...

rohan-at-sentry: https://github.com/codecov/feedback/issues/364

PR #545, opened 2024-05-07
Add GH action to do mypy check for patch

JerrySentry:
Add mypy type checking in the GH CI. This does type checking for all files that are touched in the PR. This job runs at the same time the other main jobs run, however right now a failure on this does not block mergning (but it is encouraged to not fail the job), eventually it may start blocking merge to main.

### Legal Boilerplate

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -11,4 +11,4 @@ on:
 jobs:
   patch-typing-check:
     name: Run Patch Type Check
-    uses: codecov/gha-workflows/.github/workflows/mypy.yml@may_06_mypy
+    uses: codecov/gha-workflows/.github/workflows/mypy.yml@v1.2.18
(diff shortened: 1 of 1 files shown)


.github/workflows/mypy.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches:
+      - main
+      - staging
+  pull_request:
+  merge_group:
+
+jobs:
+  patch-typing-check:
+    name: Run Patch Type Check
+    uses: codecov/gha-workflows/.github/workflows/mypy.yml@may_06_mypy
```
JerrySentry, line 14: TODO: change the pinned version once the gha-workflows PR is merged and a new release is created

joseph-sentry (approved):

matt-codecov (approved):

PR #546, opened 2024-05-07
coverage over time return component ID and name

JerrySentry:
Have `ComponentMeasurements` type return both component ID and name, if no name is supplied in the YML file the field will default to the ID.

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/component/component.graphql
+++ b/graphql_api/types/component/component.graphql
@@ -5,6 +5,7 @@ type Component {
 }
 
 type ComponentMeasurements {
+    componentId: String!
     name: String!
     percentCovered: Float
     percentChange: Float
--- a/graphql_api/types/repository/repository.py
+++ b/graphql_api/types/repository/repository.py
@@ -426,9 +426,7 @@ def resolve_component_measurements(
         return []
 
     if filters and "components" in filters:
-        components = [
-            c_id for c_id in components if c_id.component_id in filters["components"]
-        ]
+        components = [c for c in components if c.component_id in filters["components"]]
 
     component_ids = [c.component_id for c in components]
     all_measurements = component_measurements(
@@ -445,6 +443,10 @@ def resolve_component_measurements(
         row["measurable_id"]: row["last_uploaded"] for row in last_measurements
     }
 
+    components_mapping = {
+        component.component_id: component.name for component in components
+    }
+
     queried_measurements = [
         ComponentMeasurements(
             raw_measurements=all_measurements.get(component_id, []),
@@ -453,6 +455,7 @@ def resolve_component_measurements(
             after=after,
             before=before,
             last_measurement=last_measurements_mapping.get(component_id),
+            components_mapping=components_mapping,
         )
         for component_id in component_ids
     ]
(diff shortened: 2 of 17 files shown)


graphql_api/types/component/component.graphql
```diff
@@ -5,6 +5,7 @@ type Component {
 }
 
 type ComponentMeasurements {
+    componentId: String!
```
JerrySentry, line 8: Real change.

graphql_api/types/repository/repository.py
```diff
@@ -426,9 +453,7 @@ def resolve_component_measurements(
         return []
 
     if filters and "components" in filters:
-        components = [
-            c_id for c_id in components if c_id.component_id in filters["components"]
-        ]
+        components = [c for c in components if c.component_id in filters["components"]]
```
JerrySentry, line 457: Real change. just rename var name

graphql_api/types/repository/repository.py
```diff
@@ -445,6 +470,10 @@ def resolve_component_measurements(
         row["measurable_id"]: row["last_uploaded"] for row in last_measurements
     }
 
+    components_mapping = {
+        component.component_id: component.name for component in components
+    }
+
     queried_measurements = [
```
JerrySentry, line 477: Real change.

graphql_api/types/repository/repository.py
```diff
@@ -453,6 +482,7 @@ def resolve_component_measurements(
             after=after,
             before=before,
             last_measurement=last_measurements_mapping.get(component_id),
+            components_mapping=components_mapping,
```
JerrySentry, line 485: Real change.

mypy.ini
```diff
@@ -4,4 +4,5 @@
 disallow_untyped_defs = True
 ignore_missing_imports = True
 disable_error_code = attr-defined,import-untyped,name-defined
-follow_imports = silent
\ No newline at end of file
+follow_imports = silent
+warn_no_return = False
```
JerrySentry, line 8: Added this new rule to not check implicit `return None`, not a fan of that.

services/components.py
```diff
@@ -80,42 +80,50 @@ def patch_totals(self) -> ReportTotals:
 class ComponentMeasurements:
     def __init__(
         self,
```
JerrySentry, line 82: Mostly real changes on this file

graphql_api/types/repository/repository.py
ajay-sentry: Dictionary comprehension is pretty neat

ajay-sentry (approved):
lgtm

services/components.py
```diff
@@ -80,42 +80,50 @@ def patch_totals(self) -> ReportTotals:
...
         self.interval = interval
         self.after = after
         self.before = before
         self.last_measurement = last_measurement
+        self.components_mapping = components_mapping
 
     @cached_property
-    def name(self):
+    def name(self) -> str:
+        if self.components_mapping.get(self.component_id):
```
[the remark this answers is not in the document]
JerrySentry: Not really because the `shared.Components` object would always set the `name` to be empty string if it is not defined (never None). So we would have dicts like:
```
components_mapping = {
  "Id1": "",
  "Id2": "Name1",
}
```
using the `.get` would return the empty string as name now because it found the empty string name value for the ID, but ofc doing `if ""` returns False

PR #552, opened 2024-05-08
Run correct migration command

trent-codecov:
### Purpose/Motivation
Run correct migration command

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


matt-codecov (commented):
i think this is a narrower command than the original? it would prevent the ts_telemetry migrations from running (although i will kill those whenever i get around to it)

adrian-codecov (approved):
What Matt said is correct, ts_telemetry wouldn't get ran, is that okay?

Also, would enterprise eventually want partitions? Not for this PR but to keep in mind

PR #559, opened 2024-05-13
Add update self hosted settings mutation

rohitvinnakota-codecov:
### Purpose/Motivation

This PR closes https://github.com/codecov/engineering-team/issues/1473. We want to move our REST endpoints to GQL. This change creates a mutation and adds an additional resolver that will be consumed in gazebo in `useSelfHostedSettings` and `AutoActivateMembers` (we are currently making a `GET` and `PATCH` REST call). 

A follow up PR will be made in gazebo to migrate the endpoints. 

After that, we will delete the old REST code, once we confirm that it is no longer in use. 

### Legal Boilerplate


codecov_auth/commands/owner/interactors/tests/test_update_self_hosted_settings.py
barnett-yuxiang, line 1: Main improvements:

1. Fixture usage:  Use the fixture mechanism of pytest to create test data and improve the reusability and readability of the code.

2. Parameterized testing:  Use pytest's @pytest.mark.parametrize to parameterize test cases and reduce duplicate code.

3. Error assertion improvement:  When using the pytest.raises context manager, use the match parameter to more precisely match exception messages.

4. Test naming optimization:  Make the name of the test case more reflective of the specific behavior of the test.

5. Magic Mocking removed: The  patch was not used for actual testing and should be removed.

barnett-yuxiang (commented):
ok

codecov_auth/commands/owner/interactors/update_self_hosted_settings.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class UpdateSelfHostedSettingsInteractor(BaseInteractor):
+    def validate(self) -> None:
+        if not self.current_user.is_authenticated:
+            raise Unauthenticated()
+
+        if not settings.IS_ENTERPRISE:
+            raise ValidationError(
+                "enable_autoactivation and disable_autoactivation are only available in self-hosted environments"
```
ajay-sentry, line 24: nit: enable_auto_activation vs. enable_autoactivation

ajay-sentry: I could see it either way, but because we're camelcasing the input on AutoActivate, it seems more consistent

rohitvinnakota-codecov: these are what the methods are called in `self_hosted.py`

ajay-sentry (approved):
one small comment, otherwise lgtm

codecov_auth/commands/owner/interactors/update_self_hosted_settings.py
ajay-sentry: ooo okay makes sense

PR #560, opened 2024-05-13
delete session security fix

JerrySentry:
fixes: https://github.com/codecov/internal-issues/issues/469

Only allow sessions to be deleted if they are belonging to the authenticated user.

### Legal Boilerplate


codecov_auth/commands/owner/interactors/delete_session.py
```diff
@@ -7,14 +7,24 @@
...
 
 class DeleteSessionInteractor(BaseInteractor):
-    def validate(self):
+    def validate(self) -> None:
         if not self.current_user.is_authenticated:
             raise Unauthenticated()
 
     @sync_to_async
-    def execute(self, sessionid: int):
+    def execute(self, sessionid: int) -> None:
```
ajay-sentry, line 15: I thought there was a mypy rule that makes "return as none" implicit if no type is presented?

I could see it either way tho, related thread here: https://github.com/python/mypy/issues/7511

ajay-sentry: Yeah fair enough, Im good to keep as is

codecov_auth/tests/unit/views/test_okta.py
```diff
@@ -10,8 +10,9 @@
 
 from codecov_auth.models import OktaUser
 from codecov_auth.tests.factories import OktaUserFactory, OwnerFactory, UserFactory
-from codecov_auth.views.okta import OktaLoginView, validate_id_token
+from codecov_auth.views.okta import OktaLoginView
 from codecov_auth.views.okta import auth as okta_basic_auth
+from codecov_auth.views.okta import validate_id_token
```
ajay-sentry, line 15: how come we split this to a separate line?

ajay-sentry (approved):

codecov_auth/tests/unit/views/test_okta.py
JerrySentry: Uh ruff did that..

PR #568, opened 2024-05-15
Add encode secret string resolver 

rohitvinnakota-codecov:
Adds a resolver to return an encoded secret string. We want to move our REST endpoints to GQL(we are currently using a `POST` request to do this from gazebo). 

A follow up PR will be made in gazebo to migrate the endpoints.

After that, we will delete the old REST code, once we confirm that it is no longer in use.

This issue closes https://github.com/codecov/engineering-team/issues/1746

### Legal Boilerplate


core/commands/repository/interactors/encode_secret_string.py
```diff
@@ -0,0 +1,27 @@
...
+from core.models import Repository
+
+
+class EncodeSecretStringInteractor(BaseInteractor):
+    @sync_to_async
+    def execute(self, owner: Owner, repo: Repository, value: str) -> str:
+        if not self.current_user.is_authenticated:
+            raise Unauthenticated()
+        if not repo:
+            raise ValidationError("Repo not found")
```
JerrySentry, line 17: 1. What does the encoded secret string actually do?
2. Currently this allows authenticated users not belonging to the org to generate this value as well, is that ok?

rohitvinnakota-codecov: We have a widget in gazebo that lets us generate a secret string(for any text, but meant to be used for YAML values) 

Screenshot 2024-05-19 at 2 35 31 PM

It returns an encrypted string for the text entered. It's basically a big wrapper around the `encode_secret_string` function. 

Since the user can enter any string, and they will need org validation to see other sensitive data, I am comfortable with allowing any authenticated user to call this. 

Let me know if you think I'm missing anything, thanks.

JerrySentry: Gotcha, as long as this has the same auth validation as the internal REST endpoint its fine.

JerrySentry (approved):

PR #570, opened 2024-05-16
 Add erase repository content mutation 

rohitvinnakota-codecov:
Adds a mutation to erase a repository's content, replacing the REST endpoint we currently have in place. 

A future PR will update gazebo to call this GQL mutation, and a follow up to that will remove the old REST code once we confirm it is no longer in use. 

This PR closes https://github.com/orgs/codecov/projects/31/views/11?pane=issue&itemId=63021740. 

### Legal Boilerplate


core/commands/repository/interactors/erase_repository.py
```diff
@@ -0,0 +1,33 @@
...
+from codecov.db import sync_to_async
+from codecov_auth.helpers import current_user_part_of_org
+from codecov_auth.models import Owner
+from core.models import Repository
+from services.task.task import TaskService
+
+
+class EraseRepositoryInteractor(BaseInteractor):
+    def validate_owner(self, owner: Owner):
+        if not current_user_part_of_org(self.current_owner, owner):
```
ajay-sentry, line 16: I noticed for a lot of Interactors we also add code like this:

https://github.com/codecov/codecov-api/blob/a29cd709908af81dac427a30dff8c6436480373f/codecov_auth/commands/owner/interactors/update_default_organization.py#L15-L16

Is that not needed here?

graphql_api/types/mutation/erase_repository/erase_repository.graphql
```diff
@@ -0,0 +1,9 @@
+union EraseRepositoryError = UnauthenticatedError | UnauthorizedError | ValidationError
```
ajay-sentry, line 1: I don't think we have a test case for the UnauthenticatedError yet, right?

graphql_api/types/mutation/erase_repository/erase_repository.py
```diff
@@ -0,0 +1,21 @@
+from ariadne import UnionType, convert_kwargs_to_snake_case
```
ajay-sentry, line 1: nit: convert_kwargs_to_snake_case looks unused here

core/commands/repository/interactors/erase_repository.py
rohitvinnakota-codecov: I had it initially and realized that the `@require_authenticated` decorator takes care of that. 

The other code is a result of what I had originally thought so I will remove that.

ajay-sentry (approved):

PR #574, opened 2024-05-20
dev: Enable most F rules for Ruff

ajay-sentry:
### Purpose/Motivation

Part of the Ruff Enhancement epic, this adds all the "F" rules to our lint configuration for API, disabling a few super noisy ones that we can fix in individual PR's later. The benefit of doing it this way is we get ~40 rules "for free" that are on now checking our code :) At the cost of letting 4 of them be excluded til we have more bandwidth.

Rule list: https://docs.astral.sh/ruff/rules/#pyflakes-f
- The F Rules are all the "PyFlakes" rules that Ruff has configured

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/1716

### What does this PR do?

The rules we've mainly had to fix were around duplicate imports or duplicate keys in dicts. The rest we were already good on. Woohoo!

### Legal Boilerplate

### DIFF
--- a/codecov/settings_base.py
+++ b/codecov/settings_base.py
@@ -1,6 +1,7 @@
 import os
 from urllib.parse import urlparse
 
+import django_prometheus
 import sentry_sdk
 from corsheaders.defaults import default_headers
 from sentry_sdk.integrations.celery import CeleryIntegration
@@ -252,7 +253,7 @@ if TIMESERIES_ENABLED:
         }
 
 # See https://django-postgres-extra.readthedocs.io/en/master/settings.html
-POSTGRES_EXTRA_DB_BACKEND_BASE: "django_prometheus.db.backends.postgresql"
+POSTGRES_EXTRA_DB_BACKEND_BASE: "django_prometheus.db.backends.postgresql"  # type: ignore
 
 # Allows to use the pgpartition command
 PSQLEXTRA_PARTITIONING_MANAGER = (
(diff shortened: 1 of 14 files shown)


codecov/settings_base.py
```diff
@@ -252,7 +253,7 @@
         }
 
 # See https://django-postgres-extra.readthedocs.io/en/master/settings.html
-POSTGRES_EXTRA_DB_BACKEND_BASE: "django_prometheus.db.backends.postgresql"
+POSTGRES_EXTRA_DB_BACKEND_BASE: "django_prometheus.db.backends.postgresql"  # type: ignore
```
ajay-sentry, line 256: Even post import it didn't resolve this correctly; but docs here: https://django-dbbackup.readthedocs.io/en/stable/databases.html

ruff.toml
```diff
@@ -37,8 +37,8 @@ target-version = "py312"
 
 [lint]
 # Currently only enabled for F541 and I: https://docs.astral.sh/ruff/rules/
-select = ["F541", "I"]
-ignore = []
+select = ["F", "I"]
+ignore = ["F841", "F401", "F405", "F403"]
```
ajay-sentry, line 41: My preferred way of tackling this initiative; then we can have separate tickets for F841, etc. or multiple PR's

JerrySentry: These 4 doesn't have autofixes?

JerrySentry (approved):

adrian-codecov (approved):

PR #576, opened 2024-05-21
Lower log severity level to remove Sentry noise

JerrySentry:
### Purpose/Motivation

fixes: https://github.com/codecov/internal-issues/issues/463

Change the severity level from `log.exception` to `log.info`, showing on exception level doesn't offer much as there's no actionable things for when fetching a non existent path on GH (we always return null when that happens), instead it would only spawn a Sentry error. Downgrading the severity level removes the Sentry noise.

### Legal Boilerplate

### DIFF
--- a/core/commands/commit/interactors/get_file_content.py
+++ b/core/commands/commit/interactors/get_file_content.py
@@ -21,7 +21,7 @@ class GetFileContentInteractor(BaseInteractor):
             return content.get("content").decode("utf-8")
         # TODO raise this to the API so we can handle it.
         except Exception:
-            log.exception(
+            log.info(
                 "GetFileContentInteractor - exception raised",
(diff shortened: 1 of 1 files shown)


core/commands/commit/interactors/get_file_content.py
```diff
@@ -21,7 +21,7 @@ async def get_file_from_service(self, commit, path):
             return content.get("content").decode("utf-8")
         # TODO raise this to the API so we can handle it.
         except Exception:
-            log.exception(
+            log.info(
                 "GetFileContentInteractor - exception raised",
```
adrian-codecov, line 25: Should the message change from "exception raised" as well?

adrian-codecov (approved):

PR #579, opened 2024-05-22
dev: Enable most "E" rules for Ruff

ajay-sentry:
### Purpose/Motivation

Part of the Lint Enhancement epic, https://github.com/codecov/engineering-team/issues/1716, this PR aims to add a majority of the E rules to our ruff setup.

The ruff "E" rules are the "error" rules for the pycodestyle package. A majority of the fixes in this PR were for rules E721 and E722, around using type == instead of instanceof() and using bare excepts vs. explicit excepts.

### Legal Boilerplate

### DIFF
--- a/codecov_auth/helpers.py
+++ b/codecov_auth/helpers.py
@@ -52,7 +52,7 @@ class History:
         if action_flag is None:
             action_flag = CHANGE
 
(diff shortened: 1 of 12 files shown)


spalmurray-codecov (approved):
lgtm

PR #580, opened 2024-05-27
feat: Allow logout by POST request

spalmurray-codecov:
Allows logout by POST request. Part one of a three-part CSRF vuln fix with our existing GET logout strategy.

Closes https://github.com/codecov/internal-issues/issues/486

### DIFF
--- a/codecov_auth/views/logout.py
+++ b/codecov_auth/views/logout.py
@@ -1,12 +1,13 @@
 from django.conf import settings
 from django.contrib.auth import logout
+from django.http import HttpRequest
 from django.shortcuts import HttpResponse, redirect
 from rest_framework.decorators import api_view
 from rest_framework.response import Response
 
 
 @api_view(["GET", "POST"])
-def logout_view(request, **kwargs) -> HttpResponse:
+def logout_view(request: HttpRequest, **kwargs) -> HttpResponse:
     if request.method == "POST":
         response = Response(status=205)
     else:
--- a/codecov_auth/views/logout.py
+++ b/codecov_auth/views/logout.py
@@ -1,12 +1,12 @@
 from django.conf import settings
 from django.contrib.auth import logout
-from django.shortcuts import redirect
+from django.shortcuts import HttpResponse, redirect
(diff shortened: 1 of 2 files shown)


codecov_auth/views/logout.py
```diff
@@ -1,11 +1,21 @@
...
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
 
 
-def logout_view(request, **kwargs):
-    redirect_url = settings.CODECOV_DASHBOARD_URL
-    response = redirect(redirect_url)
+@api_view(["GET", "POST"])
+def logout_view(request: HttpRequest, **kwargs: str) -> HttpResponse:
+    print(kwargs)
```
ajay-sentry, line 11: reminder: remove print

codecov_auth/views/logout.py
```diff
@@ -1,11 +1,21 @@
...
+    if request.method == "POST":
+        response = Response(status=205)
```
ajay-sentry, line 13: 205 is interesting, never seen that but I like it

codecov_auth/views/logout.py
spalmurray-codecov: Good shout 💀

codecov_auth/views/logout.py
spalmurray-codecov: Lol yep. Love using http statuses to their fullest 😄 

I wanted to do a 301 and redirect to login, but I was having issues with CORS because the request includes creds.

ajay-sentry: ah yeah I wouldn't worry about that. We can also push redirects to the frontend if needed (which I think is current behavior?)

ajay-sentry (approved):

codecov_auth/views/logout.py
spalmurray-codecov: Current behavior follows the redirect returned in the this file. With the POST will have to do:
1. POST to logout
2. redirect manually to /login

but I think this is just fine personally!

PR #581, opened 2024-05-27
Bundle Analysis: Add to timeseries MeasurementName enum

JerrySentry:
Adds some timeseries measurement types to be used for various bundle analysis trend analytics.
closes: https://github.com/codecov/engineering-team/issues/1766

### Legal Boilerplate


nicholas-codecov (commented):
Should we add in an "other" just in case we run into something we didn't account for?

rohitvinnakota-codecov (approved):

adrian-codecov (approved):

PR #583, opened 2024-05-27
Add regenerate repo upload token mutation 

rohitvinnakota-codecov:
Adds a mutation that allows a user to regenerate a repository level upload token. This should replace the internal REST endpoint. A follow up PR will be made in gazebo to update hook logic to use this mutation and one more here to remove the old code once we ensure it is no longer in use. 

### Legal Boilerplate


core/commands/repository/interactors/regenerate_repository_upload_token.py
```diff
@@ -0,0 +1,25 @@
...
+from codecov.commands.base import BaseInteractor
+from codecov.commands.exceptions import ValidationError
+from codecov.db import sync_to_async
+from codecov_auth.models import Owner
+from core.models import Repository
+
+
+class RegenerateRepositoryUploadTokenInteractor(BaseInteractor):
+    @sync_to_async
+    def execute(self, repo_name: str, owner_username: str) -> uuid:
```
ajay-sentry, line 12: is uuid a type? it looks like we're setting it to the library itself

graphql_api/types/mutation/regenerate_repository_upload_token/regenerate_repository_upload_token.py
```diff
@@ -0,0 +1,26 @@
+from ariadne import UnionType, convert_kwargs_to_snake_case
```
ajay-sentry, line 1: convert_kwargs_to_snake_case looks unused here I think

core/commands/repository/interactors/regenerate_repository_upload_token.py
rohitvinnakota-codecov: updated

ajay-sentry (approved):

PR #586, opened 2024-05-29
feat: Add GQL Invoice type with related resolvers

ajay-sentry:
### Purpose/Motivation

We want to convert REST endpoints -> GQL where it makes sense. This PR aims to tackle the API side of the /invoices endpoint, creating resolvers for the related models

Stripe invoice model can be found here: https://docs.stripe.com/api/invoices

### Links to relevant tickets

Relates to https://github.com/codecov/engineering-team/issues/1471

### What does this PR do?

Aims to add Invoice Type to the owner model, and invoice resolvers for the new type created.

There are also some comments around some unused fields on the invoice model that we were returning from stripe, but don't seem to be used in Gazebo anywhere. These are definitely worth removing, since there isn't really a point in overfetching data atm.

The resolvers for the invoice model are fairly straightforward, we're just pulling off attributes from the invoice object being returned from the stripe API called in the invoice resolver on the owner model.

### Legal Boilerplate


graphql_api/types/invoice/invoice.py
```diff
@@ -0,0 +1,129 @@
...
+def resolve_invoice_due_date(invoice: Invoice, info) -> int | None:
+    return invoice["due_date"]
+
+
+@invoice_bindable.field("customerName")
+def resolve_invoice_customer_name(invoice: Invoice, info) -> str | None:
+    return invoice["customer_name"]
+
+
+# NOTE: This doesn't currently look to be used in gazebo, maybe can remove?
```
JerrySentry, line 55: Fine to remove these unused ones imo.

JerrySentry (approved):

graphql_api/types/invoice/invoice.py
ajay-sentry: Gonna do all the gut removal stuff in a followup PR if that's aight

PR #587, opened 2024-05-29
fix: Remove prefilled email for customer if exists

ajay-sentry:
### Purpose/Motivation

Customer issue around prefilled email being prefilled but incorrect, stripe docs don't seem to allow changing of prefilled email when creating stripe session so we're looking to remove the prefill altogether.

Related Docs: https://docs.stripe.com/api/checkout/sessions/create

### Links to relevant tickets

N/A

### What does this PR do?

This PR simply removes email from being passed into the stripe checkout session, but still defaulting the customer value if it exists.

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -366,12 +366,9 @@ class StripeService(AbstractPaymentService):
             extra=dict(owner_id=owner.ownerid),
         )
 
-        if not owner.stripe_customer_id:
-            customer_email = owner.email
-            customer = None
-        else:
+        customer = None
+        if owner.stripe_customer_id:
             customer = owner.stripe_customer_id
-            customer_email = None
 
         session = stripe.checkout.Session.create(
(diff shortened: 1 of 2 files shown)


services/billing.py
```diff
@@ -381,7 +378,6 @@ def create_checkout_session(self, owner: Owner, desired_plan):
             success_url=success_url,
             cancel_url=cancel_url,
             customer=customer,
```
adrian-codecov, line 380: I think you could write this as `customer = owner.stripe_customer_id`, as it will be None or the value, so 369-371 isn't needed

adrian-codecov (approved):

services/billing.py
ajay-sentry: oh yeah good point! will make the change

PR #593, opened 2024-05-31
Add commit status GQL type and filtering of commits

JerrySentry:
closes: https://github.com/codecov/engineering-team/issues/1726

Add a concept of commit status, the difference between this and commit.state is that the value of status examines all the uploads (ie sessions) belonging to the commit, will return:
- completed if all uploads are processed state
- error if any one upload is error state
- pending if any one upload is uploaded state (error has prio)

Also added a filter option `coverageStatus` to Repository.commits

### Legal Boilerplate


matt-codecov (approved):

graphql_api/actions/commits.py
```diff
@@ -45,6 +46,34 @@ def commit_uploads(commit: Commit) -> QuerySet[ReportSession]:
...
+        return None
+
+    # Only care about these 3 states, ignoring fully and partially overwritten
+    upload_states = [
+        s.state for s in sessions if s.state in ["processed", "uploaded", "error"]
+    ]
+
+    # Returning commit status in order of priority
+    # If all are processed -> COMPLETED
+    if all(state == "processed" for state in upload_states):
```
ajay-sentry, line 67: could combine this into a single pass if you want, not sure the typical size of states

return_status = CommitStatus.COMPLETED.value
for state in upload_states:
  if state == 'error':
    has_error = true
 if state == 'uploaded":
    has_pending = true

if has_error:
  return CommitStatus.ERROR.value
if has_pending:
  return CommitStatus.PENDING.value

At the very least, I don't think we need the first check since if any are error or uploaded we return early, and if not we would have gone through the whole list :)

ajay-sentry (approved):

PR #601, opened 2024-06-04
Bundle analysis: create timeseries dataset upon upload

JerrySentry:
When the upload bundle stat file is called, we will create the supported datasets for trend data. When there's values in Dataset model, measurements will be able to start populating during processing of the stat file (this will be implemented next in the worker repo). This means that for BA there will not be a opt-in feature similar to flags/component trends, all users will be automatically opted in for this. Additionally there will not be a "backfilling" process for past bundle stats.

closes https://github.com/codecov/engineering-team/issues/1771

### Legal Boilerplate


upload/views/bundle_analysis.py
```diff
@@ -142,4 +144,29 @@ def post(self, request):
                 is_shelter_request=self.is_shelter_request(),
             ),
         )
+
+        if settings.TIMESERIES_ENABLED:
```
matt-codecov, line 148: do we want to jump right to creating or do we want to create if it doesn't already exist / isn't backfilled? the latter is what is done for coverage i think https://github.com/codecov/codecov-api/blob/66ec9fa93603d062868618a140fe6c154dc74d6f/timeseries/helpers.py#L341-L348

JerrySentry: I think its better to create it as soon as possible then it is to create it when we fetch the measurements, that way we don't miss out on a datapoint in the general flow of upload->process->query since the process step is when we insert the measurement datapoints which relies on the presence of the dataset rows being inserted.

With the coverage measurements its ok because when querying and not having dataset created it can still compute and return measurements on the fly and then do a backfill. Also with BA we don't plan on doing backfills because the previous bundle schemas don't support it.

upload/tests/views/test_bundle_analysis.py
```diff
@@ -246,3 +253,98 @@ def test_upload_bundle_analysis_github_oidc_auth(
         format="json",
     )
     assert res.status_code == 201
+
+
+@pytest.mark.django_db(databases={"default", "timeseries"})
+def test_upload_bundle_analysis_measurement_datasets_created(
```
ajay-sentry, line 259: jw, is it possible to parametrize this and the other test with the pytest.mark.parametrize() decorator?


https://github.com/codecov/codecov-api/blob/7fc9bc556c81a50855f0ab8b2402fd0d85be8cc3/services/tests/test_repo_providers.py#L41

ajay-sentry (approved):
lgtm

upload/views/bundle_analysis.py
matt-codecov: oh on second read the code is `get_or_create()`, i thought it was just `create()`. my concern was that you would create multiple `Dataset`s with the same type/repoID which sounded wrong. but this is all good

matt-codecov (approved):

PR #610, opened 2024-06-07
ref: Fix GQL casing inconsistency throughout API

ajay-sentry:
### Purpose/Motivation

The default object case behavior is deprecated: [snake case fallback resolver](https://ariadnegraphql.org/docs/api-reference#snake_case_fallback_resolvers)
Ariadne calls out [convert_names_case](https://ariadnegraphql.org/docs/api-reference#make_executable_schema) as the new correct way to convert case names

This PR aims to update all the inconsistencies throughout API and consequently, this also means we can remove all the instances of @convert_kwargs_to_snake_case in a followup!

In terms of my pattern for finding areas to update:
- Fixed Failing Tests
- searched for input.get() and similar in our GQL mutations
- did a scan of our "filters" on GQL requests

Follow up ticket: https://github.com/codecov/engineering-team/issues/1879

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/1881

### What does this PR do?

PR updates all the camelCasing throughout the graphQL layer to snake_casing since we get it for free now with the new schema option

### Legal Boilerplate


graphql_api/schema.py
```diff
@@ -1,8 +1,7 @@
...
 
 from .types import bindables, types
 
-# snake_case_fallbck_resolvers gives use a default resolver which convert automatically
-# the field name from camelCase to snake_case and try to get it from the object
-# see https://ariadnegraphql.org/docs/resolvers#fallback-resolvers
-schema = make_executable_schema(types, *bindables, snake_case_fallback_resolvers)
+# convert_names_case automatically converts the field name from camelCase
+# to snake_case. See: https://ariadnegraphql.org/docs/api-reference#optional-arguments-10
+schema = make_executable_schema(types, *bindables, convert_names_case=True)
```
suejung-sentry, line 7: Nice inventorying, thanks for all the work!

Any potential cases of below? Could see them being gotchas-
* Anywhere we've appended numbers or special characters? I could see that getting resolved unexpectedly (e.g., myName0 could be `my_name_0` or `my_name0`) based on how the casing resolver wants to handle it.
* Any implicit case sensitivity handling we were getting "for free" in the old resolver that may be missing in the new? As in the old resolver was converting `MyName` --> `My_name` that now treated like `my_name` or something?

Based on the contents of this PR/ your sweep, it seems like No, but thought I'd put it out there!

ajay-sentry: Great questions! The answer to the first is no special characters or numbers, at least none that I came across while doing this investigation.

The one "potential" key that I'm a little concerned with is "sessionid," but I think since it's fully lowercase we shouldn't see any unintended side effects from this change.

In terms of keys starting with a capital, my hope would be an ESLint rule would have caught that in Gazebo prior to it getting merged 😂 Capitals in the middle of a word where we didn't expect a "_" to occur would be problematic though, we can keep an eye on this in Sentry and see if anything pops up!

suejung-sentry: Sounds good, thanks for the thorough checks!! Must have been tedious 😂

PR #613, opened 2024-06-12
gitlab oauth scope setting

JerrySentry:
Test locally with adding `gitlab_scopes` to the  `config/codecov.yml` file in umbrella-hat.

### Legal Boilerplate

### DIFF
--- a/codecov_auth/views/gitlab.py
+++ b/codecov_auth/views/gitlab.py
@@ -42,8 +42,8 @@ class GitlabLoginView(LoginMixin, StateMixin, View):
         base_url = urljoin(redirect_info["repo_service"].service_url, "oauth/authorize")
         state = self.generate_state()
 
-        scope = get_config("gitlab_scopes", default="api")
-        log.info(f"Gitlab oauth with scopes: {scope}")
+        scope = " ".join(settings.GITLAB_SCOPES)
+        log.info(f"Gitlab oauth with scope: '{scope}'")
 
         query = dict(
             response_type="code",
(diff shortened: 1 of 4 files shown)


codecov_auth/views/gitlab.py
```diff
@@ -40,11 +41,16 @@ def get_url_to_redirect_to(self):
         redirect_info = self.redirect_info
         base_url = urljoin(redirect_info["repo_service"].service_url, "oauth/authorize")
         state = self.generate_state()
+
+        scope = get_config("gitlab_scopes", default="api")
```
RulaKhaled, line 45: Let's make sure it's set correctly in gitlab config, change this to `get_config("gitlab", "scopes", default="api")` This is how codecov yaml  should look like:
Screenshot 2024-06-12 at 10 49 40 AM

RulaKhaled, line 45: Let's add this config it to settings base, and use it in redirect info, it's easier to tweak/follow convention here

JerrySentry: Trying not to set scopes like this is because this would get parsed as a list when parsing yaml file while when we set the scopes as env var it would be parsed as a string, so don't want to check for both string and list types at run time. So I'm just using a space separated string for setting the scope.

RulaKhaled (approved):
Looks good to me. One thing i would add is a test where we mock settings variable and make sure it's sent in the request. Not gonna let it block the PR though as i tested things locally.

PR #615, opened 2024-06-12
Change encode secret string resolver to mutation 

rohitvinnakota-codecov:
### Purpose/Motivation

This PR is part of transitioning our REST endpoints to GQL. This lets us handle updates in a more streamlined way with `sync_to_async` on the front end. 

### Legal Boilerplate


graphql_api/types/mutation/encode_secret_string/encode_secret_string.py
```diff
@@ -0,0 +1,24 @@
+from ariadne import UnionType
+
+from graphql_api.helpers.mutation import (
+    require_authenticated,
+    resolve_union_error_type,
+    wrap_error_handling_mutation,
+)
+
+
```
JerrySentry, line 9: Why move this to a mutation from a regular query field? afaict tell this operation isn't actually writing any data to the system.

rohitvinnakota-codecov: I agree, but it does not play well with the way we have things set up on the [front end](https://github.com/codecov/gazebo/blob/main/src/services/repo/useEncodeString.js). There is probably a way to make it work, but it will require some refactoring so I went with this approach instead(I already tested w/ staging and everything works as expected).

JerrySentry: Yeah I'm sure this works but it doesn't follow the API conventions which is not ideal. I would encourage the option to refactor the frontend if its not a lot of work, but if it's not feasible then just merge this change.

JerrySentry (approved):

PR #621, opened 2024-06-18
fix: Address Glue Updates for modifying default payment method

ajay-sentry:
### Purpose/Motivation

This updates the existing update_billing_address function to actually update the stripe default payment method billing details, as well as the customer billing address.

Stemmed from doing some local testing with the new address component where i saw the default_payment_method billing details address wasn't being updated when we were updating the users address. Now we update both the customer as well as the payment method address with this set of changes.

### Notes to Reviewer

Most of the review is UT updates, but for the stripe stuff we are doing three things:
- Retrieve the stripe customer details, and pull off the default payment method from that detail
- modify the default payment method address fields with the address fields being passed in
- update the customer object's address itself with the address fields being passed in.
  - I believe this last piece will cover us when the user wants to add or remove a payment method in the dashboard with support for example

https://github.com/codecov/codecov-api/assets/159853603/74eb44f7-7be4-4b1b-a382-73db616dc18a

### Legal Boilerplate

### DIFF
--- a/api/internal/owner/views.py
+++ b/api/internal/owner/views.py
@@ -113,8 +113,18 @@ class AccountDetailsViewSet(
         if not billing_address:
             raise ValidationError(detail="No billing_address sent")
         owner = self.get_object()
+
+        formatted_address = {
+            "line1": billing_address["line_1"],
+            "line2": billing_address["line_2"],
+            "city": billing_address["city"],
+            "state": billing_address["state"],
+            "postal_code": billing_address["postal_code"],
+            "country": billing_address["country"],
+        }
+
         billing = BillingService(requesting_user=request.current_owner)
-        billing.update_billing_address(owner, billing_address)
+        billing.update_billing_address(owner, billing_address=formatted_address)
         return Response(self.get_serializer(owner).data)
 
 
--- a/services/billing.py
+++ b/services/billing.py
@@ -403,7 +403,7 @@ class StripeService(AbstractPaymentService):
                 f"stripe_subscription_id is None, no updating card for owner {owner.ownerid}"
             )
             return None
-        # attach the payment method + set ass default on the invoice and subscription
+        # attach the payment method + set as default on the invoice and subscription
         stripe.PaymentMethod.attach(payment_method, customer=owner.stripe_customer_id)
         stripe.Customer.modify(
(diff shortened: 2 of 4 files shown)


api/internal/owner/views.py
```diff
@@ -113,8 +113,18 @@ def update_billing_address(self, request, *args, **kwargs):
         if not billing_address:
             raise ValidationError(detail="No billing_address sent")
         owner = self.get_object()
+
+        formatted_address = {
+            "line1": billing_address["line_1"],
```
adrian-codecov, line 118: Why are line1/2 camel? Shouldn't it be snake?

services/billing.py
```diff
@@ -432,15 +432,33 @@ def update_email_address(self, owner: Owner, email_address: str):
...
-                f"stripe_subscription_id is None, cannot update billing address for owner {owner.ownerid}"
+                f"stripe_customer_id is None, cannot update default billing address for owner {owner.ownerid}"
             )
             return None
-        stripe.Customer.modify(owner.stripe_customer_id, address=billing_address)
-        log.info(
-            f"Stripe successfully updated billing address for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
-        )
+
+        try:
```
adrian-codecov, line 441: This logic makes sense, but I find it weird we were originally doing it w/ subscription an not customer_id; can you think why that was the case?

api/internal/owner/views.py
ajay-sentry: ah yeah this is what I was mentioning on the applications channel; stripe considers line1 and line2 to be the correct key format instead of line_1 and line_2, which is different than our snakeify method. This seemed to be the most painless way to isolate that "rewrapping"

services/billing.py
ajay-sentry: This was on me and my lack of understanding of where the address / billing details lay among the stripe objects. After actually doing the testing locally and gaining more context on what exactly we are changing and where made me realize this update had nothing to do with the stripe subscription and everything to do with the customer.

Namely, you can have a default payment method even without having a subscription, since that's tied to the customer through payment methods ultimately.

adrian-codecov (approved):

PR #625, opened 2024-06-19
add signal to dispatch sync task to gcloud

adrian-codecov:
### Purpose/Motivation
Shelter needs access to a commit's branch. We can detect when a commit has an update (or create) and check if the branch meets the criteria for branches we want to save for tokenless in shelter. This change adds that signal to update the commit collection in GCP.

Pending on https://github.com/codecov/shared/pull/253

### Legal Boilerplate


codecov_auth/signals.py
```diff
@@ -31,7 +29,7 @@ def _get_pubsub_publisher():
 @receiver(
     post_save, sender=OrganizationLevelToken, dispatch_uid="shelter_sync_org_token"
 )
-def update_repository(sender, instance: OrganizationLevelToken, **kwargs):
+def update_org_token(sender, instance: OrganizationLevelToken, **kwargs):
```
adrian-codecov, line 32: Just renaming, was probably a copy paste that wasn't renamed at the time

JerrySentry (approved):

PR #626, opened 2024-06-19
[feat] Store "Logged in with" user metrics 

rohitvinnakota-codecov:
### Purpose/Motivation

Eng design: https://www.notion.so/sentry/Eng-Design-Tracking-Codecov-user-signup-to-first-commit-conversion-rate-6097411516f84563820276aafe3436ea

Stores metrics for when users login with a specific provider. 

### Legal Boilerplate

### DIFF
--- a/codecov_auth/views/gitlab.py
+++ b/codecov_auth/views/gitlab.py
@@ -89,7 +89,7 @@ class GitlabLoginView(LoginMixin, StateMixin, View):
         self.login_owner(user, request, response)
         self.remove_state(state, delay=5)
         UserOnboardingMetricsService.create_user_onboarding_metric(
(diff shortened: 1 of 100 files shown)


codecov_auth/views/gitlab.py
```diff
@@ -85,6 +88,10 @@ def actual_login_step(self, request):
         response = redirect(redirection_url)
         self.login_owner(user, request, response)
         self.remove_state(state, delay=5)
+        UserOnboardingMetricsService.create_user_onboarding_metric(
+            org_id=user.ownerid, event="INSTALLED_APP", payload={"login": "gitlab"}
+        )
+        print("WE PAST THI")
```
JerrySentry, line 94: print

JerrySentry (approved):

PR #629, opened 2024-06-21
fix: tokenless commit fetching

joseph-sentry:


joseph-sentry:
@adrian-codecov is the one who pointed this out to me when he was working on shelter, otherwise i would've completely missed it as well

PR #630, opened 2024-06-21
fix: Update default payment method on subscription for user

ajay-sentry:
### Purpose/Motivation

Stemmed from this sentry feedback that came in: https://sentry.slack.com/archives/C04LZKE8RMJ/p1718994052609629

Long story short, a stripe subscription will use it's default payment method if exists, before using a customer's default payment method. Because we give the default payment method to a subscription at the time of creation, when a user goes to update their default payment method it doesn't update on the subscription, leading to some weird behavior

The fix then is to also update the default payment method on the subscription to the updated payment method, which is also the payment method shown on the portal for the user.

Relevant stripe doc:
https://docs.stripe.com/api/subscriptions/update#update_subscription-default_payment_method
- Calls out when this payment method is used vs. the default on the customer

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/1996

### What does this PR do?

Adds a stripe subscription modify call to the payment method

Also adds a guard where if the customer_id doesn't exist we return early (and changes that log to a warning as well)

### Legal Boilerplate

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -397,11 +397,24 @@ class StripeService(AbstractPaymentService):
         return session["id"]
 
     @_log_stripe_error
-    def update_payment_method(self, owner, payment_method):
-        log.info(f"Stripe update payment method for owner {owner.ownerid}")
-        if owner.stripe_subscription_id is None:
-            log.info(
-                f"stripe_subscription_id is None, no updating card for owner {owner.ownerid}"
+    def update_payment_method(self, owner: Owner, payment_method):
+        log.info(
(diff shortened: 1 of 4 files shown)


services/billing.py
```diff
@@ -410,8 +423,17 @@ def update_payment_method(self, owner, payment_method):
             owner.stripe_customer_id,
             invoice_settings={"default_payment_method": payment_method},
         )
+        stripe.Subscription.modify(
```
ajay-sentry, line 426: This is the only functional update, everything else is just logs

PR #632, opened 2024-06-24
bundle analysis: update trend measurement compute

JerrySentry:
closes: https://github.com/codecov/engineering-team/issues/1976

Make it so that when fetching measurements for trends, if the start date range doesn't have any value, then we look back further until we find the first instance of a measurement point. Then we use that point's value as the value of the start date range. Otherwise it becomes ambiguous if the measurement ID has no measurement at all in history or that the start date has no measurement.

With this if the start date range is still null it means there was never any measurements before the start date. If the start date range is not null there were measurements carried over from the further past.

### Legal Boilerplate

### DIFF
--- a/graphql_api/actions/measurements.py
+++ b/graphql_api/actions/measurements.py
@@ -44,7 +44,7 @@ def measurements_by_ids(
     return measurements
 
 
-def measurements_last_uploaded_by_before_date(
+def measurements_last_uploaded_before_start_date(
     repo_id: int,
     measurable_name: str,
     measurable_ids: Iterable[str],
--- a/graphql_api/actions/measurements.py
+++ b/graphql_api/actions/measurements.py
@@ -34,7 +34,7 @@ def measurements_by_ids(
     )
 
     # group by measurable_id
-    measurements = {}
+    measurements: Mapping[int, Iterable[dict]] = {}
     for measurement in queryset:
         measurable_id = measurement["measurable_id"]
         if measurable_id not in measurements:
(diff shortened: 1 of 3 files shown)


graphql_api/actions/measurements.py
```diff
@@ -44,6 +44,28 @@ def measurements_by_ids(
     return measurements
 
 
+def measurements_last_uploaded_by_before_date(
```
adrian-codecov, line 47: a bit confused w/ the naming here since we're using start_date in the timestamp, wouldn't these be measurements_before_start_date?

services/bundle_analysis.py
```diff
@@ -368,22 +372,65 @@ def __init__(
...
+            repository=self.repository,
+            measurable_name=measurable_name,
+            measurable_ids=measurable_ids,
+            interval=self.interval,
+            after=self.after,
+            before=self.before,
+            branch=self.branch,
+        )
+
+        print("what on earth", all_measurements)
```
adrian-codecov, line 388: 👀

services/bundle_analysis.py
```diff
@@ -368,22 +372,65 @@ def __init__(
...
+                    repo_id=self.repository.repoid,
+                    measurable_name=measurable_name,
+                    measurable_ids=[measurable_id],
+                    start_date=self.after,
+                    branch=self.branch,
+                )
+
+                # There isn't any measurements before the start date range, it will just be null
+                if not carryover_measurement:
+                    continue
```
adrian-codecov, line 403: Nit: change this to "if measurements: ..." and get rid of the continue/1 line of code

graphql_api/actions/measurements.py
JerrySentry: Makes sense, renamed.

services/bundle_analysis.py
JerrySentry: 😛

services/bundle_analysis.py
JerrySentry: No need to save a carryover with null values because there's a helper function that populations non-existent measurements with null values within the query date range.

services/bundle_analysis.py
```diff
@@ -368,22 +372,61 @@ def __init__(
...
+                    measurable_name=measurable_name,
+                    measurable_ids=[measurable_id],
+                    start_date=self.after,
+                    branch=self.branch,
+                )
+
+                # Create a new datapoint in the measurements and prepend it to the existing list
+                # If there isn't any measurements before the start date range, measurements will be untouched
+                if carryover_measurement:
+                    value = (
```
adrian-codecov, line 402: This could be shortened too since the conditional is met above 😬

adrian-codecov (approved):

PR #643, opened 2024-06-27
feat: Public test results API 

RulaKhaled:
### Purpose/Motivation
Screenshot 2024-07-09 at 1 58 39 PM

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1975

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/api/public/v2/test_results/views.py
+++ b/api/public/v2/test_results/views.py
@@ -15,13 +15,18 @@ from .serializers import TestInstanceSerializer
 class TestResultsFilters(django_filters.FilterSet):
     commit_id = django_filters.CharFilter(field_name="commitid")
     outcome = django_filters.CharFilter(field_name="outcome")
-    duration_min = django_filters.NumberFilter(field_name="duration_seconds", lookup_expr="gte")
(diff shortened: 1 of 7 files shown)


api/public/v2/test_results/views.py
```diff
@@ -0,0 +1,106 @@
...
+    mixins.ListModelMixin,
+    mixins.RetrieveModelMixin,
+    RepoPropertyMixin,
+):
+    serializer_class = TestInstanceSerializer
+    permission_classes = [RepositoryArtifactPermissions]
+    filter_backends = [DjangoFilterBackend]
+    filterset_class = TestResultsFilters
+
+    def get_queryset(self):
```
JerrySentry, line 78: Would it be right if the entire function just be?
```
def get_queryset(self):
    return TestInstance.objects.filter(repoid=self.repo.repoid)
```
It should be the case that a Repository instance would always have a repoid.

Also `return TestInstance.objects.none()` would this return `None` when the callers would always expect an iterable?

RulaKhaled: makes sense!

JerrySentry (approved):

PR #649, opened 2024-06-28
bundle analysis upload endpoint add custom compare SHA

JerrySentry:
closes https://github.com/codecov/engineering-team/issues/2009

Add a `compareSha` param to the BA upload endpoint, this allows overriding default way of fetching comparison commit SHAs (ie Pull.compare_to or Commit.parent) with the current uploaded commit. This will be used as a solution for handling the "pre-merge commit" issue. This param will be consumed by the BundleAnalysisProcessor.

### Legal Boilerplate


upload/views/bundle_analysis.py
```diff
@@ -73,6 +76,9 @@ def post(self, request):
         else:
             raise NotAuthenticated()
 
+        if repo is None:
```
giovanni-guidini, line 79: The authentication classes don't check if the repo exists already?
(asking for my own understanding)

giovanni-guidini (approved):

upload/views/bundle_analysis.py
JerrySentry: I don't believe so, I just added this to appease the type checker because if the repo doesn't exist it will just report null for the repo var.

PR #655, opened 2024-07-03
feat: trigger PR sync

giovanni-guidini:
It's possible with a combination of not having webhooks and temporary errors when
fetching info from the git provider that a PR will be left forever open.

One solution is to make the PR view trigger a sync_pull task in the background.
Notice that for the person that is actually viewing the PR it will probably be out of date,
but if they refresh the page it will eventually be updated.

closes: https://github.com/codecov/internal-issues/issues/513

### DIFF
--- a/core/commands/pull/interactors/fetch_pull_request.py
+++ b/core/commands/pull/interactors/fetch_pull_request.py
@@ -1,9 +1,24 @@
+from datetime import UTC, datetime, timedelta
+
+from shared.django_apps.core.models import Pull
+
 from codecov.commands.base import BaseInteractor
 from codecov.db import sync_to_async
-from core.models import Pull
+from services.task.task import TaskService
 
 
 class FetchPullRequestInteractor(BaseInteractor):
+    def _should_sync_pull(self, pull: Pull | None) -> bool:
+        return (
+            pull is not None
+            and pull.state == "open"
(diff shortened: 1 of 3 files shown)


core/commands/pull/interactors/fetch_pull_request.py
```diff
@@ -1,9 +1,24 @@
...
+            and pull.state == "open"
+            and (datetime.now(tz=None) - pull.updatestamp) > timedelta(hours=1)
+        )
+
     @sync_to_async
     def execute(self, repository, id):
-        return repository.pull_requests.filter(pullid=id).first()
+        pull = repository.pull_requests.filter(pullid=id).first()
+        if self._should_sync_pull(pull):
+            TaskService().pulls_sync(repository.repoid, id)
```
adrian-codecov, line 22: Could the pulls_sync task leave the pull syncing indefinitely? I suppose there's a hard timeout, but trying to think if this could end up in another "stale" state

giovanni-guidini: Well even if the task does have an issue it doesn't matter for this request in particular because the calling of it only enqueues it for the workers. It's not blocking in any way.

adrian-codecov (approved):

PR #659, opened 2024-07-03
be more flexible with responses coming through repo_auth_custom_exception_handler

nora-codecov:
### Purpose/Motivation
This code introduced an issue when it was added, sentry entry [here](https://codecov.sentry.io/releases/production-release-252b887/?project=5215654). lots of responses were getting a `KeyError` because they didn't have the format that was expected. 

### What does this PR do?
Makes the logic in the function much more cautious, does several checks instead of making assumptions

### DIFF
--- a/codecov_auth/authentication/repo_auth.py
+++ b/codecov_auth/authentication/repo_auth.py
@@ -43,10 +43,17 @@ def repo_auth_custom_exception_handler(exc, context):
     give the user something better than "Invalid Token" or "Authentication credentials were not provided."
     """
     response = exception_handler(exc, context)
-    if response is not None:
+    # we were having issues with this block, I made it super cautions.
+    # Re-evaluate later whether this is overly cautious.
+    if (
+        response is not None
+        and hasattr(response, "status_code")
+        and response.status_code == 401
+        and hasattr(response, "data")
+    ):
         try:
-            exc_code = response.data["detail"].code
-        except TypeError:
+            exc_code = response.data.get("detail").code
+        except (TypeError, AttributeError):
             return response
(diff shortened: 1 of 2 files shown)


matt-codecov (approved):

codecov_auth/authentication/repo_auth.py
```diff
@@ -43,10 +43,15 @@ def repo_auth_custom_exception_handler(exc, context):
     give the user something better than "Invalid Token" or "Authentication credentials were not provided."
     """
     response = exception_handler(exc, context)
-    if response is not None:
+    if (
+        response is not None
+        and hasattr(response, "status_code")
+        and response.status_code == 401
+        and hasattr(response, "data")
+    ):
```
matt-codecov, line 51: these `hasattr`s are making me think we're sometimes getting something that is a totally different type than we're expecting. i am confused

the `KeyError`s in sentry were just about that `detail` key, were you also seeing errors about `status_code` and `data` not even existing? the link in the PR description is just to the project, not a specific issue

nora-codecov: you right -  I think lots of types of exceptions are flowing through here but this is an overreaction, to make it fully bomb-proof, so that this type of issue can't arise again. It's doing the safe checks for every element of the object that we check - just because I was thinking, what if I just patch this KeyError with `detail` and another similar issue arises with this block.

What do you think - should I pull it back to just the KeyError, or should I leave this super cautious logic?

nora-codecov: Added some in-line comments to reflect my thoughts here, but going to merge so that the issue can be fixed.

PR #670, opened 2024-07-10
feat: Update owner query with `isOktaAuthenticated` field

RulaKhaled:
### Purpose/Motivation
We need this as a flag to tell if current user is OKTA authenticated to display okta banners 

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2053

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/owner/owner.py
+++ b/graphql_api/types/owner/owner.py
@@ -251,3 +251,12 @@ def resolve_owner_invoice(
 def resolve_owner_account(owner: Owner, info) -> dict:
     account_id = owner.account_id
     return Account.objects.filter(pk=account_id).first()
+
+
+@owner_bindable.field("isUserOktaAuthenticated")
+@sync_to_async
+def resolve_is_user_okta_authenticated(owner: Owner, info) -> bool:
+    current_user = info.context["request"].user
+    if not current_user.is_authenticated:
+        return False
+    return current_user.is_okta_authenticated_for_org(owner.ownerid)
--- a/graphql_api/tests/test_owner.py
+++ b/graphql_api/tests/test_owner.py
@@ -705,3 +705,33 @@ class TestOwnerType(GraphQLTestHelper, TransactionTestCase):
         except GraphQLError as e:
             assert e.message == UnauthorizedGuestAccess.message
             assert e.extensions["code"] == UnauthorizedGuestAccess.code
+
+
+    def test_fetch_current_user_is_okta_authenticated(self):
+        account = AccountFactory()
+        current_user = OwnerFactory(username="sample-user", service="okta")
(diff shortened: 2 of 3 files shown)


graphql_api/types/owner/owner.py
```diff
@@ -274,6 +274,19 @@ def resolve_owner_account(owner: Owner, info) -> dict:
...
+@sync_to_async
+@require_part_of_org
+def resolve_is_user_okta_authenticated(owner: Owner, info) -> bool:
+    okta_signed_in_accounts = info.context["request"].session.get(
+        "okta_signed_in_accounts"
+    )
+    if not okta_signed_in_accounts:
+        return False
+
+    return owner.account_id in okta_signed_in_accounts
```
michelletran-codecov, line 287: `owner.account_id` can potentially be `None`, so we'll probably want a check on that.

If we return default, then we can simplify this a bit:
```suggestion
    okta_signed_in_accounts = info.context["request"].session.get(
        "okta_signed_in_accounts", [],
    )
    if owner.account_id:
        return owner.account_id in okta_signed_in_accounts
    return False
```

RulaKhaled: checking `if None in []` should also return False, unless we can have None in account ids session?

michelletran-codecov: We _shouldn't_ have `None` in the list, but it's likely one of the ways that this breaks (i.e. we accidentally start allowing None into the list or session corruption when loading from DB). I might be potentially overly conservative with this check, so I don't mind leaving this out.

michelletran-codecov (approved):

graphql_api/types/owner/owner.py
```diff
@@ -274,6 +274,19 @@ def resolve_owner_account(owner: Owner, info) -> dict:
     return Account.objects.filter(pk=account_id).first()
 
 
+@owner_bindable.field("isUserOktaAuthenticated")
+@sync_to_async
+@require_part_of_org
+def resolve_is_user_okta_authenticated(owner: Owner, info) -> bool:
+    okta_signed_in_accounts = info.context["request"].session.get(
+        "okta_signed_in_accounts",
```
michelletran-codecov, line 282: The session key value is incorrect. I think it's actually `okta_logged_in_accounts` (the variable is `OKTA_SIGNED_IN_ACCOUNTS`, which is bad on my part). 

I would suggest to import the variable rather than hard-coding the session key here to avoid the mismatch in different parts of the code: https://github.com/codecov/codecov-api/blob/58faac07fae36e9738f8ef0901fcf718101af67b/codecov_auth/views/okta_cloud.py#L17

(also, feel free to update the variable name or the value stored in the session, as it's inconsistent and I missed it).

adrian-codecov (approved):

PR #671, opened 2024-07-10
fix: add test results aggregate utils

joseph-sentry:
depends on: https://github.com/codecov/shared/pull/282

Helper functions for test results aggregate queries

### DIFF
--- a/utils/test_results.py
+++ b/utils/test_results.py
@@ -1,0 +1,74 @@
+import datetime as dt
+
+from django.contrib.postgres.aggregates import ArrayAgg
+from django.db.models import Avg, Case, FloatField, Value, When
+from shared.django_apps.reports.models import Test, TestInstance
+
+thirty_days_ago = dt.datetime.now(dt.UTC) - dt.timedelta(days=30)
+
+
+def aggregate_test_results(
+    repoid: int, branch: str | None = None, history: dt.timedelta | None = None
+):
+    time_ago = (
+        dt.datetime.now(dt.UTC) - dt.timedelta(days=30)
+        if history is not None
+        else thirty_days_ago
+    )
+
+    pass_failure_error_test_instances = TestInstance.objects.filter(
+        repoid=repoid,
+        created_at__gt=time_ago,
+        outcome__in=["pass", "failure", "error"],
+    )
+
+    failure_error_test_instances = pass_failure_error_test_instances.filter(
+        outcome__in=["failure", "error"]
+    )
+
+    if branch is not None:
+        pass_failure_error_test_instances = pass_failure_error_test_instances.filter(
+            branch=branch
+        )
+        failure_error_test_instances = failure_error_test_instances.filter(
+            branch=branch
+        )
+
+    failure_rates_queryset = pass_failure_error_test_instances.values(
+        "test_id"
+    ).annotate(
+        failure_rate=Avg(
+            Case(
+                When(outcome="pass", then=Value(0.0)),
+                When(outcome__in=["failure", "error"], then=Value(1.0)),
+                output_field=FloatField(),
+            )
+        )
+    )
+    failure_rate_dict = {
+        obj["test_id"]: obj["failure_rate"] for obj in failure_rates_queryset
+    }
+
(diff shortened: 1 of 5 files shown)


utils/test_results.py
```diff
@@ -0,0 +1,74 @@
...
+from django.contrib.postgres.aggregates import ArrayAgg
+from django.db.models import Avg, Case, FloatField, Value, When
+from shared.django_apps.reports.models import Test, TestInstance
+
+thirty_days_ago = dt.datetime.now(dt.UTC) - dt.timedelta(days=30)
+
+
+def aggregate_test_results(
+    repoid: int, branch: str | None = None, history: dt.timedelta | None = None
+):
```
giovanni-guidini, line 12: [nit] I'd love to see docstrings for this function (although reading through the code it's decently clear what it does. Good job with var naming

utils/test_results.py
```diff
@@ -0,0 +1,74 @@
...
+    average_duration_queryset = pass_failure_error_test_instances.values(
+        "test_id"
+    ).annotate(average_duration=Avg("duration_seconds"))
+
+    average_duration_dict = {
+        obj["test_id"]: obj["average_duration"] for obj in average_duration_queryset
+    }
+
+    tests = Test.objects.filter(repository_id=repoid).all()
+    result = dict()
```
giovanni-guidini, line 66: [kinda nit] please create a [TypedDict](https://typing.readthedocs.io/en/latest/spec/typeddict.html#typeddict) or a class for this result type (and the typehint on the function return)

utils/tests/unit/test_test_results.py
```diff
@@ -0,0 +1,144 @@
...
+        create_test_instance_func(test_2, TestInstance.Outcome.SKIP, "c", duration=1)
+        create_test_instance_func(test_2, TestInstance.Outcome.ERROR, "d", duration=1)
+        create_test_instance_func(test_2, TestInstance.Outcome.PASS, "d", duration=1)
+        create_test_instance_func(test_2, TestInstance.Outcome.FAILURE, "d", duration=1)
+        create_test_instance_func(test_2, TestInstance.Outcome.FAILURE, "d", duration=1)
+        create_test_instance_func(test_2, TestInstance.Outcome.ERROR, "e", duration=1)
+
+        d = aggregate_test_results(repo_fixture.repoid)
+
+        assert d[test_1.id]["failure_rate"] == 0.5
```
giovanni-guidini, line 30: can we have an "expected" dict and just compare both dicts? It'd be easier to understand what we actually expect :E

It also has the (small) added benefit to check that there's nothing extra that we don't expect in the response

giovanni-guidini (commented):
In general I think it looks good, but left some comments

utils/test_results.py
```diff
@@ -0,0 +1,74 @@
+import datetime as dt
+
+from django.contrib.postgres.aggregates import ArrayAgg
+from django.db.models import Avg, Case, FloatField, Value, When
+from shared.django_apps.reports.models import Test, TestInstance
+
+thirty_days_ago = dt.datetime.now(dt.UTC) - dt.timedelta(days=30)
+
+
+def aggregate_test_results(
```
RulaKhaled, line 10: How does this function work with pagination?

utils/test_results.py
```diff
@@ -0,0 +1,74 @@
...
+
+    average_duration_dict = {
+        obj["test_id"]: obj["average_duration"] for obj in average_duration_queryset
+    }
+
+    tests = Test.objects.filter(repository_id=repoid).all()
+    result = dict()
+    for test in tests:
+        result[test.id] = {
+            "failure_rate": failure_rate_dict.get(test.id),
```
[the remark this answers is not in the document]
joseph-sentry: it currently just gets the aggregated fields, I can change it so it gets all the information relevant for display

utils/test_results.py
joseph-sentry: currently it does not, but i see that we'd want to paginate by test, like get the first 10 tests then the next 10 tests

PR #680, opened 2024-07-12
Remove old REST repo endpoints and update new GQL mutations 

rohitvinnakota-codecov:
### Purpose/Motivation

This PR removes old REST endpoints for 
- EncodeSecretString
- Regenerate Repo Upload token 
- Erase Repo 

- [x] Test on staging

It also contains a fix to an edge case to update viewable repositories so the UI can see all repos in the org, not just those authored by the logged in user. 

### Legal Boilerplate


JerrySentry (approved):
These delete endpoints are not called anywhere right, like all the cloud and self hosted apps?

PR #685, opened 2024-07-16
[Okta Login] Part 2: Add Okta Login API for Organizations/Account 

michelletran-codecov:
### Purpose/Motivation
This adds the login endpoint for uses to sign into Okta for specific orgs/account.
Depends on Part 1: https://github.com/codecov/codecov-api/pull/691

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1988

### What does this PR do?
Note that this is _not_ to authenticate the user. We check that the user has already authenticated via another mechanism (i.e. Github). This adds extra permissions for organizations that want extra level of security to see private repos.

The authentication flow looks something like this:

```mermaid 
sequenceDiagram
    User->>+Codecov: /login/okta/{service}/{org}
    Codecov->>+Okta: if not signed in, then redirect them to okta.com/oauth2/v1/authorized
    Okta->>+User: shows the user the Okta sign-in UI
    User->>+Okta: enters in credentials and signs in
    Okta->>+Codecov: sends back credentials info to Codecov via /login/okta/callback
    Codecov->>+Okta: queries auth token via okta.com/oauth2/v1/token
    Codecov->>+Okta: decrypts auth token with keys okta.com/oauth2/v1/keys
    Codecov->>+User: if auth is successful, stores the authenticated account in user session
```

This PR adds the 2 endpoints:
* `/login/okta/{service}/{org}` to send the user to Okta
* `/login/okta/callback` to finish authenticating the user on Codecov

### Notes to Reviewer

* I'm not sure about the naming of OktaCloudView. Feel free to suggest another name.

### Legal Boilerplate

### DIFF
--- a/codecov_auth/tests/unit/views/test_okta_cloud.py
+++ b/codecov_auth/tests/unit/views/test_okta_cloud.py
@@ -1,0 +1,453 @@
+from logging import LogRecord
+from typing import Any
+from urllib.parse import unquote, urlencode, urlparse
+
+import pytest
+from django.test import override_settings
+from pytest import LogCaptureFixture
+from pytest_mock import MockerFixture
+from shared.django_apps.codecov_auth.models import Account, OktaSettings, Owner
+from shared.django_apps.codecov_auth.tests.factories import (
+    AccountFactory,
+    OktaSettingsFactory,
+)
+
+from codecov_auth.tests.factories import OwnerFactory
+from codecov_auth.views.okta_cloud import (
+    OKTA_CURRENT_SESSION,
+    OKTA_SIGNED_IN_ACCOUNTS_SESSION_KEY,
+)
+from codecov_auth.views.okta_mixin import OktaIdTokenPayload
+from utils.test_utils import Client as TestClient
+
+
+@pytest.fixture
+def signed_in_client() -> TestClient:
+    new_client = TestClient()
+    new_client.force_login_owner(OwnerFactory())
+    return new_client
+
+
+@pytest.fixture
+def okta_org_name() -> str:
+    return "foo-bar-organization"
+
+
+@pytest.fixture
+def okta_org(okta_org_name: str) -> Owner:
+    org: Owner = OwnerFactory.create(username=okta_org_name, service="github")
+    org.save()
+    return org
+
+
+@pytest.fixture
+def okta_account(okta_org: Owner):
+    account = AccountFactory()
+    okta_org.account = account
+    okta_org.save()
+
+    okta_settings: OktaSettings = OktaSettingsFactory(account=account)
+    okta_settings.url = "https://foo-bar.okta.com"
+    okta_settings.save()
+    return account
+
+
+@pytest.fixture
+def mocked_okta_token_request(mocker):
+    return mocker.patch(
+        "codecov_auth.views.okta_mixin.requests.post",
+        return_value=mocker.MagicMock(
+            status_code=200,
+            json=mocker.MagicMock(
+                return_value={
+                    "access_token": "test-access-token",
+                    "refresh_token": "test-refresh-token",
+                    "id_token": "test-id-token",
+                    "state": "test-state",
+                },
+            ),
+        ),
+    )
+
+
+@pytest.fixture
+def mocked_validate_id_token(mocker):
+    return mocker.patch(
+        "codecov_auth.views.okta_cloud.validate_id_token",
+        return_value=OktaIdTokenPayload(
+            sub="test-id",
+            email="test@example.com",
+            name="Some User",
+            iss="https://example.com",
+            aud="test-client-id",
+        ),
+    )
+
+
+def log_message_exists(message: str, logs: list[LogRecord]) -> bool:
+    """Helper method to check that a particular log record was emitted"""
+    for log in logs:
+        if log.message == message:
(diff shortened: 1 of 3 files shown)


codecov_auth/views/okta_cloud.py
```diff
@@ -0,0 +1,200 @@
...
+# This is so that the callback can reference the orgs/accounts that we're
+# signing in for.
+OKTA_CURRENT_SESSION = "okta_current_session"
+
+log = logging.getLogger(__name__)
+
+
+def get_app_redirect_url(org_username: str, service: str) -> str:
+    """The Codecov app page we redirect users to."""
+    return f"{settings.CODECOV_DASHBOARD_URL}/{service}/{org_username}"
```
RulaKhaled, line 29: question, will the client redirect to `/login/okta` or this?

RulaKhaled (commented):
It all looks good to me! let's test it today and we shall merge it after

codecov_auth/views/okta_cloud.py
```diff
@@ -0,0 +1,200 @@
...
+                "okta_settings_id": okta_settings.id,
+            }
+            return consent
+
+
+class OktaCloudCallbackView(OktaLoginMixin, View):
+    service = "okta_cloud"
+
+    def get(self, request: HttpRequest) -> HttpResponse:
+        current_okta_session: dict[str, int] = request.session.get(OKTA_CURRENT_SESSION)
```
Swatinem, line 107: as you introduced pylance in part 1, would it make sense to use a stricter checked type here rather than `dict[str, int]` (which is also missing a `| None`)?

codecov_auth/tests/unit/views/test_okta_cloud.py
```diff
@@ -0,0 +1,453 @@
...
+def test_okta_login_unauthenticated_user(
+    client: TestClient,
+    caplog: LogCaptureFixture,
+):
+    res = client.get("/login/okta/github/some-unknown-service")
+    assert log_message_exists(
+        "User needs to be signed in before authenticating organization with Okta.",
+        caplog.records,
+    )
+    assert res.status_code == 403
```
Swatinem, line 105: does it make sense to also (or instead) include those messages as part of the response, instead of just making those generic 403 pages?

codecov_auth/tests/unit/views/test_okta_cloud.py
```diff
@@ -0,0 +1,453 @@
...
+    assert res.status_code == 404
+
+
+@pytest.mark.django_db
+def test_okta_login_no_account(signed_in_client: TestClient, caplog: LogCaptureFixture):
+    org: Owner = OwnerFactory.create(username="org-no-account", service="github")
+    org.save()
+    res = signed_in_client.get("/login/okta/github/org-no-account")
+    assert log_message_exists(
+        "Okta settings not found. Cannot sign into Okta", caplog.records
```
Swatinem, line 124: this is a different condition from the one below?

Swatinem (approved):
lgtm, some general suggestions / comments:

- I’m by far no expert on django internals, but isn’t it possible to have some kind of automatic decorators that resolve url parameters automatically, so you don’t have to query / check for a valid `org` manually?
- You use logging and assert log messages for certain outcomes. IMO it would make sense to just put those things into the response and assert that instead.
- Is it even possible and/or desirable to test against "real world" okta? pro: it creates more confidence than a mocked response (which itself could be broken), but also con: now your CI depends on an external service which might be slow and flaky

codecov_auth/views/okta_cloud.py
michelletran-codecov: Ah whoops good catch! I'll fix this.

> would it make sense to use a stricter checked type here rather than dict[str, int]

Unfortunately, `request.session` needs to be serializable (so can't just store the models directly). I considered creating a Pydantic model to dump back into a dict to go into a session (i.e. `okta_session.model_dump()`) but thought that this object might be too simple to need several rounds of conversion. Thoughts on whether it's worth going back to this?

codecov_auth/views/okta_cloud.py
```diff
@@ -0,0 +1,200 @@
...
+            organization: Owner = Owner.objects.get(
+                service=service, username=org_username
+            )
+        except Owner.DoesNotExist:
+            log.warning("The organization doesn't exist.", extra=log_context)
+            return HttpResponse(status=404)
+
+        okta_settings = get_okta_settings(organization)
+        if not okta_settings:
+            log.warning(
```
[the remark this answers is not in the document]
michelletran-codecov: Ah good point. There's no point logging it twice, so I'll remove the one in `get_okta_settings`.

codecov_auth/tests/unit/views/test_okta_cloud.py
michelletran-codecov: Hmm, it's a tradeoff between giving users too much info for attackers to learn about our system and usability for users. For authentication APIs, I would lean more conservatively towards exposing less information automatically to users (and adjust based on user feedback). For example, the log message above exposes that Codecov has a 2 step authentication process involving Okta, and that this endpoint is supposed to authenticate users for specific organizations (rather than a user generically).

nora-codecov (approved):

PR #691, opened 2024-07-18
[Okta Login] Part 1: Refactoring existing Okta code

michelletran-codecov:
### Purpose/Motivation
Refactoring this so that it is more easily shared. Split this from [part 2](https://github.com/codecov/codecov-api/pull/685) to make reviewing both parts easier. See Part 2 for more context in how this refactor will be used.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/1988

### What does this PR do?
This moves some of the shared code that will be common in all Okta implementations to a new `mixin` that can be shared across both Views. 

### Notes to Reviewer
* I added Pydantic to better type check the responses coming from Okta (and better document what is in the responses).

### Legal Boilerplate

### DIFF
--- a/codecov_auth/tests/unit/views/test_okta_mixin.py
+++ b/codecov_auth/tests/unit/views/test_okta_mixin.py
@@ -1,0 +1,110 @@
+from unittest.mock import MagicMock, patch
+
+import jwt
+
+from codecov_auth.views.okta import OKTA_BASIC_AUTH, OktaEnterpriseLoginView
+from codecov_auth.views.okta_mixin import validate_id_token
+
+private_key = """-----BEGIN PRIVATE KEY-----
+MIIEvwIBADANBgkqhkiG9w0BAQEFAASCBKkwggSlAgEAAoIBAQC7VJTUt9Us8cKj
+MzEfYyjiWA4R4/M2bS1GB4t7NXp98C3SC6dVMvDuictGeurT8jNbvJZHtCSuYEvu
+NMoSfm76oqFvAp8Gy0iz5sxjZmSnXyCdPEovGhLa0VzMaQ8s+CLOyS56YyCFGeJZ
+qgtzJ6GR3eqoYSW9b9UMvkBpZODSctWSNGj3P7jRFDO5VoTwCQAWbFnOjDfH5Ulg
+p2PKSQnSJP3AJLQNFNe7br1XbrhV//eO+t51mIpGSDCUv3E0DDFcWDTH9cXDTTlR
+ZVEiR2BwpZOOkE/Z0/BVnhZYL71oZV34bKfWjQIt6V/isSMahdsAASACp4ZTGtwi
+VuNd9tybAgMBAAECggEBAKTmjaS6tkK8BlPXClTQ2vpz/N6uxDeS35mXpqasqskV
+laAidgg/sWqpjXDbXr93otIMLlWsM+X0CqMDgSXKejLS2jx4GDjI1ZTXg++0AMJ8
+sJ74pWzVDOfmCEQ/7wXs3+cbnXhKriO8Z036q92Qc1+N87SI38nkGa0ABH9CN83H
+mQqt4fB7UdHzuIRe/me2PGhIq5ZBzj6h3BpoPGzEP+x3l9YmK8t/1cN0pqI+dQwY
+dgfGjackLu/2qH80MCF7IyQaseZUOJyKrCLtSD/Iixv/hzDEUPfOCjFDgTpzf3cw
+ta8+oE4wHCo1iI1/4TlPkwmXx4qSXtmw4aQPz7IDQvECgYEA8KNThCO2gsC2I9PQ
+DM/8Cw0O983WCDY+oi+7JPiNAJwv5DYBqEZB1QYdj06YD16XlC/HAZMsMku1na2T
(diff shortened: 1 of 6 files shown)


nora-codecov (approved):
great cleanup and reorganization!

codecov_auth/tests/unit/views/test_okta_mixin.py
```diff
@@ -0,0 +1,110 @@
+from unittest.mock import MagicMock, patch
+
+import jwt
+
+from codecov_auth.views.okta import OKTA_BASIC_AUTH, OktaLoginView
+from codecov_auth.views.okta_mixin import validate_id_token
+
+private_key = """-----BEGIN PRIVATE KEY-----
```
adrian-codecov, line 8: Is this a legit private key? I doubt it is, but would you be able to name this something like "fake_private_key" or something? We get pinged at times with potential security vulns that see things like this make it to prod, so this would make it very obvious if the security team or someone asks in retrospect

codecov_auth/views/okta_mixin.py
```diff
@@ -0,0 +1,112 @@
...
+from requests.auth import HTTPBasicAuth
+
+from codecov_auth.views.base import StateMixin
+
+log = logging.getLogger(__name__)
+
+ISS_REGEX = re.compile(r"https://[\w\d\-\_]+.okta.com/?")
+
+
+class OktaTokenResponse(pydantic.BaseModel):
```
adrian-codecov, line 20: What are your thoughts on classes vs dataclasses? I tend to go back and forth w/ dataclasses; my goal at the end of the day is to know what an object's shape is going to look like, and this does this which is totally fine. Curious to hear if you have a strong opinion for/against the way you type things

adrian-codecov (approved):

codecov_auth/views/okta.py
```diff
@@ -179,15 +114,29 @@ def _login_user(self, request: HttpRequest, iss: str, user_data: dict):
...
+            return None
+        if not ISS_REGEX.match(iss):
+            log.warning("Invalid Okta issuer")
+            return None
+        return iss
+
+    def get(self, request: HttpRequest) -> HttpResponse:
+        iss = self.validate_issuer()
+        if iss is None:
+            return redirect(f"{settings.CODECOV_DASHBOARD_URL}/login")
```
Swatinem, line 131: If I read this right, the only variable here is `settings.OKTA_ISS`, which I believe should be an immutable deployment configuration.

Does it make sense to run the validation on every request in that case?
Also, does it make sense to do a redirect to `/login` in case okta is misconfigured? Although whether you display a "okta login is not enabled" error page, or redirect to the login page does not matter that much.

From a SaaS product perspective: is there a hard SLO for the okta integration? In which case it would make sense to assert on deployment that a valid issuer is configured.

---

Long story short, I see you are just moving existing code around, so this is not a regression in itself, just some in-depth questions I’m asking :-)

codecov_auth/views/okta_mixin.py
Swatinem: I’m also a bit surprised to see it like this, as it is pulling in `pydantic` as external dependency. Does this offer some more safety features vs just using plain `mypy`?

Swatinem (approved):

codecov_auth/views/okta_mixin.py
michelletran-codecov: Pydantic also does runtime type checking (i.e. ensures that 1. the fields exist in the response and that 2. that they're string typed, or can be strings). `dataclass.dataclasses` doesn't do type checking (it just defines the structure) and `mypy` only checks for types when you run `mypy` (or "compile" time type checking... pydantic is more similar to rust's [serde lib](https://serde.rs/)). `pydantic` _might_ be overkill for this, but I ended up using this because `dataclass.dataclass` doesn't let me ignore fields that I don't care about (and I didn't want to recreate the whole datastructure of the response). In terms of why I used `BaseModel` instead of `pydantic.dataclasses`... I don't really have a good reason other than it's probably a bit more flexible? Pydantic `BaseModels` offer JSON dumping, which is also kind of useful if we want to convert back to `dicts` (i.e. sessions need to store serialized objects, for instance). [Here](https://docs.pydantic.dev/latest/concepts/dataclasses/) is a doc on some of the differences.

codecov_auth/views/okta.py
michelletran-codecov: > Does it make sense to run the validation on every request in that case?

Probably not :p Ideally, we would just check this on initialization and use it as it for the rest of the requests. I didn't want to modify this (because I was unsure about the consequences), which is why I just took the conservative approach to leave it for now (especially this isn't really related to my goal).

> Also, does it make sense to do a redirect to /login in case okta is misconfigured?

It's going to be hard to tell from the Okta response whether it's mis-configured. For admins that are integrating with Okta, they will probably need to test it out before deploying it. We probably don't want to leak that the integration is mis-configured to non-admin users though, so redirecting them to the login page is probably fine.

> From a SaaS product perspective: is there a hard SLO for the okta integration?

AFAICT, this Okta integration is _only_ for self-hosted. We don't support this on Codecov's core cloud platform, so I doubt there's any SLAs around this. The Okta integration that I'm adding in Part 2 will be different though, and I think we're potentially going to provide feedback to the users integrating on the admin page if their login integration has worked or not. Not sure if we're going to add an SLA around this, but hopefully there will be more feedback.

PR #696, opened 2024-07-18
add action to link users to accounts

nora-codecov:
### Purpose/Motivation
Gives us a method to do the initial setup when an Account is created. This action will create the link between the Account object and the User object for any Owner who is a `plan_activated_user` on any of the Organizations linked to the Account.

Can be run multiple times safely (in case it errors or times out)
Won't do the link if the Account doesn't have enough seats

For "auditing" its behavior - it gives messages about how many Users and AccountUsers were created. You can see the Users linked to the Account on the Account edit page (when you click into an Account on the admin). You can remove a AccountUsers connection on the User admin page.

### Links to relevant tickets
Uses https://github.com/codecov/shared/pull/295
part of https://github.com/codecov/engineering-team/issues/2058

### DIFF
--- a/codecov_auth/admin.py
+++ b/codecov_auth/admin.py
@@ -1,3 +1,4 @@
+from gettext import gettext, ngettext
 from typing import Optional
 
 import django.forms as forms
@@ -297,6 +298,7 @@ class AccountAdmin(AdminMixin, admin.ModelAdmin):
     search_fields = ("name__iregex", "id")
     search_help_text = "Search by name (can use regex), or id (exact)"
     inlines = [OwnerOrgInline, StripeBillingInline, InvoiceBillingInline]
+    actions = ["link_users_to_account"]
 
     readonly_fields = ["id", "created_at", "updated_at", "users"]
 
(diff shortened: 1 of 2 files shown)


codecov_auth/admin.py
```diff
@@ -310,6 +311,66 @@ class AccountAdmin(AdminMixin, admin.ModelAdmin):
         "is_delinquent",
     ]
 
+    @admin.action(description="Link Users to Account")
+    def link_users_to_account(self, request, queryset):
+        for account in queryset:
+            all_plan_activated_user_ids = set()
```
adrian-codecov, line 317: Nit/Optional: I find the "all" a bit redundant since the variable is plural. I also got tripped up a bit reading this like "(all plan) (activated user ids)" instead  of "(all) (plan activated user ids)" 😅, but that might be me. I think calling this "activated_user_ids_in_orgs" represents this a bit better? 

Similar idea for the rest of the "alls" as well

codecov_auth/admin.py
```diff
@@ -310,6 +311,66 @@ class AccountAdmin(AdminMixin, admin.ModelAdmin):
...
+            for org in account.organizations.all():
+                all_plan_activated_user_ids.update(set(org.plan_activated_users))
+
+            all_plan_activated_users = Owner.objects.filter(
+                ownerid__in=all_plan_activated_user_ids
+            ).prefetch_related("user")
+            non_student_count = all_plan_activated_users.exclude(student=True).count()
+            total_seats_for_account = account.plan_seat_count + account.free_seat_count
```
adrian-codecov, line 325: What's a free seat count? I see it's defaulted to 0, is this for like free plans?

codecov_auth/admin.py
```diff
@@ -310,6 +311,66 @@ class AccountAdmin(AdminMixin, admin.ModelAdmin):
...
+    def link_users_to_account(self, request, queryset):
+        for account in queryset:
+            all_plan_activated_user_ids = set()
+            for org in account.organizations.all():
+                all_plan_activated_user_ids.update(set(org.plan_activated_users))
+
+            all_plan_activated_users = Owner.objects.filter(
+                ownerid__in=all_plan_activated_user_ids
+            ).prefetch_related("user")
+            non_student_count = all_plan_activated_users.exclude(student=True).count()
```
adrian-codecov, line 324: Optional: maybe adding a comment on why we exclude students. Thanks for being so attentive and catching this this early, this is a edge case for suuuure

codecov_auth/admin.py
```diff
@@ -310,6 +311,66 @@ class AccountAdmin(AdminMixin, admin.ModelAdmin):
...
+                    messages.ERROR,
+                )
+                return
+
+            owners_without_user_objects = all_plan_activated_users.filter(
+                user__isnull=True
+            )
+            userless_owners = []
+            for userless_owner in owners_without_user_objects:
+                new_user = User.objects.create()
```
adrian-codecov, line 340: Is this new User object meant to be empty?

codecov_auth/admin.py
```diff
@@ -310,6 +311,66 @@ class AccountAdmin(AdminMixin, admin.ModelAdmin):
...
+                userless_owner.user = new_user
+                userless_owners.append(userless_owner)
+            total = Owner.objects.bulk_update(userless_owners, ["user"])
+            self.message_user(
+                request,
+                f"Created a User for {total} Owners",
```
adrian-codecov, line 346: Would probably be handy to log the owner ids we did this for 👀, up to you

adrian-codecov (commented):
Left some comments, overall lgtm! Lmk and I can approve

michelletran-codecov (approved):
Generally LGTM! This is only for adding orgs for now, correct? We would need some different mechanism if we were to remove orgs from accounts...

codecov_auth/admin.py
nora-codecov: it's a thing that exists on Owners currently (named `free`), but it's a way for us to give Orgs additional seats not in the count they pay for (`plan_seat_count`) 🤷

codecov_auth/admin.py
nora-codecov: hmmm, not really an edge case - check out `activated_user_count` or `can_activate_user` methods on `Owner` and `Account` - it's our built in policy.
I'm following the methods defined on the models to calculate things like `can_activate_user` - but having to write it out instead of using the pre-built methods so that I can do bulk operations 👍

codecov_auth/admin.py
nora-codecov: no! fixed!
ty ty ty

codecov_auth/admin.py
adrian-codecov: damn, was not aware of this

codecov_auth/admin.py
adrian-codecov: Right, what I meant was that this is an easy thing to slip since the majority of our customers aren't students, so I was praising you for being diligent, looking at previous code and learning about our policy

adrian-codecov (approved):

PR #697, opened 2024-07-19
Upgrade old packages that relied on distutils

michelletran-codecov:
The following libraries were causing problems with the recent setuptools upgrade:

* cerberus
* django-model-utils
* django(?)

setuptools 71.0.0 removed its own vendoring of `distutils` (ref)[https://setuptools.pypa.io/en/stable/history.html#v71-0-0] and that caused any dependencies that we had that referenced `distutils` to break. So upgraded those dependencies to fix the problem.

(not ready, need to unpin a `shared` dependency).

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


docker/Dockerfile
```diff
@@ -8,7 +8,7 @@ FROM us-docker.pkg.dev/berglas/berglas/berglas:$BERGLAS_VERSION as berglas
 FROM $REQUIREMENTS_IMAGE as app
 COPY . /app
 WORKDIR /app
-RUN pip install setuptools==70.3.0
+RUN pip install setuptools==71.1.0
```
joseph-sentry, line 11: is there a specific reason we want to upgrade setuptools to this version or is this just to be "healthy" and keep up with new versions as they come out?

joseph-sentry (approved):

docker/Dockerfile
michelletran-codecov: This is the newest version available for `setuptools` and I want us to be at least on the newest version. Pinning this will require us to be proactive about updating `setuptools` when new versions come out. I don't think that Dependabot is smart enough to pick up dep updates from Dockerfiles, so we will need to be vigilant about updating `setuptools` if we want to continue to be "up to date" on our dependencies (which is the tradeoff with pinning). On the plus side, if `setuptools` was pinned in the first place, we wouldn't have spend an afternoon of a bunch of engineer hours on this problem. But maybe the problem is actually deeper... we should be more proactive about upgrading our dependencies in general, and that `setuptools` broke in this way is a symptom of that. 

Anyways, tl;dr is that I'm pinning the latest stable version which feels like the right thing to do (for stability, health, keep things working etc.), but it means that we need to be proactive about updating this version because it won't be automatically updated for us now.

ajay-sentry (approved):

PR #707, opened 2024-07-24
feat: Support ordering in test results

RulaKhaled:
### Purpose/Motivation
needed for failed tests. 
ps. i didn't add tests for updated at because it's a pain and it's not worth it.

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/enums/enums.py
+++ b/graphql_api/types/enums/enums.py
@@ -12,7 +12,10 @@ class OrderingParameter(enum.Enum):
 
 
 class TestResultsOrderingParameter(enum.Enum):
-    NAME = "name"
+    AVG_DURATION = "avg_duration"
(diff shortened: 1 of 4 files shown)


graphql_api/types/enums/enums.py
```diff
@@ -12,7 +12,10 @@ class OrderingParameter(enum.Enum):
 
 
 class TestResultsOrderingParameter(enum.Enum):
-    NAME = "name"
+    AVG_DURATION = "avg_duration"
+    FAILURE_RATE = "failure_rate"
+    COMMITS_WHERE_FAIL = "commits_where_fail"
```
ajay-sentry, line 17: nit: can we just call this failed_commits?

ajay-sentry (approved):
looks good

PR #709, opened 2024-07-24
clean up

nora-codecov:
### Purpose/Motivation
Removing unused model that shares the same name as the new model I've added


michelletran-codecov (approved):
Yay cleanup!

PR #711, opened 2024-07-24
Filter out private repos with Enforced Okta setting

michelletran-codecov:
### Purpose/Motivation
For organizations with `enforced` Okta configurations, we only want to show the public repos for the user. They will need to be authenticated to get access to private repos on our UI.

### Links to relevant tickets
ref: https://github.com/codecov/engineering-team/issues/1992

### What does this PR do?
Filters out repos on the GraphQL endpoint for orgs that the user is not authenticated with for various queries that requests repository information.

### Notes to Reviewer
We aren't filtering out the repos for public API endpoints. This is because these endpoints auth through token authentication are (likely, sometimes) to be used in some automated pipeline where Okta doesn't make sense.

I would like some feedback on whether there's any other API endpoints I'm missing that should have repos filtered. 

### Legal Boilerplate

### DIFF
--- a/core/commands/repository/interactors/fetch_repository.py
+++ b/core/commands/repository/interactors/fetch_repository.py
@@ -1,14 +1,27 @@
+from shared.django_apps.codecov_auth.models import Owner
+from shared.django_apps.core.models import Repository
+
 from codecov.commands.base import BaseInteractor
 from codecov.db import sync_to_async
-from core.models import Repository
 
 
 class FetchRepositoryInteractor(BaseInteractor):
     @sync_to_async
-    def execute(self, owner, name):
+    def execute(
+        self,
+        owner: Owner,
+        name: str,
+        okta_authenticated_accounts: list[int],
+        exclude_okta_enforced_repos: bool = True,
+    ):
+        queryset = Repository.objects.viewable_repos(self.current_owner)
+        if exclude_okta_enforced_repos:
+            queryset = queryset.exclude_accounts_enforced_okta(
+                okta_authenticated_accounts
+            )
+
         return (
-            Repository.objects.viewable_repos(self.current_owner)
-            .filter(author=owner, name=name)
+            queryset.filter(author=owner, name=name)
             .with_recent_coverage()
             .with_oldest_commit_at()
             .select_related("author")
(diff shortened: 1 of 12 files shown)


graphql_api/tests/helper.py
```diff
@@ -20,10 +20,10 @@ def gql_request(
             self.client = Client()
             self.client.force_login_owner(owner)
 
-            if okta_signed_in_accounts:
-                session = self.client.session
-                session[OKTA_SIGNED_IN_ACCOUNTS_SESSION_KEY] = okta_signed_in_accounts
-                session.save()
+        if okta_signed_in_accounts:
```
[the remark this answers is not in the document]
michelletran-codecov: Ah whoops, it isn't required. I'll removed it. It's from an older version that I rebased off. Thanks for the catch!

core/commands/repository/interactors/fetch_repository.py
```diff
@@ -1,14 +1,27 @@
...
     @sync_to_async
-    def execute(self, owner, name):
+    def execute(
+        self,
+        owner: Owner,
+        name: str,
+        okta_authenticated_accounts: list[int],
+        exclude_okta_enforced_repos: bool = True,
+    ):
+        queryset = Repository.objects.viewable_repos(self.current_owner)
```
adrian-codecov, line 17: I'm curious if it would make sense for viewable_repos itself to have this kind of check. This works for now, I'm wondering though if we're going to have to exclude these on a "resolver to resolver" basis you know?

adrian-codecov (approved):
I think you got everywhere we fetch repos for gazebo!

core/commands/repository/interactors/fetch_repository.py
michelletran-codecov: Having the `viewable_repos` not exclude okta is pretty convenient for being more specific about how to filter repos. For instance, the `RepositoryViewSetMixin` calls `viewable_repos` but without `exclude_accounts_enforced_okta`. The only reason why I have this conditional here is because this interactor is shared with the legacy upload API. Otherwise, if this interactor is only for GraphQL, then we can just have `exclude_accounts_enforced_okta` by default.

PR #712, opened 2024-07-25
add Account to Owner Serializer

nora-codecov:
### Purpose/Motivation
Provide read-only access to an Org's Account (if they have one)

### Links to relevant tickets
Part of https://github.com/codecov/engineering-team/issues/2060

### What does this PR do?
If your Org gets the "Enterprise Experience", you can access the details about your `Account` via the `account-details` endpoint

### DIFF
--- a/api/internal/account/serializers.py
+++ b/api/internal/account/serializers.py
@@ -1,0 +1,47 @@
+from rest_framework import serializers
+from shared.django_apps.codecov_auth.models import (
+    Account,
+    InvoiceBilling,
+    StripeBilling,
+)
+
+from api.internal.owner.serializers import PlanSerializer
+
+
+class StripeBillingSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = StripeBilling
+        fields = ("customer_id", "subscription_id", "is_active")
+
+
+class InvoiceBillingSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = InvoiceBilling
+        fields = ("account_manager", "is_active")
+
+
+class BaseAccountSerializer(serializers.ModelSerializer):
+    stripe_billing = StripeBillingSerializer(many=True, read_only=True)
+    invoice_billing = InvoiceBillingSerializer(many=True, read_only=True)
+    plan = PlanSerializer(source="pretty_plan")
+
+    class Meta:
+        model = Account
+        read_only_fields = (
+            "name",
+            "is_delinquent",
+            "free_seat_count",
+            "plan_seat_count",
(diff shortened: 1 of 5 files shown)


api/internal/account/serializers.py
```diff
@@ -0,0 +1,47 @@
...
+        fields = ("account_manager", "invoice_notes", "is_active")
+
+
+class BaseAccountSerializer(serializers.ModelSerializer):
+    stripe_billing = StripeBillingSerializer(many=True, read_only=True)
+    invoice_billing = InvoiceBillingSerializer(many=True, read_only=True)
+    plan = PlanSerializer(source="pretty_plan")
+
+    class Meta:
+        model = Account
```
adrian-codecov, line 29: Curious if there's a reason why we want to expose this via this serializer (which ends up being REST) vs exposing it via graphql? 

We've had this "unspoken" desire to move off REST and do graphql if this is being surfaced to gazebo, so wondering if there was another use case for it

api/internal/owner/serializers.py
```diff
@@ -246,18 +246,20 @@ class RootOrganizationSerializer(serializers.Serializer):
 
 
 class AccountDetailsSerializer(serializers.ModelSerializer):
+    # circular imports
+    from api.internal.account.serializers import BaseAccountSerializer
```
adrian-codecov, line 250: This leads to circular imports? 💀

nora-codecov: yeah, the location and naming of serializers is a bit of a mess, but the `PlanSerializer` is defined here (in the `Owners` file 💀 ), and I want to use it on `BaseAccountSerializer` (it's important that the Account `plan` is serialized the same as the Org `plan` so that one can replace the other), which gets circular since I want account on `AccountDetailsSerializer` (which is actually a `ModelSerializer` for `Owners` 💀 💀 💀 )

api/internal/account/serializers.py
adrian-codecov: Gotchaa, yeah those are valid points. I think the end goal here is to make resolvers for this (resolvers are mostly done by the application team) as we'll move off from rest, but I don't want to stop you from delivering this change. Since these aren't used in gazebo by now, I'd say part of the req for the applications to use this is to make resolvers as an active effort to transition off of rest - some plan related things are available in gql like the `plan` resolver, but I'm not expecting you to make the resolvers now and here (unless you're interested in learning and I can help you 🙌, it should be fairly easy to expose these via gql). 

If you merge this, could you make tickets in the apps team to consume these fields via gql?

adrian-codecov (approved):

PR #729, opened 2024-08-01
Sync account when creating in Okta

michelletran-codecov:
### Purpose/Motivation
This just syncs a bunch of other fields for accounts created during the Okta settings creation process so that the account objects make more sense.

### Legal Boilerplate


codecov_auth/commands/owner/interactors/save_okta_config.py
```diff
@@ -44,11 +44,23 @@ def execute(self, input: dict):
...
+            account = Account.objects.create(
+                name=owner.username,
+                plan=owner.plan,
+                plan_seat_count=owner.plan_user_count,
+                free_seat_count=owner.free,
+                plan_auto_activate=owner.plan_auto_activate,
+            )
             account_id = account.id
             owner.account_id = account_id
             owner.save()
```
nora-codecov, line 56:
```suggestion
            owner.account = account
            owner.save()
```

codecov_auth/commands/owner/interactors/save_okta_config.py
```diff
@@ -44,11 +44,23 @@ def execute(self, input: dict):
...
 
+            # Update the activated users to be added to the account
+            plan_activated_user_owners: list[int] = owner.plan_activated_users
+            for activated_user_owner in plan_activated_user_owners:
+                user_owner: Owner = Owner.objects.get(pk=activated_user_owner)
+                account.activate_owner_user_onto_account(user_owner)
+
```
nora-codecov, line 63: might be nice to do this in a `bulk_create`?
example here https://github.com/codecov/codecov-api/blob/e245d323e0fb96f35eeafb3209c3e167c53e30b6/codecov_auth/admin.py#L451
or `bulk_update`, example here https://github.com/codecov/codecov-api/blob/e245d323e0fb96f35eeafb3209c3e167c53e30b6/codecov_auth/admin.py#L402

positive: all these `AccountsUsers` would be created in 1 query, rather than 1 per `Owner` (or maybe 2 per `Owner` since it checks and adds a `User`)
negative: you have to hold the objects in local memory, so potentially all new AccountsUsers objects that you create here. Can do some clever stuff with cursors to break into chunks if you want to look into that.

I'm not sure exactly where/how this `interactor` is used and whether it is worth the added effort to make it more efficient, so I'll leave it up to you, but feel free to slack me if you wanna chat :)

nora-codecov (approved):

codecov_auth/commands/owner/interactors/save_okta_config.py
michelletran-codecov: Oo, good point about using bulk insert. I think it might also be worth maybe adding a metric around this (just in case we have lot of users. I'll probably want to add a limit to the batch size, so I'll do that. Thanks for the suggestion!

michelletran-codecov: Updated this to use bulk insert of 50 batches (only for AccountsUsers, I'm creating Users one by one). Decided not to add a metric for now because it would be pretty sparse.

PR #735, opened 2024-08-05
dev: Add Rate limits for rest endpoints

ajay-sentry:
### Purpose/Motivation

This PR aims to update our DRF (Django Rest Framework) configuration to also include a basic ratelimiter. Rudimentarily, it uses the default "anon" and "user" throttle classes that DRF gives us out the box and sets the request rate to 30/min for anon requests and 90 for user requests. Additionally, we have a "sustained" rate limit of 1000/day for anons and 2000/day for users

Testing it locally with 10x reduction in the above values and it seems to work pretty well. We should be able to track all the instances of 429s coming to sentry and create charts / graphs as well. Maybe even alerts too if there's a way to capture the IP.

It should also be noted that the rate limit currently is shared across ALL endpoints. so the user gets a "90 limit" each minute shared between every call of account-details / users / etc.

[Related notion doc](https://www.notion.so/sentry/API-Rate-Limiting-Overview-cbe77d85e2b140d2bb379b59923a008a)

[Reference](https://www.django-rest-framework.org/api-guide/throttling/)

**Hooked up with Gazebo**

Screenshot 2024-08-12 at 11 57 40 AM

### Legal Boilerplate


codecov/settings_staging.py
```diff
@@ -73,3 +73,18 @@
 CSRF_TRUSTED_ORIGINS = [
     get_config("setup", "trusted_origin", default="https://*.codecov.dev")
 ]
+
+REST_FRAMEWORK = {
```
ajay-sentry, line 77: Dont want to add rate limits to staging

rohitvinnakota-codecov (approved):

PR #736, opened 2024-08-05
fix: Add validation for trial start/cancel for user in calling org

ajay-sentry:
### Purpose/Motivation

This PR aims to fix the IDOR hole mentioned in https://github.com/codecov/internal-issues/issues/646 where users can start or cancel trials even if they don't belong to the organization

To fix, we simply check to see that the user is part of the organization in the validate step prior to executing the mutation

### Links to relevant tickets

Closes https://github.com/codecov/internal-issues/issues/646

### Legal Boilerplate

### DIFF
--- a/codecov_auth/commands/owner/interactors/start_trial.py
+++ b/codecov_auth/commands/owner/interactors/start_trial.py
@@ -1,14 +1,17 @@
 from codecov.commands.base import BaseInteractor
-from codecov.commands.exceptions import ValidationError
+from codecov.commands.exceptions import Unauthorized, ValidationError
 from codecov.db import sync_to_async
+from codecov_auth.helpers import current_user_part_of_org
 from codecov_auth.models import Owner
 from plan.service import PlanService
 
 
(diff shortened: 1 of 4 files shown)


codecov_auth/commands/owner/interactors/start_trial.py
```diff
@@ -1,14 +1,17 @@
 from codecov.commands.base import BaseInteractor
-from codecov.commands.exceptions import ValidationError
+from codecov.commands.exceptions import Unauthorized, ValidationError
 from codecov.db import sync_to_async
+from codecov_auth.helpers import current_user_part_of_org
 from codecov_auth.models import Owner
 from plan.service import PlanService
 
 
 class StartTrialInteractor(BaseInteractor):
```
RulaKhaled, line 9: if we use this interactor in graphql fields, we probably can make use of the decorator `@require_part_of_org`

RulaKhaled (approved):

PR #737, opened 2024-08-06
add metrics in upload urls to better match ongoing + changing shelter…

adrian-codecov:
Still trying to reach parity w/ shelter logs, so this is an effort to bring both closer together. Basically adding some logs + adjusting tests for it

### Legal Boilerplate


upload/views/bundle_analysis.py
```diff
@@ -150,9 +154,17 @@ def post(self, request: HttpRequest) -> Response:
                 request=self.request,
                 repository=repo,
                 is_shelter_request=self.is_shelter_request(),
+                position="end",
             ),
         )
 
+        dispatch_upload_task(
```
matt-codecov, line 161: why move this?

matt-codecov (approved):

upload/views/bundle_analysis.py
adrian-codecov: Just wanted the metric to come before the actual dispatching

PR #738, opened 2024-08-06
update gitlab webhook flow

nora-codecov:
### Purpose/Motivation
Refine the flow, use webhook_secret if it is set on repo.

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/359

### DIFF
--- a/webhook_handlers/tests/test_gitlab.py
+++ b/webhook_handlers/tests/test_gitlab.py
@@ -246,59 +246,31 @@ class TestGitlabWebhookHandler(APITestCase):
 
         pulls_sync_mock.assert_called_once_with(repoid=self.repo.repoid, pullid=pullid)
 
-    def test_handle_system_hook_not_enterprise(self):
-        username = "jsmith"
-        project_id = 74
-        owner = OwnerFactory(service="gitlab", username=username)
+    def test_handle_system_hook_when_not_enterprise(self):
+        owner = OwnerFactory(service="gitlab")
+        repo = RepositoryFactory(author=owner)
 
-        response = self._post_event_data(
-            event=GitLabWebhookEvents.SYSTEM,
-            data={
-                "created_at": "2020-01-21T07:30:54Z",
-                "updated_at": "2020-01-21T07:38:22Z",
+        system_hook_events = [
+            "project_create",
+            "project_destroy",
+            "project_rename",
+            "project_transfer",
+            "user_add_to_team",
+            "user_remove_from_team",
+        ]
+
+        event_data = {
+            "event": GitLabWebhookEvents.SYSTEM,
+            "data": {
                 "event_name": "project_create",
-                "name": "StoreCloud",
-                "owner_email": "johnsmith@gmail.com",
-                "owner_name": "John Smith",
-                "path": "storecloud",
-                "path_with_namespace": f"{username}/storecloud",
-                "project_id": project_id,
-                "project_visibility": "private",
+                "project_id": repo.service_id,
             },
-        )
-        assert response.status_code == status.HTTP_403_FORBIDDEN
+        }
 
-        new_repo = Repository.objects.filter(
-            author__ownerid=owner.ownerid, service_id=project_id
-        ).first()
-        assert new_repo is None
-
-    def test_handle_system_hook_project_create(self):
-        username = "jsmith"
-        project_id = 74
-        owner = OwnerFactory(service="gitlab", username=username)
-
-        response = self._post_event_data(
-            event=GitLabWebhookEvents.SYSTEM,
-            data={
-                "created_at": "2020-01-21T07:30:54Z",
-                "updated_at": "2020-01-21T07:38:22Z",
-                "event_name": "project_create",
-                "name": "StoreCloud",
-                "owner_email": "johnsmith@gmail.com",
-                "owner_name": "John Smith",
-                "path": "storecloud",
-                "path_with_namespace": f"{username}/storecloud",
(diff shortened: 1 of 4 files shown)


webhook_handlers/views/gitlab.py
```diff
@@ -170,15 +200,15 @@ def _handle_system_hook_event(self, repo):
...
         elif event_name == "project_rename":
-            new_name = self.request.data.get("path_with_namespace").split("/")[-1]
-            repo.name = new_name
-            repo.save(update_fields=["name"])
-            message = "Repository renamed"
+            owner_username, _ = self.request.data.get("path_with_namespace").split("/")
+            self._try_initiate_sync_for_owner_by_username(owner_username=owner_username)
+            message = "Sync initiated"
 
         elif event_name == "project_transfer":
```
matt-codecov, line 207: we can't express this as a sync for the old owner and a sync for the new owner? i guess that wouldn't transfer historical data

webhook_handlers/tests/test_gitlab.py
```diff
@@ -303,317 +293,19 @@ def test_handle_system_hook_project_create(self):
...
 
-        new_repo = Repository.objects.get(
+        new_repo = Repository.objects.filter(
             author__ownerid=owner.ownerid, service_id=project_id
-        )
-        assert new_repo is not None
-        assert new_repo.private is True
-        assert new_repo.name == "storecloud"
-
-    def test_handle_system_hook_project_destroy(self):
```
matt-codecov, line 316: are these all deleted because we don't intend to support these system hooks without an enterprise license? should we assert that a PermissionDenied() is raised?

webhook_handlers/views/gitlab.py
nora-codecov: !!! I hadn't thought of that! I didn't think we had access to the old owner but i just took a look at the docs and I think I can get this field and can offload to sync! amazing

webhook_handlers/tests/test_gitlab.py
nora-codecov: yes - System hooks are only for IS_ENTERPRISE, and it does a PermissionDenied() is raised test in `test_handle_system_hook_not_enterprise`.
I deleted them because it seemed redundant to test every type of system hook - but your comment made me realize that I should add them back to make it clear that this is the intended behavior.

webhook_handlers/views/gitlab.py
matt-codecov: i think changing it to do two syncs might be noticeably different - existing data from the repo on the old owner won't be available to the new owner. are we okay making that change as part of this webhook fix, or do we need to preserve how it worked before?

webhook_handlers/views/gitlab.py
```diff
@@ -129,112 +155,92 @@ def _handle_merge_request_event(self, repo):
...
-                    service_id=self.request.data.get("project_id"),
-                    name=repo_name,
-                    private=self.request.data.get("project_visibility") == "private",
+                    service=self.service_name,
+                    oauth_token__isnull=False,
+                    name=owner_name,
+                    email=owner_email,
                 )
-                message = "Repository created"
+                self._initiate_sync_for_owner(owner)
```
matt-codecov, line 195: i think this could be one try/catch block?
```
try:
  owner = Owner.objects.get(
    service=self.service_name,
    oauth_token__isnull=False,
    username=owner_username_best_guess,
  )

  if not owner:
    owner_email = self.request.data.get("owner_email")
    owner_name = self.request.data.get("owner_name")
    owner = Owner.objects.get(
      service=self.service_name,
      oauth_token__isnull=False,
      name=owner_name,
      email=owner_email,
    )
except Owner.DoesNotExit:
    message = ":("
    return

self._initiate_sync_for_owner(owner)
```

should we just always use the `self.request` stuff or will it not always be there/correct?

webhook_handlers/views/gitlab.py
```diff
@@ -129,112 +155,92 @@ def _handle_merge_request_event(self, repo):
...
-            obj, created = Owner.objects.update_or_create(
-                service=self.service_name,
-                service_id=self.request.data.get("user_id"),
-                username=self.request.data.get("username"),
-                email=self.request.data.get("email"),
-                name=self.request.data.get("name"),
-            )
-            message = "User created"
+        elif event_name in ("project_rename", "project_transfer"):
+            self._try_initiate_sync_for_owner()
```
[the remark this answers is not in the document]
nora-codecov: so my current hypothesis is this - and i think we were thinking about it wrong before:
A user can only transfer a repo if they have access to the previous location and the new location (both OwnerOrgs).
So doing a sync on the OwnerUser making the transfer request should give us access to both projects (OwnerOrgs), and the sync process must be able to self-correct project/group structure and organization because that's the sync function we use when the user isn't seeing their repos correctly...
So I think this sync (on the OwnerUser) is able to initiate a sync for both OwnerOrgs :)

webhook_handlers/views/gitlab.py
nora-codecov: good call, I'll update 👍

matt-codecov (approved):
looks good!

i think the `Owner` model also has a `bot` column and that might give us another way to make sure we successfully sync: if `owner_email`, `path_with_namespace`, and `old_path_with_namespace` all point to `Owner`s without `oauth_token` set, we can maybe check if they have `bot` set and use that bot

up to you if you want to look into that in this PR or merge the webhook validation stuff and follow up with that idea separately

PR #741, opened 2024-08-08
bundle analysis - add support for tokenless upload

JerrySentry:
closes https://github.com/codecov/engineering-team/issues/2206

Creates a subclass of `TokenlessAuthentication` to validate the BA upload endpoint, for tokenless upload to be successful will need the following in the request body (in addition to what is normally required):
1. `branch` must be present with value of format `forkname:main`
2. `git_service` must be present with value one of: github, gitlab, bitbucket, github_enterprise, gitlab_enterprise, bitbucket_server
3. the repo is a public repo

With this the token head need not be provided.

### Legal Boilerplate


giovanni-guidini (approved):
LGTM

codecov_auth/authentication/repo_auth.py
```diff
@@ -318,3 +318,33 @@ def authenticate(self, request):
...
+        except json.JSONDecodeError:
+            # Validate request body format
+            raise exceptions.AuthenticationFailed(self.auth_failed_message)
+        except ValueError:
+            # Validate provider
+            raise exceptions.AuthenticationFailed(self.auth_failed_message)
+
+    def get_branch(self, request, repoid=None, commitid=None):
+        body = json.loads(str(request.body, "utf8"))
+        return body.get("branch")
```
matt-codecov, line 350: what should you do if the commit already exists and it has a different branch name from what was passed in the request?

the logic for coverage is:
- for `create-commit`, decide if we're allowed to create the commit based on the branch name in the request
- for `create-report` and `do-upload`, decide if we're allowed to perform the action by looking up the commit and checking the branch name it was created with

that way someone can't upload to a commit on `main` and just lie in the request and say the branch name is actually something else to bypass the check.

JerrySentry: Added check to make sure requested commit's branch matches what was stored in our DB

PR #742, opened 2024-08-09
ref: Remodel graphql schemas

suejung-sentry:
This is a draft PR that serves as a way to view the proposed diff for Restructure GraphQL project - https://www.notion.so/sentry/Project-Plan-Restructure-GraphQL-9b47b646b797451090c733be880615e7?pvs=4

Please comment any suggestions!!


graphql_api/types/repository/repository.graphql
```diff
@@ -1,15 +1,28 @@
...
   active: Boolean!
+  # activated fetches the repo's Codecov activation status
   activated: Boolean!
+  # private fetches the repo's public/private status as designated at provider
   private: Boolean!
-  coverage: Float
-  coverageSha: String
-  hits: Int
-  misses: Int
-  lines: Int
```
suejung-sentry, line 11: Shown as deleted in this diff for illustrative purposes. In practice, we plan to delete the old fields only after everything is migrated over to the new fields and stable.

graphql_api/types/0_TEMP/coverage.graphql
```diff
@@ -0,0 +1,51 @@
...
+
+  # formerly repo.measurements
+  coverageMeasurements(
+    interval: MeasurementInterval!
+    after: DateTime
+    before: DateTime
+    branch: String
+  ): [Measurement!]!
+
+  ### FLAGS ###
```
suejung-sentry, line 22: Do we want Flags and Components to have a subtype, something like below? To futureproof it getting added to TestAnalysis or BundleAnalysis as filters in the future
```
type FlagMeasurements {
   flagsCount: Int!
...
}
```

graphql_api/types/0_TEMP/coverage.graphql
```diff
@@ -0,0 +1,51 @@
+# Coverage is information related to a repo's
+# main branch test coverage
+type CoverageAnalysis {
+  hits: Int
+  misses: Int
+  lines: Int
+  coverageSha: String
+  percentCovered: Float # formerly repo.coverage
```
suejung-sentry, line 8: We have the option of using the `CoverageTotals` [type](https://github.com/codecov/codecov-api/blob/324fd728cd905c073e96c6657ac3d363a92c6070/graphql_api/types/coverage_totals/coverage_totals.graphql#L2) instead of these flat fields `hits`, `mises`, etc.

graphql_api/types/repository/repository.graphql
```diff
@@ -25,6 +36,7 @@ type Repository {
     last: Int
     before: String
   ): PullConnection @cost(complexity: 10, multipliers: ["first", "last"])
+  # TODO - does this refer to commits on main only?
```
calvin-codecov, line 39: If anyone can answer this question, please let us know. What happens for other branches?

spalmurray-codecov: It's just on default branch I'm pretty sure. You can get any commit by its SHA with `Repository { commit(id: SHA) {} }` per the API docs. idk how that maps onto this file, however lol.

JerrySentry: `Repository.commits` can be any set of commits depending on the filters used (see: `filters: CommitsSetFilters`)

In the filters it specifies:
```
type CommitsSetFilters {
  hideFailedCI: Boolean
  branchName: String
  pullId: Int
  search: String
  states: [CommitState!]
  coverageStatus: [CommitStatus!]
}
```

graphql_api/types/0_TEMP/coverage.graphql
JerrySentry: The schema under "Another option is" looks so clean! But yeah right now there's some small nuances between different types of measurements (eg coverage, flag, component, bundle analysis) that will require alot of refactoring to make it happen :(

graphql_api/types/repository/repository.graphql
spalmurray-codecov: Dangg that's good to know thanks for clarifying @JerrySentry

graphql_api/types/repository/repository.graphql
JerrySentry: Will you be duplicating the resolvers during the migration process as well? Ideally there would be a way to bind both the new and old schemas to the same resolvers at the same time so that the Python codes don't get duped.

graphql_api/types/repository/repository.graphql
```diff
@@ -1,15 +1,26 @@
+# Repository is a named collection of files uploaded
```
JerrySentry, line 1: btw if you follow this notation:
https://spec.graphql.org/June2018/#sec-Descriptions
these doc strings would appear in the ariadne gql playground tool.

PR #746, opened 2024-08-12
Get app_redirect_url from owner object rather than request params

michelletran-codecov:
### Purpose/Motivation
This is to appease the static analysis security scan to ensure that we are not directly using user input for redirect.

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/645

### What does this PR do?
Change which username and service we're using to generate the app url.

### Legal Boilerplate

### DIFF
--- a/codecov_auth/views/okta_cloud.py
+++ b/codecov_auth/views/okta_cloud.py
@@ -29,7 +29,7 @@ def get_app_redirect_url(org_username: str, service: str) -> str:
     return f"{settings.CODECOV_DASHBOARD_URL}/{service}/{org_username}"
 
 
-def get_oauth_redirect_url(org_username: str, service: str) -> str:
+def get_oauth_redirect_url() -> str:
     """The Okta callback URL for us to finish the authentication."""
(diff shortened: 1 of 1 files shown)


codecov_auth/views/okta_cloud.py
```diff
@@ -72,7 +72,9 @@ def get(
             )
             return HttpResponse(status=404)
 
-        app_redirect_url = get_app_redirect_url(org_username, service)
+        app_redirect_url = get_app_redirect_url(
+            organization.username, organization.service
+        )
         oauth_redirect_url = get_oauth_redirect_url(org_username, service)
```
Swatinem, line 78: does the same also make sense for `oauth_redirect_url`?

Swatinem (approved):
lgtm, but left a comment

codecov_auth/views/okta_cloud.py
michelletran-codecov: Oh, actually the `oauth_redirect_url` is not using either fields, so I can probably just remove them. Thanks for the catch!

PR #748, opened 2024-08-13
update has_seats_left on plan service to work properly with self-hosted and DEC

nora-codecov:
### Purpose/Motivation
We calculate seats differently for self-hosted and DEC, this updates the `has_seats_left` on `PlanService` to work correctly for self-hosted and DEC.

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/679

### What does this PR do?
The function was comparing the total seat count to the number of active users for the Org. With self-hosted and DEC, it will now compare total seat count to the number of unique users across all Orgs.

### DIFF
--- a/services/self_hosted.py
+++ b/services/self_hosted.py
@@ -7,6 +7,7 @@ from django.db import transaction
 from django.db.models import F, Func, Q, QuerySet
 from shared.license import get_current_license
 
+from codecov.db import sync_to_async
 from codecov_auth.models import Owner
 from services import ServiceException
 from utils.config import get_config
@@ -44,16 +45,16 @@ def admin_owners() -> QuerySet:
 
 def is_admin_owner(owner: Optional[Owner]) -> bool:
     """
-    Returns true iff the given owner is an admin.
+    Returns true if the given owner is an admin.
     """
     return owner is not None and admin_owners().filter(pk=owner.pk).exists()
 
 
-def activated_owners() -> QuerySet:
-    """
-    Returns all owners that are activated in ANY org's `plan_activated_users`
-    across the entire instance.
-    """
+async def activated_owners_async() -> QuerySet:
+    return await sync_to_async(activated_owner_query)()
+
+
+def activated_owner_query() -> QuerySet:
     owner_ids = (
         Owner.objects.annotate(
             plan_activated_owner_ids=Func(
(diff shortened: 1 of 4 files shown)


services/self_hosted.py
```diff
@@ -83,9 +91,15 @@ def license_seats() -> int:
...
+async def enterprise_has_seats_left() -> bool:
+    owners = await activated_owners_async()
+    count = await sync_to_async(owners.count)()
+    return count < license_seats()
+
+
 def can_activate_owner(owner: Owner) -> bool:
     """
-    Returns true iff there are available seats left for activation.
+    Returns true if there are available seats left for activation.
```
ajay-sentry, line 102: weird how all of these had 2 "f"s lol
Maybe they meant if and only if?

services/self_hosted.py
```diff
@@ -44,16 +45,16 @@ def admin_owners() -> QuerySet:
...
-def activated_owners() -> QuerySet:
-    """
-    Returns all owners that are activated in ANY org's `plan_activated_users`
-    across the entire instance.
-    """
+async def activated_owners_async() -> QuerySet:
+    return await sync_to_async(activated_owner_query)()
+
+
+def activated_owner_query() -> QuerySet:
```
ajay-sentry, line 57: should we be worried about perf at all with this fn? genuinely curious

ajay-sentry (approved):
logic looks good, a couple questions

services/self_hosted.py
```diff
@@ -83,9 +91,15 @@ def license_seats() -> int:
     return license.number_allowed_users or 0
 
 
+async def enterprise_has_seats_left() -> bool:
+    owners = await activated_owners_async()
```
ajay-sentry, line 95: Are able to do the owners.count() here directly on the query set?

services/self_hosted.py
nora-codecov: yes 🙃 
I gated it behind a license check - so it will only run if the license is valid

services/self_hosted.py
nora-codecov: yeah is that a shorthand I don't know about either?

nora-codecov: [microsoft says](https://github.com/microsoft/terminal/issues/4010) it's if and only if but `iff` only works with 2 statements right? so i think typo

services/self_hosted.py
nora-codecov: it throws error if I don't make the `count()` sync_to_async as well, the tradeoff is that I get to use the same activated_owner_query that is used elsewhere, but i have to do this additional async step

PR #750, opened 2024-08-13
feat: Add rate limiter to GQL endpoint

ajay-sentry:
### Purpose/Motivation

This PR aims to add a rate limiter to our graphQL endpoint, with two different "key" mechanisms depending on if the user is logged in or not.

Our first check will check for the user being logged in, and having a "pk" on their user object passed into the request. If they _do_ have this value, we will use it as the basis for their rate limit. If they don't have this value (i.e. their just a guest or not logged in), we will fall back to trying to derive their IP address via some request headers being passed in. From my investigation on which headers are typically defined with an IP address, it seemed to be either via x_forwarded_at or remote_addr. We'll use the former but fallback to the latter if it doesn't exist. In the event none of those exist we're a little S.O.L, but we can monitor on the initial launch and test on stage prior to see if we can "break it" before going to prod

Outside of that stuff though, the rate limiter implementation is pretty straight forward. You get a flat 300 req/min per "key," and we can easily modify and extend this implementation to include daily rate limits too if we want.

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/2148

### Testing

- Tested on stage, the most amount of requests I was able to come up with seemed to be clicking into a repo -> clicking back to repo list -> clicking into a repo -> back to repo list -> etc until I hit the rate limit
- I had to do this pretty quick to hit the rate limit even at 300 requests per min so I'm somewhat confident this is a safe upperbound to start with
- After hitting it while logged in, I tried navigating to a separate page after logging out and was able to, meaning I wasn't hitting the IP level check either ✅
- After doing the same process I successfully was able to rate limit myself both via IP and via userId lol 

### Screenshots

Screenshot 2024-08-27 at 4 55 42 PM

### Legal Boilerplate


graphql_api/views.py
```diff
@@ -290,6 +305,40 @@ def _get_user(self, request):
...
+        try:
+            # eagerly try to get user_id from request object
+            user_id = request.user.pk
+        except AttributeError:
+            user_id = None
+
+        if user_id:
+            key = f"rl-user:{user_id}"
+        else:
+            key = f"rl-ip:{user_ip}"
```
rohitvinnakota-codecov, line 320: Can we just fetch the ip here? Seems like we don't use it unless the user id does not exist

ajay-sentry: Good point, will make the change

rohitvinnakota-codecov (approved):

PR #752, opened 2024-08-15
remove session_totals field from report json

matt-codecov:
https://github.com/codecov/internal-issues/issues/515
https://github.com/codecov/shared/pull/329

this field is bugged and, as far as i can tell, unused. see ticket + shared PR for details


Swatinem (approved):
always happy to see code go.
I guess the only real change here is in `services/report.py` which seems trivial?

PR #756, opened 2024-08-16
fix: replace special char in test name with space

joseph-sentry:
We collate the class name and test case name parsed from the JUnit XMLs with a special char `\x1f`, we want to display it as a space though. Perhaps this logic should occur in Gazebo?


ajay-sentry (approved):
send it

rohan-at-sentry:
@joseph-sentry for monday? ^

PR #762, opened 2024-08-20
Handle owner user potentially null in base interactor 

RulaKhaled:
### Purpose/Motivation
If owner.user is null, we can't treat user as an anonymous user anymore. This is fixing to reset to anonymous user if user does not exist. 

### Legal Boilerplate


michelletran-codecov (approved):

RulaKhaled:
@codecov-ai-reviewer test

RulaKhaled:
@codecov-ai-reviewer test

PR #763, opened 2024-08-21
Strip Okta ISS URL of trailing slashes

michelletran-codecov:
### Purpose/Motivation
This removes trailing slashes from their usage in the code, and also from saving to the database. This should fix issues where we validate the payload from Okta with the ISS URL (they send it without trailing slashes).

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/709

### What does this PR do?
1. Strip the URL of trailing slashes at the point of using the URL
2. Strip the URL of trailing slashes before saving to the DB

We _probably_ don't need both, but having both will just be safer.

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/codecov_auth/commands/owner/interactors/save_okta_config.py
+++ b/codecov_auth/commands/owner/interactors/save_okta_config.py
@@ -20,7 +20,7 @@ class SaveOktaConfigInput:
 
 
 class SaveOktaConfigInteractor(BaseInteractor):
-    def validate(self, owner: Owner) -> bool:
+    def validate(self, owner: Owner) -> None:
         if not self.current_user.is_authenticated:
             raise Unauthenticated()
         if not owner:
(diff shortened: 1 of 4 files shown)


codecov_auth/commands/owner/interactors/save_okta_config.py
```diff
@@ -16,11 +16,11 @@ class SaveOktaConfigInput:
     url: Optional[str] = None
     enabled: bool = False
     enforced: bool = False
-    org_username: str = None
+    org_username: str | None = None
 
 
 class SaveOktaConfigInteractor(BaseInteractor):
-    def validate(self, owner: Owner):
+    def validate(self, owner: Owner) -> bool:
```
Swatinem, line 23:
```suggestion
    def validate(self, owner: Owner) -> None:
```
there is no `return` statement

michelletran-codecov: Whoops! Thanks for the catch!

Swatinem (approved):

PR #765, opened 2024-08-23
REST API: add flag coverage to flag list endpoint

JerrySentry:
closes https://github.com/codecov/feedback/issues/362

Updates `/api/v2/{service}/{owner_username}/repos/{repo_name}/flags/` endpoint to include the coverage value for each flag in the response object.

Old: `"results": [{"flag_name": string}]`
New: `"results": [{"flag_name": string, "coverage": float}]`

This coverage value comes from the computed report of the head commit of the repo's default branch. If failure to retrieve such value a default value of 0 will be returned.

### Legal Boilerplate


api/public/v2/flag/views.py
```diff
@@ -15,11 +21,24 @@ class FlagViewSet(viewsets.GenericViewSet, mixins.ListModelMixin, RepoPropertyMi
...
 
-    def get_queryset(self):
-        return self.repo.flags.all()
+    def get_queryset(self) -> QuerySet:
+        results = [
+            {"flag_name": f.flag_name, "coverage": 0} for f in self.repo.flags.all()
+        ]
+        try:
+            report = self.get_commit().full_report
+            if not report:
```
ajay-sentry, line 30: just curious, when would we throw a NotFound vs. return None?

api/public/v2/flag/views.py
```diff
@@ -15,11 +21,24 @@ class FlagViewSet(viewsets.GenericViewSet, mixins.ListModelMixin, RepoPropertyMi
...
+                return results
+        except NotFound:
+            return results
+
+        for i, val in enumerate(results):
+            flag_report = report.filter(flags=[val["flag_name"]])
+            results[i]["coverage"] = flag_report.totals.coverage or 0
```
ajay-sentry, line 37: totals will always exist?

api/public/v2/tests/test_flag_viewset.py
```diff
@@ -38,15 +66,64 @@ def _request_flags(self):
...
+        get_repo_permissions.return_value = (True, True)
+
+        res = self._request_flags()
+        assert res.status_code == 200
+        assert res.json() == {
+            "count": 2,
+            "next": None,
+            "previous": None,
+            "results": [
+                {"flag_name": "foo", "coverage": 0},
```
ajay-sentry, line 79: Jw on returning 0 for coverage here vs. None 🤔

api/public/v2/flag/views.py
JerrySentry: this line would return the list of flags without the coverage information for each flag if the coverage report of the head commit is not yet processed.
in general imo we should throw 404 if the requested data is not found (eg in the param you have a repo/owner that doesn't exist) so there's nothing to return. And return null, I don't think we would return that as a response to the entire request, rather the response object has some partial data is not available.

api/public/v2/flag/views.py
JerrySentry: from what I can tell yes

api/public/v2/tests/test_flag_viewset.py
JerrySentry: yeah I feel none is better

api/public/v2/flag/views.py
ajay-sentry: Gotcha! Makes sense to me, thanks for the context

api/public/v2/flag/views.py
ajay-sentry: sggg

ajay-sentry (approved):
looks good 👌

PR #778, opened 2024-08-28
Add an admin action to remove stale `plan_activated_users`

Swatinem:
### Purpose/Motivation

Enterprise orgs often want to clean up their stale users to free up seats. This implements such an "admin" action.

### Links to relevant tickets

- fixes https://github.com/codecov/engineering-team/issues/2101
- fixes https://github.com/codecov/engineering-team/issues/2296

### What does this PR do?

This adds a new function which finds all the "stale" `plan_activated_users` for a set of given `orgs`. Users are considered stale if they don’t have any recent auth session or pull request. Those users will then be removed from the `plan_activated_users`.

### DIFF
--- a/codecov_auth/admin.py
+++ b/codecov_auth/admin.py
@@ -1,14 +1,17 @@
 import logging
-from typing import Optional
+from datetime import timedelta
+from typing import Optional, Sequence
 
 import django.forms as forms
 from django.conf import settings
 from django.contrib import admin, messages
 from django.contrib.admin.models import LogEntry
+from django.db.models import OuterRef, Subquery
 from django.db.models.fields import BLANK_CHOICE_DASH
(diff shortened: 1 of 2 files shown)


nora-codecov (approved):
great use of set math!

one note - I saw you added this on Account admin, not very many Orgs have Accounts, and I think potentially none of the enterprise orgs (that you mention in the description) would have Accounts. To get the effect you want, you would need to add the action to Owner admin.

Swatinem:
> I saw you added this on Account admin, not very many Orgs have Accounts, and I think potentially none of the enterprise orgs (that you mention in the description) would have Accounts. To get the effect you want, you would need to add the action to Owner admin.

Thanks for this explanation. I will circle back on this and will implement that as a followup when needed.

PR #783, opened 2024-08-29
query errors during async process

nora-codecov:
### Purpose/Motivation
In django, adding a dot notation to an object does a db query, so when I added `self.current_org.account` to `PlanService`, this adds a db call. This breaks an async flow, which gazebo uses. Solution is to do a prefetch_related when we get the `current_org` so that `self.current_org.account` does not call the db.

### Links to relevant tickets
https://codecov.sentry.io/issues/5771897965/events/a320d20a70534672bac6ce9db263f4e0/

### DIFF
--- a/graphql_api/tests/test_plan.py
+++ b/graphql_api/tests/test_plan.py
@@ -5,6 +5,7 @@ import pytest
 from django.test import TransactionTestCase
 from django.utils import timezone
 from freezegun import freeze_time
+from shared.django_apps.codecov_auth.tests.factories import AccountFactory
 from shared.license import LicenseInformation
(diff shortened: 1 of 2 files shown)


graphql_api/tests/test_plan.py
```diff
@@ -83,6 +84,38 @@ def test_owner_plan_data_when_trialing(self):
             "planUserCount": 123,
         }
 
+    @freeze_time("2023-06-19")
```
ajay-sentry, line 87: nit: do we need the freeze time?

ajay-sentry (approved):
lgtm!

PR #789, opened 2024-09-03
fix: actually use history in aggregate_test_results

joseph-sentry:


utils/test_results.py
```diff
@@ -48,9 +48,7 @@ def aggregate_test_results(
 
     """
     time_ago = (
-        dt.datetime.now(dt.UTC) - dt.timedelta(days=30)
-        if history is not None
-        else thirty_days_ago
+        (dt.datetime.now(dt.UTC) - history) if history is not None else thirty_days_ago
```
adrian-codecov, line 51: What values can history range from?

joseph-sentry: i guess it could technically be anything if the caller to this function decides

adrian-codecov: If this is going to leverage the partitions, I'd perhaps leave a comment noting the partition size and to be aware of preferred ranges 👌

adrian-codecov (approved):

PR #798, opened 2024-09-05
bundle analysis: add a paginated option to retrieve assets

JerrySentry:
Add a `assetsPaginated` field to `BundleReport` that allows pagination of asset retrieval. This is done primarily to speed up asset trend retrieval.

New GQL schema:
```
type BundleReport {
  ...
  assetsPaginated(
    ordering: AssetOrdering
    orderingDirection: OrderingDirection
    first: Int
    after: String
    last: Int
    before: String
  ): AssetConnection
}

type AssetConnection {
  edges: [AssetEdge]!
  totalCount: Int!
  pageInfo: PageInfo!
}

type AssetEdge {
  cursor: String!
  node: BundleAsset!
}

enum AssetOrdering {
  NAME
  SIZE
  TYPE
}
```

### Legal Boilerplate


graphql_api/types/bundle_analysis/base.py
```diff
@@ -135,6 +145,65 @@ def resolve_assets(
...
+    info: GraphQLResolveInfo,
+    ordering: AssetOrdering = AssetOrdering.SIZE,
+    ordering_direction: OrderingDirection = OrderingDirection.DESC,
+    first: Optional[int] = None,
+    after: Optional[str] = None,
+    last: Optional[int] = None,
+    before: Optional[str] = None,
+) -> Dict[str, object]:
+    # All filtered assets before pagination
+    assets = list(
```
JerrySentry, line 160: A non-ideal way to perform pagination. It is still fetching all the assets (minus the ones filtered out) from the report and then return the requested blocks. Since fetching all assets is trivial for a given report we will live with this non optimized way. The major reason we want to paginate assets is because each asset object contains timeseries measurement data which is what's causing 99% of the load time. Paginating the assets mean we can lazy load the timeseries data on the UI. Note that fetching all assets in the first step doesn't begin the computation of the measurements data which means its fine.

graphql_api/types/bundle_analysis/base.py
```diff
@@ -135,6 +145,65 @@ def resolve_assets(
     return list(bundle_report.assets())
 
 
+@bundle_report_bindable.field("assetsPaginated")
+def resolve_assets_paginated(
+    bundle_report: BundleReport,
+    info: GraphQLResolveInfo,
+    ordering: AssetOrdering = AssetOrdering.SIZE,
```
ajay-sentry, line 152: should this and ordering_direction be optional as well since we're defaulting a value?

graphql_api/types/bundle_analysis/base.py
```diff
@@ -135,6 +145,65 @@ def resolve_assets(
...
+            ordering=ordering.value,
+            ordering_desc=ordering_direction.value == OrderingDirection.DESC.value,
+        )
+    )
+
+    total_count, has_next_page, has_previous_page = len(assets), False, False
+    start_cursor, end_cursor = None, None
+
+    # Apply cursors to edges
+    if after is not None:
```
ajay-sentry, line 171: in the event after and before are both not None, this code will default to before; but I'm wondering if we should throw an error maybe?

graphql_api/types/bundle_analysis/base.py
```diff
@@ -135,6 +145,65 @@ def resolve_assets(
...
+        if after_edge > -1:
+            assets = assets[after_edge + 1 :]
+
+    if before is not None:
+        before_edge = _find_index_by_cursor(assets, before)
+        if before_edge > -1:
+            assets = assets[:before_edge]
+
+    # Slice edges by return size
+    if first is not None and first >= 0:
```
ajay-sentry, line 182: Similar story with first and last

graphql_api/types/bundle_analysis/base.py
```diff
@@ -135,6 +145,65 @@ def resolve_assets(
...
+        if len(assets) > first:
```
ajay-sentry, line 183: nit: we can use total_count var instead of len(assets) for all these references to len(assets) right?

graphql_api/types/bundle_analysis/base.py
JerrySentry: yes, good catch

graphql_api/types/bundle_analysis/base.py
JerrySentry: Just implementing off of the official spec for this style of pagination [here](https://relay.dev/graphql/connections.htm) it says:
>Including a value for both first and last is strongly discouraged, as it is likely to lead to confusing queries and results. The PageInfo section goes into more detail here.
With the current setup it does indeed become messed up when both are used. I agree with changing to throw an error if both are used at the same time.

graphql_api/types/bundle_analysis/base.py
JerrySentry: total_count might not equal to len(assets) at this point of the code path. assets could have gotten sliced by the after/before if statements. total_count is for the grand total of results after filtreing.

graphql_api/types/bundle_analysis/base.py
JerrySentry: wait I think it is currently ok, right now it should mean that `ordering` is of type `AssetOrdering`, but if that param is not passed in to the call it would be defaulted to `AssetOrdering.SIZE`. It should never be the case that `ordering` is null.

graphql_api/types/bundle_analysis/base.py
JerrySentry: Done.

graphql_api/types/bundle_analysis/base.py
JerrySentry: Done.

graphql_api/types/bundle_analysis/base.py
ajay-sentry: Fair enough, I can appreciate the defensive nature of having a fallback value here

graphql_api/types/bundle_analysis/base.py
ajay-sentry: ahhh okay yeah I totally missed that previously, makes sense

ajay-sentry (approved):
looks good 👌

PR #804, opened 2024-09-06
fix: only call commit_update task when commit is newly created

joseph-sentry:
we have to move the get_or_create call in the commits endpoint to the top level function so we can know if a commit object was newly created

we don't want to call the commit update task every time this endpoint is called because it may be the cause of lock contention in the DB for commits


giovanni-guidini (commented):
Is [this function](https://github.com/codecov/codecov-api/blob/ab61f47dd59fed8c6085aa82994a1123031b3bd8/upload/serializers.py#L137) in the serializer not used anymore?

If that's the case please remove that too.

giovanni-guidini (approved):

PR #811, opened 2024-09-10
chore: reduce usage of PreProcess task

giovanni-guidini:
Originally I wanted to delete PreProcess task completely, but after reviewing
certain usage patterns I realized the task is necessary (for `empty-upload` and
`static-analysis` for example).

So the alternative is to reduce the amount of times it is called, to only be called
when we effectively create a new report. Similar to https://github.com/codecov/codecov-api/pull/804

### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/rollouts/__init__.py
+++ b/rollouts/__init__.py
@@ -9,8 +9,5 @@ def owner_slug(owner: Owner) -> str:
 
 __all__ = ["Feature"]
 
-NO_PREPROCESS_UPLOAD = Feature("no_PreProcessUpload")
-
-
 # By default, features have one variant:
 #    { "enabled": FeatureVariant(True, 1.0) }
--- a/upload/serializers.py
+++ b/upload/serializers.py
@@ -161,7 +161,7 @@ class CommitReportSerializer(serializers.ModelSerializer):
         )
         fields = read_only_fields + ("code",)
 
-    def create(self, validated_data):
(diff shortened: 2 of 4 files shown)


Swatinem (approved):

Swatinem:
One question though:
As these tasks are asynchronous, can we end up in a situation where the `Upload` task is being run before the `PreProcessUpload` task finishes?

Also, why is this an async task in the first place, and why not synchronously do whatever needs to be done within the API request?

giovanni-guidini:
> One question though: As these tasks are asynchronous, can we end up in a situation where the `Upload` task is being run before the `PreProcessUpload` task finishes?
> 
> Also, why is this an async task in the first place, and why not synchronously do whatever needs to be done within the API request?

**For question 1**:
yes. But the `PreProcess` and `Upload` tasks are redundant in terms of "carryforwarding reports" (both do it, but the `Upload` task does other things too).

However there are scenarios - `empty_upload` being an example, apparently - in which there is no `Upload` task.

For question 2: I think 2 reasons:
(1) code instrumentation. The `worker` has the code that is used (currently) to handle report processing, and carryforwarding reports is part of that. Just for this reason we naturally turned to making it an async task

(2) the operation takes too long. It can take up to several seconds to actually do the carryforwarding of reports. That's a bit too much (I think) do do in the request cycle

PR #812, opened 2024-09-10
bundle analysis: fix slow trend query

JerrySentry:
use the owner_id index when querying the timeseries_measurement table
### Legal Boilerplate

### DIFF
--- a/graphql_api/actions/measurements.py
+++ b/graphql_api/actions/measurements.py
@@ -49,15 +49,17 @@ def measurements_by_ids(
 
 
 def measurements_last_uploaded_before_start_date(
+    owner_id: int,
     repo_id: int,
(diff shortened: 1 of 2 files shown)


graphql_api/actions/measurements.py
```diff
@@ -49,15 +49,17 @@ def measurements_by_ids(
 
 
 def measurements_last_uploaded_before_start_date(
+    owner_id: int,
     repo_id: int,
     measurable_name: str,
-    measurable_ids: List[int],
+    measurable_id: int,
```
JerrySentry, line 55: only ever send in 1 value here, no point in have a list

suejung-sentry (approved):

PR #818, opened 2024-09-12
Log all type of "owner" arg for fetch_commit_yaml

JerrySentry:
The call from this function is coming from a lot of places, while it's only expecting a `Owner` type there has been other types being passed in.

### Legal Boilerplate

### DIFF
--- a/services/yaml.py
+++ b/services/yaml.py
@@ -8,7 +8,7 @@ from shared.yaml import UserYaml, fetch_current_yaml_from_provider_via_reference
 from shared.yaml.validation import validate_yaml
 from yaml import safe_load
 
-from codecov_auth.models import Owner, get_config
+from codecov_auth.models import Owner, User, get_config
 from core.models import Commit
 from services.repo_providers import RepoProviderService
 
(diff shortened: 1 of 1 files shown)


nora-codecov (approved):

services/yaml.py
```diff
@@ -39,11 +39,21 @@ def fetch_commit_yaml(commit: Commit, owner: Owner | None) -> Dict | None:
         # have various exceptions, which we do not care about to get the final
         # yaml used for a commit, as any error here, the codecov.yaml would not
         # be used, so we return None here
+
+        if isinstance(owner, Owner):
+            owner_arg = owner.ownerid
```
nora-codecov, line 44: does `owner.pk` not work here?

services/yaml.py
```diff
@@ -39,11 +39,21 @@ def fetch_commit_yaml(commit: Commit, owner: Owner | None) -> Dict | None:
...
+            owner_arg = str(type(owner))
+        else:
+            owner_arg = "None"
+
         log.warning(
             f"Was not able to fetch yaml file for commit. Ignoring error and returning None. Exception: {e}",
             extra={
                 "commit_id": commit.commitid,
-                "owner": owner.ownerid if owner else None,
+                "owner_arg": owner_arg,
```
nora-codecov, line 56: while we're logging things, do you want to log the object type? so when you have the id you know which table to `.get()` from?

nora-codecov (commented):
could you add tests?

JerrySentry:
just gonna log the type of the arg, don't really need to know any details of what the owner is.

PR #823, opened 2024-09-13
Make sure calls to fetch_commit_yaml are Owner type

JerrySentry:
- In empty-upload depending on the auth classes used, the `request.user` can be `Owner` or `RepositoryAsUser`, need to make sure when sending to `final_commit_yaml` it will be `Owner`
- Several places in public API, it passes in to `fetch_commit_yaml ` a `User` object of the authenticated request maker, which is wrong. Instead it should be the `Owner` of the requested organization
- Add an assertion to `fetch_commit_yaml` to make sure if an `owner` arg is passed in it is in fact `Owner`, previously this error is not surfaced because we just do a catch all exception.

### Legal Boilerplate

### DIFF
--- a/services/yaml.py
+++ b/services/yaml.py
@@ -30,9 +30,10 @@ def fetch_commit_yaml(commit: Commit, owner: Owner | None) -> Dict | None:
     # Previously this issue is masked by the catch all exception handling. Most badly typed calls have
     # been addressed, but there might still be some entrypoints unaccounted for, will fix new discoveries
     # from this assertion being raised.
-    assert owner is None or isinstance(
-        owner, Owner
-    ), f"fetch_commit_yaml owner arg must be Owner or None. Provided: {type(owner)}"
+    if owner is not None and not isinstance(owner, Owner):
+        raise TypeError(
+            f"fetch_commit_yaml owner arg must be Owner or None. Provided: {type(owner)}"
+        )
 
     try:
         repository_service = RepoProviderService().get_adapter(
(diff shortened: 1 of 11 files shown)


services/yaml.py
```diff
@@ -25,6 +25,15 @@ def fetch_commit_yaml(commit: Commit, owner: Owner | None) -> Dict | None:
...
     Service provider API request is made on behalf of the given `owner`.
     """
+
+    # There's been a lot of instances where this function is called where the owner arg is not Owner type
+    # Previously this issue is masked by the catch all exception handling. Most badly typed calls have
+    # been addressed, but there might still be some entrypoints unaccounted for, will fix new discoveries
+    # from this assertion being raised.
+    assert owner is None or isinstance(
+        owner, Owner
+    ), f"fetch_commit_yaml owner arg must be Owner or None. Provided: {type(owner)}"
```
michelletran-codecov, line 35: Although we have a lot of `asserts` in the code, using this for checks is generally not good practice (i.e. can be turned off in production for performance). Can we explicitly raise a `TypeError` here instead?

upload/views/empty_upload.py
```diff
@@ -112,7 +113,12 @@ def post(self, request, *args, **kwargs):
                 status=status.HTTP_200_OK,
             )
 
-        yaml = final_commit_yaml(commit, request.user).to_dict()
+        # Depending on the authen class used, the request.user may need to be converted to a Owner
```
michelletran-codecov, line 116:
```suggestion
        # Depending on the authentication class used, the request.user may need to be converted to a Owner
```

services/yaml.py
JerrySentry: makes sense to me. done.

michelletran-codecov (approved):
LGTM modulo comment nit.

PR #824, opened 2024-09-13
[GQL] Add filters to the testResults field on the Repository

joseph-sentry:
we want to be able to filter by flaky tests, failed tests and the
slowest tests, because we want the clickable headers in the UI


graphql_api/types/repository/repository.py
```diff
@@ -579,8 +584,22 @@ async def resolve_test_results(
     filters=None,
     **kwargs,
 ):
-    queryset = await sync_to_async(aggregate_test_results)(
-        repoid=repository.repoid, branch=filters.get("branch") if filters else None
+    parameter = None
```
ajay-sentry, line 587: do we need to handle a default case here?

graphql_api/types/test_results_aggregates/test_results_aggregates.py
```diff
@@ -0,0 +1,23 @@
+from ariadne import ObjectType
+
+test_results_aggregates_bindable = ObjectType("TestResultsAggregates")
+
+
+@test_results_aggregates_bindable.field("totalRunTime")
+def resolve_name(obj, _) -> float:
```
ajay-sentry, line 7: nit: I think our naming convention for these functions is resolve_bindable_field_name

graphql_api/types/flake_aggregates/flake_aggregates.py
```diff
@@ -0,0 +1,13 @@
+from ariadne import ObjectType
+
+flake_aggregates_bindable = ObjectType("FlakeAggregates")
+
+
+@flake_aggregates_bindable.field("flakeCount")
+def resolve_name(obj, _) -> int:
```
ajay-sentry, line 7: reminder to update these function names to the actual resolve value

graphql_api/types/enums/test_results_filtering_parameter.graphql
```diff
@@ -0,0 +1,5 @@
+enum TestResultsFilterParameter {
+  FLAKY_TESTS,
```
ajay-sentry, line 2: do we need one of these for Skips too?

graphql_api/types/repository/repository.py
```diff
@@ -593,3 +612,24 @@ async def resolve_test_results(
         else OrderingDirection.DESC,
         **kwargs,
     )
+
+
+@repository_bindable.field("testResultsHeaders")
```
ajay-sentry, line 617: thought we had agreed to keep the name generic vs. "headers" in case design changes or we want to use aspects of this object elsewhere

graphql_api/types/repository/repository.py
joseph-sentry: the default case is a noop so i don't think so

utils/test_results.py
```diff
@@ -19,14 +19,21 @@
...
     Flake,
-    TestInstance,
 )
 
 thirty_days_ago = dt.date.today() - dt.timedelta(days=30)
 
 SLOW_TEST_PERCENTILE = 95
 
 
+def slow_test_threshold(num_tests: int):
```
ajay-sentry, line 29: this can be simplified to 

percentile = (100-95)/100

slow_tests_to_return = floor(percentile * total_tests)
return max(slow_tests_to_return,1)

ajay-sentry (approved):
lgtm

PR #827, opened 2024-09-16
[temp] Add additional permissions to pulls and test results endpoint

rohitvinnakota-codecov:
### Purpose/Motivation

This change unblocks some short term workflows between Codecov and Sentry's Seer service. Seer will query Codecov's pulls endpoint for a given PR and use coverage information to generate tests. 

### Legal Boilerplate

### DIFF
--- a/api/public/v2/tests/test_api_pull_viewset.py
+++ b/api/public/v2/tests/test_api_pull_viewset.py
@@ -3,6 +3,7 @@ from unittest.mock import patch
 from django.urls import reverse
 from freezegun import freeze_time
 
+from django.test import override_settings
 from codecov.tests.base_test import InternalAPITest
 from codecov_auth.tests.factories import OwnerFactory
 from core.models import Pull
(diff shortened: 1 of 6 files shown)


adrian-codecov (commented):
These look good, although I encourage you to also add tests to ensure this works accordingly - check the test_report_viewset.py file as guidance. During local development, use print() to ensure you're going through the authentication class you're expecting

api/public/v2/tests/test_api_pull_viewset.py
```diff
@@ -201,3 +202,104 @@ def test_retrieve(self, get_repo_permissions):
...
+            reverse(
+                "api-v2-pulls-detail",
+                kwargs={
+                    "service": self.org.service,
+                    "owner_username": self.org.username,
+                    "repo_name": self.repo.name,
+                    "pullid": self.pulls[0].pullid,
+                },
+            ),
+            HTTP_AUTHORIZATION="Bearer testaxs3o76rdcdpfzexuccx3uatui2nw73r",
```
adrian-codecov, line 270: Optional: since this test succeeds because the token must match the SUPER_API_TOKEN, perhaps make a constant MOCK_SUPER_API_TOKEN up top and use it where needed

api/public/v2/tests/test_api_pull_viewset.py
```diff
@@ -201,3 +202,104 @@ def test_retrieve(self, get_repo_permissions):
...
+        assert res.status_code == 401
+        assert res.data["detail"] == "Invalid token."
+
+    @override_settings(SUPER_API_TOKEN="testaxs3o76rdcdpfzexuccx3uatui2nw73r")
+    @patch("api.shared.permissions.RepositoryArtifactPermissions.has_permission")
+    def test_no_pull_if_super_token_but_no_GET_request(
+        self, repository_artifact_permisssions_has_permission
+    ):
+        repository_artifact_permisssions_has_permission.return_value = False
+        res = self.client.post(
```
adrian-codecov, line 260: I also didn't explicitly ask, but you're expecting only GETs to these endpoints right?

api/public/v2/tests/test_api_pull_viewset.py
```diff
@@ -201,3 +202,104 @@ def test_retrieve(self, get_repo_permissions):
...
+        assert (
+            res.data["detail"] == "You do not have permission to perform this action."
+        )
+
+    @override_settings(SUPER_API_TOKEN="testaxs3o76rdcdpfzexuccx3uatui2nw73r")
+    @patch("api.shared.permissions.RepositoryArtifactPermissions.has_permission")
+    def test_pull_with_valid_super_token(
+        self, repository_artifact_permisssions_has_permission
+    ):
+        repository_artifact_permisssions_has_permission.return_value = False
```
adrian-codecov, line 282: I'm actually curious, I don't think we need this mock here. Could you try getting rid of this and see if the test still passes?

adrian-codecov (approved):
Left some suggestions, those applicable please do to the other test as well, but after that you should be gtg 👏

PR #831, opened 2024-09-19
Fix stripe subscription upgrade failures

nora-codecov:
### Purpose/Motivation
We were eagerly updating the user's `plan` in the specific case where they already have a `subscription` and they are upgrading. We were updating without checking that the payment was successful, so in the case where we have updated the `plan` but the stripe payment fails, we were bumping the user all the way back to free tier.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2407
stripe's docs: https://docs.stripe.com/billing/subscriptions/pending-updates

### What does this PR do?
Only update the plan if the payment was successful. If it's not successful, reject the changes and give they user the delinquent pop-up. The user in our db and in Stripe's system stays on the original plan until they payment is successful.

### DIFF
--- a/billing/views.py
+++ b/billing/views.py
@@ -222,9 +222,20 @@ class StripeWebhookHandler(APIView):
             stripe_customer_id=subscription.customer,
         )
 
+        indication_of_payment_failure = getattr(subscription, "pending_update", None)
+        if indication_of_payment_failure:
+            # payment failed, raise this to user by setting as delinquent
+            owner.delinquent = True
+            owner.save()
+            log.info(
+                f"Stripe subscription upgrade failed for owner {owner.ownerid}",
+                extra=dict(pending_update=indication_of_payment_failure),
+            )
+            return
+
         # Properly attach the payment method on the customer
-        # This hook will be called after a checkout session completes, updating the subscription created
-        # with it
+        # This hook will be called after a checkout session completes,
+        # updating the subscription created with it
         default_payment_method = subscription.default_payment_method
         if default_payment_method and owner.stripe_customer_id is not None:
             stripe.PaymentMethod.attach(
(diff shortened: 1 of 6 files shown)


billing/views.py
```diff
@@ -235,11 +246,9 @@ def customer_subscription_updated(self, subscription: stripe.Subscription) -> No
                 invoice_settings={"default_payment_method": default_payment_method},
             )
 
-        subscription_schedule_id = subscription.schedule
-        plan_service = PlanService(current_org=owner)
-
         # Only update if there isn't a scheduled subscription
-        if not subscription_schedule_id:
+        if not subscription.schedule:
```
ajay-sentry, line 250: love this ngl

services/billing.py
```diff
@@ -237,16 +237,31 @@ def modify_subscription(self, owner, desired_plan):
...
+            if indication_of_payment_failure:
+                # payment failed, raise this to user by setting as delinquent
+                owner.delinquent = True
+                owner.save()
+                log.info(
+                    f"Stripe subscription upgrade failed for owner {owner.ownerid} by user #{self.requesting_user.ownerid}",
+                    extra=dict(pending_update=indication_of_payment_failure),
+                )
+            else:
+                # payment successful
```
ajay-sentry, line 257: will delinquent ever be true at this point where we'd want to set to false?

ajay-sentry (approved):
looks good! Could we add the stripe docs you were looking at to the review description?

services/billing.py
nora-codecov: love this suggestion, but "yes and"ed it with this change https://github.com/codecov/codecov-api/pull/831/files#diff-dda355f2f33801fe525e8ba51493017a8957e35710e9dc1d4affed3940e18685R56 (clearing delinquency in `update_plan`!)
From what I could tell, we are only using `plan_service.update_plan` in a situation where the `owner` should not be `delinquent`, so just make sure that field is set correctly while we're updating the owner obj.

PR #832, opened 2024-09-20
fix: Allow multiple owners per subscription in relevant stripe webhook handlers

spalmurray-codecov:
This PR updates the Stripe webhook handlers to accept requests for subscriptions shared by multiple owners.

It seems previously we baked in the expectation that this could not occur, but it turns out there are cases where this is a valid state. Specifically, this was the old way we handled users with multiple orgs billed jointly. With the new Account model, we can handle this case for Enterprise cloud users, but there still exists non-enterprise cloud users who have joint billing setups which cannot be migrated to Account billing as they are not on enterprise cloud plans. Support has a spreadsheet containing some of these customers, but it seems there are more than listed there. 

Note that we do not need to handle this case for customer_subscription_created as that can only occur during the normal billing flow, where multiple owners are not possible.

Fixes https://codecov.sentry.io/issues/4282794340
Fixes https://codecov.sentry.io/issues/2972074583
Closes https://github.com/codecov/engineering-team/issues/2200

### DIFF
--- a/billing/tests/test_views.py
+++ b/billing/tests/test_views.py
@@ -43,8 +43,17 @@ class MockSubscription(object):
 class StripeWebhookHandlerTests(APITestCase):
     def setUp(self):
         self.owner = OwnerFactory(
-            stripe_customer_id="20f0", stripe_subscription_id="3p00"
+            stripe_customer_id="cus_123", stripe_subscription_id="sub_123"
         )
+
+    # Creates a second owner that shares billing details with self.owner.
+    # This is used to test the case where owners are manually set to share a 
+    # subscription in Stripe.
+    def add_second_owner(self):
+        self.other_owner = OwnerFactory(
+            stripe_customer_id="cus_123", stripe_subscription_id="sub_123"
+        )
+
 
     def _send_event(self, payload, errorSig=None):
         timestamp = time.time_ns()
@@ -99,6 +108,31 @@ class StripeWebhookHandlerTests(APITestCase):
         assert response.status_code == status.HTTP_204_NO_CONTENT
         assert self.owner.delinquent is False
 
+    def test_invoice_payment_succeeded_sets_multiple_owners_delinquent_false(self):
+        self.add_second_owner()
+        self.owner.delinquent = True
+        self.owner.save()
+        self.other_owner.delinquent = True
+        self.other_owner.save()
+
+        response = self._send_event(
+            payload={
+                "type": "invoice.payment_succeeded",
+                "data": {
+                    "object": {
+                        "customer": self.owner.stripe_customer_id,
+                        "subscription": self.owner.stripe_subscription_id,
+                    }
+                },
+            }
+        )
+
+        self.owner.refresh_from_db()
+        self.other_owner.refresh_from_db()
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert self.owner.delinquent is False
+        assert self.other_owner.delinquent is False
+
     def test_invoice_payment_failed_sets_owner_delinquent_true(self):
         self.owner.delinquent = False
         self.owner.save()
@@ -118,6 +152,31 @@ class StripeWebhookHandlerTests(APITestCase):
         self.owner.refresh_from_db()
         assert response.status_code == status.HTTP_204_NO_CONTENT
         assert self.owner.delinquent is True
+
+    def test_invoice_payment_failed_sets_multiple_owners_delinquent_true(self):
+        self.add_second_owner()
+        self.owner.delinquent = False
+        self.owner.save()
+        self.other_owner.delinquent = False
+        self.other_owner.save()
+
+        response = self._send_event(
+            payload={
+                "type": "invoice.payment_failed",
+                "data": {
+                    "object": {
+                        "customer": self.owner.stripe_customer_id,
+                        "subscription": self.owner.stripe_subscription_id,
+                    }
+                },
+            }
+        )
+
+        self.owner.refresh_from_db()
+        self.other_owner.refresh_from_db()
+        assert response.status_code == status.HTTP_204_NO_CONTENT
+        assert self.owner.delinquent is True
+        assert self.other_owner.delinquent is True
 
     def test_customer_subscription_deleted_sets_plan_to_free(self):
         self.owner.plan = "users-inappy"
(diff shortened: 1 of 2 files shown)


billing/views.py
```diff
@@ -128,15 +128,31 @@ def subscription_schedule_updated(
                 ),
             )
 
-    def subscription_schedule_released(self, schedule: stripe.Subscription) -> None:
+    def subscription_schedule_released(self, schedule: stripe.SubscriptionSchedule) -> None:
```
spalmurray-codecov, line 131: I have manually tested all of these handlers with the mock Stripe API, with the exception of this one, which I couldn't figure out how to get to fire. Test cases seem to imply it works though.

billing/views.py
```diff
@@ -36,15 +38,12 @@ def invoice_payment_succeeded(self, invoice: stripe.Invoice) -> None:
                 stripe_subscription_id=invoice.subscription,
             ),
         )
-        owner = Owner.objects.get(
+        updated: int = Owner.objects.filter(
```
ajay-sentry, line 41: if this returns None for the first filter will the update throw an error?

billing/views.py
```diff
@@ -54,38 +53,40 @@ def invoice_payment_failed(self, invoice: stripe.Invoice) -> None:
...
             )
-            owner: Owner = Owner.objects.get(
-                stripe_customer_id=subscription.customer,
-                stripe_subscription_id=subscription.id,
-            )
-            plan_service = PlanService(current_org=owner)
+            return
+
+        for owner in owners:
+            plan_service = PlanService(owner)
```
ajay-sentry, line 85: Could you fill me in on the removal of current_org = owner in lieu of owner?

billing/views.py
```diff
@@ -54,38 +53,40 @@ def invoice_payment_failed(self, invoice: stripe.Invoice) -> None:
...
-            self._log_updated(1)
-        except Owner.DoesNotExist:
-            log.info(
-                "Customer Subscription Deleted - Couldn't find owner, subscription likely already deleted",
-                extra=dict(
-                    stripe_subscription_id=subscription.id,
-                    stripe_customer_id=subscription.customer,
-                ),
-            )
+        self._log_updated(owners.count())
```
ajay-sentry, line 89: Since we're adding this log here now, what are your thoughts on improving it further by sending the owner_ids into the log itself?

billing/views.py
```diff
@@ -54,38 +53,40 @@ def invoice_payment_failed(self, invoice: stripe.Invoice) -> None:
...
+            extra=dict(
+                stripe_subscription_id=subscription.id,
+                stripe_customer_id=subscription.customer,
+            ),
+        )
+        owners: QuerySet[Owner] = Owner.objects.filter(
+            stripe_customer_id=subscription.customer,
+            stripe_subscription_id=subscription.id,
+        )
+        if not owners.exists():
```
ajay-sentry, line 74: Is there a benefit between doing .exists() and just checking not Owners or owners is not None? 

Mostly for my curiosity

billing/views.py
```diff
@@ -128,15 +129,32 @@ def subscription_schedule_updated(
...
+                "Subscription schedule released requested with invalid subscription",
+                extra=dict(
+                    stripe_subscription_id=subscription.id,
+                    stripe_customer_id=subscription.customer,
+                    plan_id=subscription.plan.id,
+                ),
+            )
+            return
+
+        plan_services = [PlanService(current_org=owner) for owner in owners]
```
ajay-sentry, line 151: Should we keep a similar syntax like we did for subscription_deleted with a top level `for owner in owners` with all the business logic in there? Just to reduce cognitive load in this file

billing/views.py
```diff
@@ -217,26 +235,36 @@ def customer_subscription_created(self, subscription: stripe.Subscription) -> No
...
 
     def customer_subscription_updated(self, subscription: stripe.Subscription) -> None:
-        owner: Owner = Owner.objects.get(
+        owners: QuerySet[Owner] = Owner.objects.filter(
             stripe_subscription_id=subscription.id,
             stripe_customer_id=subscription.customer,
         )
+        if not owners.exists():
+            log.error(
+                "Subscription update requested with invalid plan",
```
ajay-sentry, line 244: I think this log message can be reworded; it's not necessarily due to invalid plan but because we couldn't find the owner for the combo

billing/views.py
spalmurray-codecov: `.filter()` will always return a QuerySet. It can be an empty queryset, but it never returns None. To be certain though, I did test this and there was no error thrown 🙂

billing/views.py
spalmurray-codecov: Didn't even notice this lol. The `PlanService` constructor only takes one argument (`current_org`), so either way works. Happy to change to the other way if you'd prefer!

billing/views.py
spalmurray-codecov: Great idea I'll do this!

billing/views.py
spalmurray-codecov: yes this make sense here!

billing/views.py
nora-codecov: This is a really good logic update - instead of 2 db hits (1 `.get()`, then 1 `.save()`) you've reduced it to 1 hit that is also more efficient, and you correctly recognized that we aren't doing anything with the `owner` object, we don't need to load that into memory. 
Since we're scrutinizing this section, what's the reasoning behind `_log_updated`? Why are we logging a count of objects updated with no other info, when would this be a useful metric? I understand logging for the error state (`Could not find customer`) but it's also not helpful if it contains no info about the operation that was attempted. What do y'all think?

spalmurray-codecov: @nora-codecov thanks for the comment! Re `_log_updated` see https://github.com/codecov/codecov-api/pull/832#discussion_r1771780317

billing/views.py
ajay-sentry: for consistency's sake, and for helping our future self out, I think we should update 😄 

(There's always a possibility we add another optional param to this function too)

billing/views.py
ajay-sentry: This is what I was missing! Thanks for clearing this up

billing/views.py
spalmurray-codecov: Updated @ajay-sentry!

billing/views.py
spalmurray-codecov: Update: I triggered this accidentally, and it seems to have worked - not sure how to do it again though 😆

billing/tests/test_views.py
```diff
@@ -443,6 +588,75 @@ def test_customer_subscription_updated_sets_free_and_deactivates_all_repos_if_in
             invoice_settings={"default_payment_method": "pm_1LhiRsGlVGuVgOrkQguJXdeV"},
         )
 
+    @patch("services.billing.stripe.PaymentMethod.attach")
+    @patch("services.billing.stripe.Customer.modify")
+    def test_customer_subscription_updated_sets_free_and_deactivates_all_repos_if_incomplete_expired_multiple_owner(
```
ajay-sentry, line 593: this test name makin my head spin ngl

ajay-sentry (approved):
magnificent 😤

PR #835, opened 2024-09-23
debug: Add phases to error log in owner serializer's get_scheduled_phases

spalmurray-codecov:
Logs out the phases leading to an error in owner serializer's get_scheduled_phases. To be used to assist in debugging a sentry issue I'm looking in to.

### DIFF
--- a/api/internal/owner/serializers.py
+++ b/api/internal/owner/serializers.py
@@ -236,7 +236,7 @@ class ScheduleDetailSerializer(serializers.Serializer):
                 extra=dict(
                     ownerid=schedule.metadata.obo_organization,
                     requesting_user_id=schedule.metadata.obo,
-                    phases=schedule["phases"],
+                    phases=schedule.get("phases", "no phases"),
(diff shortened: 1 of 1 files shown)


api/internal/owner/serializers.py
```diff
@@ -236,6 +236,7 @@ def get_scheduled_phase(self, schedule):
                 extra=dict(
                     ownerid=schedule.metadata.obo_organization,
                     requesting_user_id=schedule.metadata.obo,
+                    phases=schedule["phases"],
```
rohitvinnakota-codecov, line 239: Are we sure this key exists? otherwise it will throw an error 

You can try, `phases=schedule.get("phases", "no phases"))` or something similar

spalmurray-codecov: I believe this will only be called if phases exist, but I'll add your changes anyway - can't hurt 🙂

rohitvinnakota-codecov (approved):

PR #853, opened 2024-10-03
feat: Add totalSeatCount resolver on Account

spalmurray-codecov:
Adds a resolver on Account to return totalSeatCount, which is the sum of plan seats and free seats.

Closes https://github.com/codecov/engineering-team/issues/2556


nora-codecov (approved):

graphql_api/types/account/account.graphql
```diff
@@ -1,4 +1,5 @@
 type Account {
   name: String!
   oktaConfig: OktaConfig
+  totalSeatCount: Int
```
rohitvinnakota-codecov, line 4: nit: should this be Int!

rohitvinnakota-codecov (approved):

graphql_api/types/account/account.graphql
spalmurray-codecov: I was copying the `config.seatCount` type, but yes I think you're right, it is always defined

spalmurray-codecov: will update

PR #857, opened 2024-10-04
feat: add history, fix percent_diff rounding and flake aggregate time

joseph-sentry:
- add history to TestResults, TestResultsAggregate and FlakeAggregate GQL models
- fix percent_diff function rounding to round up to 5 digits
- fix flake aggregate time bound checking to be accurate

### DIFF
--- a/graphql_api/types/repository/repository.py
+++ b/graphql_api/types/repository/repository.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime
+from datetime import datetime, timedelta
 from typing import List, Mapping, Optional
 
 import shared.rate_limits as rate_limits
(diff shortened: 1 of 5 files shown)


graphql_api/types/repository/repository.py
```diff
@@ -587,13 +616,24 @@ def resolve_coverage_analytics(
...
 ):
+    history = convert_history_to_timedelta(history)
     return await sync_to_async(generate_test_results_aggregates)(
-        repoid=repository.repoid
+        repoid=repository.repoid, history=history
     )
 
 
 @repository_bindable.field("flakeAggregates")
 @convert_kwargs_to_snake_case
```
ajay-sentry, line 629: we actually don't need the @convert_kwargs decorator anymore due to https://github.com/codecov/codecov-api/blob/9aa6458ad023fe0ec55993ad9134bbd0af5fb3f7/graphql_api/schema.py#L7

utils/test_results.py
```diff
@@ -296,11 +296,8 @@ def get_flake_aggregate_numbers(
         flakes = Flake.objects.filter(
             Q(repository_id=repo.repoid)
             & (
-                (Q(end_date__isnull=True))
-                | (
-                    Q(end_date__date__lte=until.date())
-                    & Q(end_date__date__gt=since.date())
-                )
+                Q(start_date__date__lte=until.date())
```
ajay-sentry, line 299: start date < until date but end date > since date?

Could you remind me again what start and end mean in this context?
I'd figure we'd want start > until and end < since but I could just be messing with my head

PR #858, opened 2024-10-04
add UploadTokenRequiredAuthenticationCheck

nora-codecov:
### Purpose/Motivation
NEW auth class to check for TrueTokenless

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2298

### What does this PR do?
This will be placed at the top of the list for the auth methods on upload endpoints (so it is checked first). In a separate pr.


codecov_auth/tests/unit/test_repo_authentication.py
```diff
@@ -554,3 +555,130 @@ def test_tokenless_success(
...
+    def test_token_not_required_matches_paths(
+        self, request_uri, repo_slug, commitid, private, token_required, db
+    ):
+        author_name, repo_name = repo_slug.split("/")
+        # Doing this because of ATS.
+        # For pytest '::' is the divider between a test class and a test function.
+        # There's a non-zero chance that the full test name would be mis-interpreted by pytest
+        # And these tests would never work with ATS.
+        # Sadly there's no way to escape the '::' sequence
+        request_uri = request_uri.replace("SEPARATOR", "::::")
```
giovanni-guidini, line 615: [nit] Ah this is no longer necessary.
Well 1 because we're not using ATS anymore, so who cares.
But 2 because @michelletran-codecov showed me that parametrized tests [can have an id](https://docs.pytest.org/en/stable/example/parametrize.html#different-options-for-test-ids).

So we can just add custom ids to the tests and we don't have to leave that SEPARATOR there.

I'm personally favor `pytest.param` with `id=` keyword argument

upload/views/helpers.py
```diff
@@ -5,22 +5,28 @@
 
 
 def get_repository_from_string(
-    service: Service, repo_identifier: str
-) -> typing.Optional[Repository]:
+    service: Service, repo_identifier: str, include_owner: bool = False
+) -> typing.Optional[Repository] | tuple[Repository | None, Owner | None]:
```
giovanni-guidini, line 9: [nit] 
this `include_owner` is unnecessarily complicating this function.

And it looks like it was added for the benefit of `get_repository_and_owner_from_string`. We have a `_get_owner_from_string` already.

I'd personally think the extra logic should be moved to `get_repository_and_owner_from_string`. Have that make use of the 2 other funcitons.

giovanni-guidini (approved):
I like how you made it easier to reuse code in the tests. Particularly having the `valid_params_to_test` be an independent list that can be reused 👍 

Also left a few nit comments

upload/views/helpers.py
```diff
@@ -32,3 +38,11 @@ def _get_owner_from_string(
...
     except Owner.DoesNotExist:
         return None
+
+
+def get_repository_and_owner_from_string(
+    service: Service, repo_identifier: str
+) -> tuple[Repository | None, Owner | None]:
+    return get_repository_from_string(
+        service=service, repo_identifier=repo_identifier, include_owner=True
+    )
```
Swatinem, line 48: I would actually turn this around:

Move all the logic that does the resolution into `get_repository_and_owner_from_string`, and remove that `include_owner` parameter.
Then call this from `get_repository_from_string`, returning just `result[0]`, which should make things simpler.

The type system can’t reason about that a `True` parameter means you get a `tuple`, and otherwise a `Repository | None`.
So I’m pretty sure the code as is will cause a type error.

nora-codecov: great suggestion!! updating...

upload/views/helpers.py
nora-codecov: agree! updating...

PR #862, opened 2024-10-07
feat: Add activatedUserCount resolver to Owner

spalmurray-codecov:
Adds activatedUserCount resolver on the Owner type. One small step towards removing the account details rest endpoint lol.

Closes https://github.com/codecov/engineering-team/issues/2555

### DIFF
--- a/graphql_api/types/owner/owner.py
+++ b/graphql_api/types/owner/owner.py
@@ -6,6 +6,7 @@ import shared.rate_limits as rate_limits
 import stripe
 import yaml
 from ariadne import ObjectType, convert_kwargs_to_snake_case
+from graphql import GraphQLResolveInfo
 
 import services.activation as activation
 import timeseries.helpers as timeseries_helpers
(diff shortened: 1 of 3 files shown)


graphql_api/types/owner/owner.py
```diff
@@ -362,3 +367,10 @@ def resolve_ai_enabled_repos(owner: Owner, info) -> List[str] | None:
         )
 
     return list(queryset.values_list("name", flat=True))
+
+
+@owner_bindable.field("activatedUserCount")
+@sync_to_async
+@require_part_of_org
+def resolve_activated_user_count(owner: Owner, info: GraphQLResolveInfo) -> int | None:
```
JerrySentry, line 375: Don't think this can return None

JerrySentry (approved):

graphql_api/types/owner/owner.py
spalmurray-codecov: You are correct, I figured it needed to be there for when user is not in the org, but this is not true! Pushing fix.

PR #863, opened 2024-10-07
feat: Add organizations array resolver on Account type

spalmurray-codecov:
Adds the paginated organizations array resolver on the Account type in the graphql API.

Closes https://github.com/codecov/engineering-team/issues/2558

### DIFF
--- a/graphql_api/types/account/__init__.py
+++ b/graphql_api/types/account/__init__.py
@@ -1,10 +1,5 @@
 from shared.license import get_current_license
 
-from graphql_api.helpers.ariadne import ariadne_load_local_graphql
+from .account import account, account_bindable
 
-from .account import account_bindable
-
-account = ariadne_load_local_graphql(__file__, "account.graphql")
-
-
-__all__ = ["get_current_license", "account_bindable"]
+__all__ = ["get_current_license", "account_bindable", "account"]
--- a/graphql_api/types/account/__init__.py
+++ b/graphql_api/types/account/__init__.py
@@ -1,5 +1,3 @@
-from shared.license import get_current_license
-
 from .account import account, account_bindable
 
-__all__ = ["get_current_license", "account_bindable", "account"]
+__all__ = ["account_bindable", "account"]
--- a/graphql_api/types/account/account.graphql
+++ b/graphql_api/types/account/account.graphql
@@ -3,4 +3,12 @@ type Account {
   oktaConfig: OktaConfig
   totalSeatCount: Int!
   activatedUserCount: Int!
+  organizations(
+    ordering: AccountOrganizationOrdering
+    orderingDirection: OrderingDirection
+    first: Int
+    after: String
+    last: Int
+    before: String
+  ): AccountOrganizationConnection! @cost(complexity: 25, multipliers: ["first", "last"])
 }
--- a/graphql_api/types/account/account.py
+++ b/graphql_api/types/account/account.py
@@ -1,4 +1,4 @@
-from ariadne import ObjectType, convert_kwargs_to_snake_case
+from ariadne import ObjectType
 from graphql import GraphQLResolveInfo
 
 from codecov.db import sync_to_async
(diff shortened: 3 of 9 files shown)


graphql_api/types/account/account.graphql
```diff
@@ -3,4 +3,12 @@ type Account {
...
   totalSeatCount: Int!
   activatedUserCount: Int!
+  organizations(
+    ordering: AccountOrganizationOrdering
+    orderingDirection: OrderingDirection
+    first: Int
+    after: String
+    last: Int
+    before: String
+  ): AccountOrganizationConnection! @cost(complexity: 25, multipliers: ["first", "last"])
```
spalmurray-codecov, line 13: complexity is same as on user organizations list. Seems to make sense to keep things consistent.

graphql_api/types/account/__init__.py
```diff
@@ -1,10 +1,5 @@
...
-from graphql_api.helpers.ariadne import ariadne_load_local_graphql
+from .account import account, account_bindable
 
-from .account import account_bindable
-
-account = ariadne_load_local_graphql(__file__, "account.graphql")
-
-
-__all__ = ["get_current_license", "account_bindable"]
+__all__ = ["get_current_license", "account_bindable", "account"]
```
ajay-sentry, line 5: I noticed this in a lot of GQL files but idk if we actually need get_current_license here, do we?

graphql_api/types/account/account.py
```diff
@@ -29,3 +37,20 @@ def resolve_total_seat_count(account: Account, info: GraphQLResolveInfo) -> int:
...
+def resolve_organizations(
+    account: Account,
+    info: GraphQLResolveInfo,
+    ordering=AccountOrganizationOrdering.ACTIVATED_USERS,
+    ordering_direction=OrderingDirection.DESC,
+    **kwargs,
+):
+    return queryset_to_connection(
+        account.organizations,
+        ordering=(ordering,),
```
ajay-sentry, line 53: do we need the tuple here for ordering? vs ordering=ordering

graphql_api/types/account/account.py
```diff
@@ -29,3 +37,20 @@ def resolve_total_seat_count(account: Account, info: GraphQLResolveInfo) -> int:
 @sync_to_async
 def resolve_activated_user_count(account: Account, info: GraphQLResolveInfo) -> int:
     return account.activated_user_count
+
+
+@account_bindable.field("organizations")
+@convert_kwargs_to_snake_case
```
ajay-sentry, line 43: we no longer need this decorator since we have this now! https://github.com/codecov/codecov-api/blob/9aa6458ad023fe0ec55993ad9134bbd0af5fb3f7/graphql_api/schema.py#L7

graphql_api/types/account/account.py
```diff
@@ -29,3 +37,20 @@ def resolve_total_seat_count(account: Account, info: GraphQLResolveInfo) -> int:
...
+def resolve_organizations(
+    account: Account,
+    info: GraphQLResolveInfo,
+    ordering=AccountOrganizationOrdering.ACTIVATED_USERS,
```
ajay-sentry, line 47: just to confirm, default ordering is by plan_activated_users in descending order

graphql_api/types/account/account.py
spalmurray-codecov: Yes we do. Not sure exactly how this works under the hood, but the queryset_to_connection helper does a `for field in ordering`. Some other calls do pass multiple values in the tuple, so seems like there's a reason for this.

graphql_api/types/account/__init__.py
spalmurray-codecov: You're right lol. I removed the unnecessary instances and tests are all green.

graphql_api/types/account/account.py
ajay-sentry: Got it 👌 Thanks for clarifying!

ajay-sentry (approved):
send it ✅

PR #868, opened 2024-10-08
feat: add history, fix percent_diff rounding and flake aggregate time

suejung-sentry:
Ported changes from https://github.com/codecov/codecov-api/pull/857 with merge conflicts resolved

### DIFF
--- a/graphql_api/types/test_analytics/test_analytics.py
+++ b/graphql_api/types/test_analytics/test_analytics.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import timedelta
 
 from ariadne import ObjectType, convert_kwargs_to_snake_case
 from graphql.type.definition import GraphQLResolveInfo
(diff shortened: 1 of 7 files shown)


joseph-sentry:
Gonna address @ajay-sentry's feedback on this PR:

> start date < until date but end date > since date?
> 
> Could you remind me again what start and end mean in this context?
> I'd figure we'd want start > until and end < since but I could just be messing with my head

this one is a bit confusing but the idea is that we want flakes that were active at some point during the time period starting at `since` and ending at `until` so, if it started before `until` and it ended after `since`. This is also assuming that `end > start` which is an obvious assumption to make.
1. if it started after `until` then it occured after the time period
2. if it ended before `start` then it occured before the time period
3. otherwise it occured during the time period

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -69,13 +69,54 @@ async def resolve_results(
...
 ):
+    history = convert_history_to_timedelta(history)
     return await sync_to_async(generate_test_results_aggregates)(
-        repoid=repository.repoid
+        repoid=repository.repoid, history=history
     )
 
 
 @test_analytics_bindable.field("flakeAggregates")
 @convert_kwargs_to_snake_case
```
ajay-sentry, line 82: Dont need these @convert_kwargs_to_snake_case

ajay-sentry (approved):
looks good

PR #874, opened 2024-10-08
add UploadTokenRequiresAuthenticationCheck to core upload endpoints

nora-codecov:
### Purpose/Motivation
This adds True Tokenless to the main upload endpoints. Did not add to endpoints that do not have the current Tokenless auth class.

Feature is not yet available to public so should not impact users at all.

BA/TA endpoints need customization to work with True Tokenless, those will come next.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2298


upload/views/test_results.py
```diff
@@ -32,6 +33,7 @@
 
 class UploadTestResultsPermission(BasePermission):
     def has_permission(self, request, view):
+        print("##################")
```
Swatinem, line 36:
```suggestion
```

Swatinem (approved):

PR #879, opened 2024-10-10
Add more fields for Test Analytics

joseph-sentry:
- add total fail, skip and pass counts to the TestResult GQL model
- add total slow test number to test results aggregates
- allow filtering test results by name using a search term


ajay-sentry (approved):
lgtm

PR #880, opened 2024-10-10
perf: Fix TestResults SQL query to not take as long

joseph-sentry:
this commit replaces the previous django orm code with an invocation
to do a raw sql query that performs much better for repos with a large
amount of data.

one worry with this approach is the risk for sql injection

in this specific case, we're passing most of the dynamic user provided
values used in the sql through the parameters which the django docs say
are sanitized and should be used to protect against sql injection

for the user provided values that are not passed through the sql parameters
they're checked to be in a strict set of values, so we shouldn't be substituting
any unexpected strings into the query

### DIFF
--- a/graphql_api/tests/test_test_result.py
+++ b/graphql_api/tests/test_test_result.py
@@ -197,7 +197,7 @@ class TestResultTestCase(GraphQLTestHelper, TransactionTestCase):
             result["owner"]["repository"]["testAnalytics"]["testResults"]["edges"][0][
                 "node"
             ]["lastDuration"]
-            == 0.0
+            == 1.0
         )
 
     def test_fetch_test_result_avg_duration(self) -> None:
--- a/utils/test_results.py
+++ b/utils/test_results.py
@@ -61,7 +61,7 @@ class GENERATE_TEST_RESULT_PARAM:
 @dataclass
 class TestResultsQuery:
     query: str
-    params: list[int | str | tuple[str, ...]]
+    params: dict[str, int | str | tuple[str, ...]]
 
 
 def convert_tuple_or_none(value: set[str] | list[str] | None) -> tuple[str, ...] | None:
@@ -120,20 +120,20 @@ def generate_base_query(
         [f"with_cursor.{order} {ordering_direction.name}" for order in ordering]
     )
 
-    params: list[int | str | tuple[str, ...] | None] = [
-        repoid,
-        f"{interval_num_days} days",
-        branch,
-        convert_tuple_or_none(test_ids),
-        convert_tuple_or_none(testsuites),
-        term_filter,
-        encode_after_or_before(after),
-        encode_after_or_before(before),
-        page_size,
-    ]
-    filtered_params: list[int | str | tuple[str, ...]] = [
-        p for p in params if p is not None
-    ]
+    params: dict[str, int | str | tuple[str, ...] | None] = {
+        "repoid": repoid,
+        "interval": f"{interval_num_days} days",
+        "branch": branch,
+        "test_ids": convert_tuple_or_none(test_ids),
+        "testsuites": convert_tuple_or_none(testsuites),
+        "term": term_filter,
+        "after": encode_after_or_before(after),
+        "before": encode_after_or_before(before),
+        "limit": page_size,
+    }
+    filtered_params: dict[str, int | str | tuple[str, ...]] = {
+        k: v for k, v in params.items() if v is not None
+    }
 
     base_query = f"""
 with
@@ -142,12 +142,12 @@ base_cte as (
 	from reports_dailytestrollups rd
     { "join reports_test rt on rt.id = rd.test_id" if testsuites or term else ""}
 	where
-        rd.repoid = %s
-		and rd.date > current_date - interval %s
-        { "and rd.branch = %s" if branch else ""}
-        { "and rd.test_id in %s" if test_ids else ""}
-        { "and rt.testsuite in %s" if testsuites else ""}
-        { "and rt.name like %s" if term else ""}
+        rd.repoid = %(repoid)s
+		and rd.date > current_date - interval %(interval)s
+        { "and rd.branch = %(branch)s" if branch else ""}
+        { "and rd.test_id in %(test_ids)s" if test_ids else ""}
+        { "and rt.testsuite in %(testsuites)s" if testsuites else ""}
+        { "and rt.name like %(term)s" if term else ""}
 ),
 failure_rate_cte as (
 	select
@@ -201,10 +201,10 @@ select * from (
         full outer join last_duration_cte using (test_id)
     ) as results join reports_test rt on results.test_id = rt.id
 ) as with_cursor
-{"where with_cursor._cursor > %s" if after else ""}
-{"where with_cursor._cursor < %s" if before else ""}
+{"where with_cursor._cursor > %(before)s" if after else ""}
+{"where with_cursor._cursor < %(after)s" if before else ""}
 order by {order_by}
-limit %s
+limit %(limit)s
 """
     return TestResultsQuery(query=base_query, params=filtered_params)
 
--- a/utils/test_results.py
+++ b/utils/test_results.py
@@ -112,12 +112,12 @@ def generate_base_query(
         [
             f"rt.{order_field}"
             if order_field in {"name", "computed_name"}
-            else f"foo.{order_field}"
+            else f"results.{order_field}"
             for order_field in ordering
         ]
     )
     order_by = ",".join(
-        [f"bar.{order} {ordering_direction.name}" for order in ordering]
+        [f"with_cursor.{order} {ordering_direction.name}" for order in ordering]
     )
 
     params: list[int | str | tuple[str, ...] | None] = [
@@ -149,7 +149,7 @@ base_cte as (
         { "and rt.testsuite in %s" if testsuites else ""}
         { "and rt.name like %s" if term else ""}
 ),
-cte1 as (
+failure_rate_cte as (
 	select
 		test_id,
 		CASE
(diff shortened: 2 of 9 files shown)


matt-codecov (commented):
chiming in on the SQL injection risk: currently looks good. `{order_by}` and `{order}` are computed based on validated data, and every other string substitution controls the presence/omission of a hardcoded clause, nothing dynamic. the hardcoded clauses use query parameters and any dynamic values are provided separately

https://docs.djangoproject.com/en/5.1/topics/db/sql/#passing-parameters-into-raw it looks like you can name your query parameters like `%(repoid)s` and then make your `params` list a dict instead where the key is the parameter name. that would probably make this query a lot easier to work with

i imagine Django chose to use Python's % string interpolation syntax for their parameterized query syntax to make switching to safe queries easier, but man i wish the safe code didn't look so much like the dangerous code

do you have a sample of the SQL that was generated with the Django ORM query? i am not a django-optimizing expert but dynamically generating a query like this is fragile enough that it's worth a second look to see if we can't improve the performance in place

utils/test_results.py
```diff
@@ -59,15 +58,201 @@ class GENERATE_TEST_RESULT_PARAM:
     SKIPPED = "skipped"
 
 
+@dataclass
+class TestResultsQuery:
+    query: str
+    params: dict[str, int | str | tuple[str, ...]]
+
+
+def convert_tuple_or_none(value: set[str] | list[str] | None) -> tuple[str, ...] | None:
```
ajay-sentry, line 67: convert_to_tuple_else_none maybe?

utils/test_results.py
```diff
@@ -59,15 +58,201 @@ class GENERATE_TEST_RESULT_PARAM:
...
+    return tuple(value) if value else None
+
+
+def encode_after_or_before(value: str | None) -> str | None:
```
ajay-sentry, line 71: this function is actually just encode_to_b64 isnt it? Nothing in particular to after or before

utils/test_results.py
```diff
@@ -59,15 +58,201 @@ class GENERATE_TEST_RESULT_PARAM:
...
+    interval_num_days: int,
+    testsuites: list[str] | None = None,
+    term: str | None = None,
+    test_ids: set[str] | None = None,
+) -> TestResultsQuery:
+    page_size = first or last
+
+    term_filter = f"%{term}%" if term else None
+
+    if interval_num_days not in {1, 7, 30}:
```
ajay-sentry, line 93: nit: Im kinda going back and forth with moving all this validation logic into a separate "validate_params" fn or similar to kinda clean up this function a bit

utils/test_results.py
```diff
@@ -59,15 +58,201 @@ class GENERATE_TEST_RESULT_PARAM:
...
+        SUM(skip_count) as total_skip_count
+	from base_cte
+	group by test_id
+),
+commits_where_fail_cte as (
+	select test_id, array_length((array_agg(distinct unnested_cwf)), 1) as failed_commits_count from (
+		select test_id, commits_where_fail as cwf
+		from base_cte
+		where array_length(commits_where_fail,1) > 0
+	) as foo, unnest(cwf) as unnested_cwf group by test_id
```
ajay-sentry, line 176: foo

utils/test_results.py
```diff
@@ -76,132 +261,143 @@ def generate_test_results(
...
-                totals.values("test")
-                .annotate(fail_count_sum=Sum("fail_count"))
-                .filter(fail_count_sum__gt=0)
-                .values("test_id")
-            )
+        totals = totals.filter(test__name__icontains=term).values("test_id")
+
+        filtered_test_ids = set([test["test_id"] for test in totals])
+
+        test_ids = test_ids & filtered_test_ids if test_ids else filtered_test_ids
```
ajay-sentry, line 291: this intersection is quicker than doing a filter on term prior to flags?

utils/test_results.py
```diff
@@ -76,132 +261,143 @@ def generate_test_results(
...
+                totals = DailyTestRollup.objects.filter(repoid=repoid, date__gt=since)
+
+                skipped_test_ids = (
+                    totals.values("test")
+                    .annotate(
+                        skip_count_sum=Sum("skip_count"),
+                        fail_count_sum=Sum("fail_count"),
+                        pass_count_sum=Sum("pass_count"),
+                    )
+                    .filter(skip_count_sum__gt=0, fail_count_sum=0, pass_count_sum=0)
```
ajay-sentry, line 330: what are your thoughts on not having the fail_count_sum check here --- signaling tests which were constantly failing that were then skipped

utils/test_results.py
```diff
@@ -76,132 +261,143 @@ def generate_test_results(
...
+                test_ids = (
+                    test_ids & slowest_test_id_set if test_ids else slowest_test_id_set
+                )
 
-    commits_where_fail_sq = (
-        totals.filter(test_id=OuterRef("test_id"))
-        .annotate(v=Distinct(Unnest(F("commits_where_fail"))))
-        .values("v")
+    if not first and not last:
+        first = 25
```
ajay-sentry, line 359: I think 20 is our default typically

utils/test_results.py
joseph-sentry: I guess it makes sense to have tests that have never passed here, but there's no way of knowing if it's like interleaved in the way that its failing and skipped, i'm cool with trying it though

graphql_api/tests/test_test_result.py
```diff
@@ -234,7 +234,7 @@ def test_fetch_test_result_last_duration(self) -> None:
             result["owner"]["repository"]["testAnalytics"]["testResults"]["edges"][0][
                 "node"
             ]["lastDuration"]
-            == 0.0
```
ajay-sentry, line 237: do you know why this guy changed?

utils/test_results.py
```diff
@@ -1,73 +1,295 @@
...
+    branch: str | None,
+    interval_num_days: int,
+    testsuites: list[str] | None = None,
+    term: str | None = None,
+    test_ids: set[str] | None = None,
+) -> TestResultsQuery:
+    term_filter = f"%{term}%" if term else None
+
+    if should_reverse:
+        ordering_direction = "DESC" if ordering_direction == "ASC" else "ASC"
```
ajay-sentry, line 152: is this a typo?

utils/test_results.py
```diff
@@ -1,73 +1,295 @@
...
+        # -1 means row value is to the left of the cursor value (search to the right)
+        # 0 means row value is equal to cursor value
+        # 1 means row value is to the right of the cursor value (search to the left)
+        row_value = getattr(row, ordering)
+        row_value_str = str(row_value)
+        cursor_value_str = cursor.ordered_value
+        return (row_value_str > cursor_value_str) - (row_value_str < cursor_value_str)
+
+    left, right = 0, len(rows) - 1
+    while left <= right:
```
ajay-sentry, line 258: holy smokes is this binary search in the wild?

utils/test_results.py
```diff
@@ -1,73 +1,295 @@
...
+        print(left, right)
```
ajay-sentry, line 259: print

utils/test_results.py
```diff
@@ -1,73 +1,295 @@
...
+        full outer join last_duration_cte using (test_id)
+    ) as results join reports_test rt on results.test_id = rt.id
+) as with_cursor
+order by {order_by}
+"""
+
+    return TestResultsQuery(query=base_query, params=filtered_params)
+
+
+def search_base_query(
```
ajay-sentry, line 240: Why are we doing all this? A function description would go a long way for this guy

utils/test_results.py
```diff
@@ -76,131 +298,152 @@ def generate_test_results(
     :param branch: optional name of the branch we want to filter on, if this is provided the aggregates calculated will only take into account
         test instances generated on that branch. By default branches will not be filtered and test instances on all branches wil be taken into
         account.
-    :param history: timedelta for filtering test instances used to calculated the aggregates by time, the test instances used will be
-        those with a created at larger than now - history.
+    :param interval: timedelta for filtering test instances used to calculated the aggregates by time, the test instances used will be
```
ajay-sentry, line 301: used to calculate*

utils/test_results.py
```diff
@@ -76,131 +298,152 @@ def generate_test_results(
...
+        those with a created at larger than now - interval.
     :param testsuites: optional list of testsuite names to filter by
```
ajay-sentry, line 303: test suites filtering is also union right?

utils/test_results.py
```diff
@@ -76,131 +298,152 @@ def generate_test_results(
...
-        last_duration=Value(0.0),
-        avg_duration=Avg("avg_duration_seconds"),
-        name=F("test__name"),
-        computed_name=F("test__computed_name"),
-    )
+        aggregation_of_test_results = cursor.fetchall()
+
+        rows = [TestResultsRow(*row) for row in aggregation_of_test_results]
+
+    page_size: int = first or last or 20
```
ajay-sentry, line 425: at this point first or last will always have a value thanks to your check on 399 right?

utils/test_results.py
joseph-sentry: indeed i will add that as a comment

utils/test_results.py
joseph-sentry: nope, we want to reverse the ordering direction when we get `last` passed to the query

graphql_api/tests/test_test_result.py
ajay-sentry: ahhhh okay yeah i remember

utils/test_results.py
ajay-sentry: oh I see, yeah that makes sense

ajay-sentry (approved):

PR #882, opened 2024-10-10
True Tokenless for BA and TA

nora-codecov:
### Purpose/Motivation
TT requires certain identifiers, BA and TA don't have the identifiers in their path, so I make custom methods to get the identifiers so that BA and TA can use TT.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2877


codecov_auth/authentication/repo_auth.py
```diff
@@ -398,3 +404,86 @@ def authenticate(
...
+        self, request: HttpRequest
+    ) -> tuple[Repository | None, Owner | None]:
+        from upload.views.test_results import UploadSerializer  # circular imports
+
+        try:
+            body = json.loads(str(request.body, "utf8"))
+        except json.JSONDecodeError:
+            return None, None  # continue to next auth class
+
+        serializer = UploadSerializer(data=body)
```
joseph-sentry, line 427: hmmm, would it be faster/cleaner if we access the parts of the request body we care about directly instead of using the Serializer here?

upload/views/helpers.py
```diff
@@ -45,3 +50,45 @@ def get_repository_from_string(
...
+        author__username=owner_username, name=repository_name
+    ).select_related("author")
+    if matching_repositories.count() == 1:
+        log.info(
+            "get_repository_and_owner_from_slug_and_commit success",
+            extra=dict(slug=slug),
+        )
+        repository = matching_repositories.first()
+        return repository, repository.author
+    else:
```
joseph-sentry, line 77: is there ever a case where matching_repositories is empty? I think that may be the case if the user has never synced repos

joseph-sentry (commented):
left a couple of comments on the test results stuff that may also apply to BA

upload/views/helpers.py
```diff
@@ -45,3 +50,45 @@ def get_repository_from_string(
...
     )
     return repository
+
+
+def get_repository_and_owner_from_slug_and_commit(
+    slug: str, commitid: str
+) -> tuple[Repository | None, Owner | None]:
+    if "::::" not in slug:
+        return None, None
+
```
JerrySentry, line 60: My understanding for this is to try to find the Repo and Owner based on the commit SHA because there's no service available. But we are ok with an astronomically low probability that we have duplicate entries in the DB with ("username", "repo name", "commit SHA in the repo")?

upload/views/helpers.py
```diff
@@ -45,3 +50,45 @@ def get_repository_from_string(
...
+    owner_username, repository_name = slug.rsplit("::::", 1)
+
+    # formatting for GL subgroups
+    if ":::" in owner_username:
+        owner_username = owner_username.replace(":::", ":")
+
```
JerrySentry, line 66: IMO its cleaner to just to do 1 queryset instead of possibly two, I don't think there would be any significant performance hit with the way these 3 tables are joined. The function can be something like this:
```
queryset = Repository.objects.filter(
    name=repository_name,
    author__username=username,
    commit__commitid=commitid
).select_related('author')

if not queryset.exists() or queryset.count() > 1:
    return None, None
    
repository = queryset.first()
return repository, repository.author

File: upload/views/helpers.py (hunk 3 above)
Comment (nora-codecov) [reply upload/views/helpers.py:60]: this is a great suggestion - but it would only work if it's guaranteed that the commit already exists in our db. 

With BA and TA, in the views we are doing a `get_or_create` with the incoming commitid, so I took that to mean that the incoming request might be the first upload for that commit. 
commitid is required so the incoming request will always have a commitid, however, if the commit isn't in our db yet then `AND`ing the 3 fields together in this way guarantees no match.

File: upload/views/helpers.py (hunk 4 above)
Comment (nora-codecov) [reply upload/views/helpers.py:66]: I agree, but it would only work if the commit already exists in our db.

Could we argue that this is a stronger match, therefore try this first, and if there's no match try again without commitid?

File: codecov_auth/authentication/repo_auth.py (hunk 1 above)
Comment (nora-codecov) [reply codecov_auth/authentication/repo_auth.py:427]: I think it's best practices to not interact with incoming data that hasn't been validated, particularly since there's no token here 😅 

I'm using the serializer to type-check the incoming fields and make sure they exist. If I didn't use the serializer, I would have to do the check that they exist manually, I'd rather have the serializer do that for me

File: upload/views/helpers.py (hunk 2 above)
Comment (nora-codecov) [reply upload/views/helpers.py:77]: yes - I think it's edge case but possible to have no matches in this repo query. I'm not kicking them out at this point because this query being empty means the secondary query with the commitid won't find any matches, but it means I am doing the secondary query knowing that it will fail. I can change that

File: upload/views/helpers.py (hunk 4 above)
Comment (JerrySentry) [reply upload/views/helpers.py:66]: Ah you're right on first call there is no Commit in the DB.. that means in the edge case that we have duplicate Owner/Repo pairs then the true tokenless auth would fail.

The other idea to consider is to require BA and TA upload endpoints to send in a `git_service` value in the request body to identify the git provider if using TT, this way the triplet of owner.username, repo.name, and owner.service will guarantee a one_or_none. And with that you don't need the Commit table nor the commitid from the request. Also no possibility of not supporting TT for duplicate Owner/Repo pairs. This is how the other (non-true) tokenless auth works for BA.

File: upload/views/helpers.py (hunk 3 above)
Comment (nora-codecov) [reply upload/views/helpers.py:60]: er, sorry, my comment above is more suited to your other suggestion.

The possibility exists for us to not be able to find a distinct match. This is a major drawback to this implementation, it means we can't guarantee tokenless uploads on BA and TA for everyone - it depends on what's in our db.

I don't like this, but without `service` or other distinct identifiers for the owner or repo, it's all we can do. I'm hoping that it works for like 95% of uploads, and if it didn't, they could re-upload once the commit is in the db. That's why I'm logging the different outcomes, so I can see how successful it is.

File: codecov_auth/authentication/repo_auth.py (hunk 1 above)
Comment (joseph-sentry) [reply codecov_auth/authentication/repo_auth.py:427]: ok 👍

File: upload/views/helpers.py (hunk 2 above)
Comment (joseph-sentry) [reply upload/views/helpers.py:77]: yeah i think it's unlikely the repo won't exist, since users would typically need to get the upload token to upload and they'd get that from the UI, i think it's fine that it crashes later during the commitid step

File: codecov_auth/authentication/repo_auth.py [hunk 5]
```diff
@@ -398,3 +403,56 @@ def authenticate(
...
+    """
+    Get Repository and Owner from request body instead of path,
+    then use the same authenticate() as parent class.
+    """
+
+    def _get_git(self, validated_data):
+        """
+        BA sends this in as git_service, TA sends this in as service.
+        Use this function so this Check class can be used by both views.
+        """
```
JerrySentry, line 426: Could probably be `return validated_data.get("git_service") or validated_data.get("service")`

upload/views/test_results.py
```diff
@@ -39,13 +39,13 @@ def has_permission(self, request, view):
 class UploadSerializer(serializers.Serializer):
     commit = serializers.CharField(required=True)
     slug = serializers.CharField(required=True)
+    service = serializers.CharField(required=False)  # git_service
```
JerrySentry, line 42: How come TA can't use `git_service`?

upload/views/helpers.py
```diff
@@ -1,8 +1,13 @@
+import logging
```
JerrySentry, line 1: This file can now be deleted right?

upload/views/test_results.py
nora-codecov: Since it's already live in prod I figured it would be a whole thing to switch over to a different var name - what do you think @joseph-sentry ?

upload/views/helpers.py
nora-codecov: I can delete `get_repository_and_owner_from_slug_and_commit`, my bad I thought I had removed it! ty for catching.

upload/views/test_results.py
```diff
@@ -156,7 +158,7 @@ def post(self, request):
             "build_url": data.get("buildURL"),  # build_url
             "job": data.get("job"),  # job_code
             "flags": data.get("flags"),
-            "service": data.get("ci_service"),  # provider
+            "service": data.get("service"),  # git provider
             "url": storage_path,  # storage_path
             # these are used for dispatching the task below
             "commit": commit.commitid,
```
nora-codecov, line 164: @joseph-sentry can you double check the changes I made to the TA fields? Is this still the design we're going with?

JerrySentry (approved):
looks good on bundle analysis side of things.

upload/views/test_results.py
nora-codecov: just confirmed, TA sends `git_service` as `service` https://github.com/codecov/codecov-cli/blob/2935526caf5db25d64b02870f9cc45e33a7daa3d/codecov_cli/commands/create_report_result.py#L36

these changes should be good to go

joseph-sentry (approved):
lgtm

PR #883, opened 2024-10-11
feat(gql): Add testSuites and flags to TestAnalytics

joseph-sentry:


graphql_api/tests/test_test_analytics.py
```diff
@@ -1134,3 +1134,48 @@ def test_flake_aggregates_7_days(self) -> None:
             "flakeCountPercentChange": -50.0,
             "flakeRatePercentChange": -43.75,
         }
+
+    def test_test_suites(self) -> None:
```
ajay-sentry, line 1138: can we add 2 more tests where there's no filter? To hit the elses

ajay-sentry (approved):
nice!

PR #884, opened 2024-10-14
feat: Update with email opt in

RulaKhaled:
### Purpose/Motivation
Needed for marketing, setting marketing consent during TOS

### Legal Boilerplate


codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -21,14 +21,11 @@ class SaveTermsAgreementInteractor(BaseInteractor):
     requires_service = False
 
     def validate(self, input: TermsAgreementInput):
-        if input.terms_agreement is None:
```
ajay-sentry, line 24: Per Rola, condition is never used so this is dead removal

ajay-sentry (approved):
lgtm ✅

PR #892, opened 2024-10-16
Add logic for BA endpoint to accept storage_path

adrian-codecov:
Shelter will pass the storage_path variable that it creates on its end. We want to acknowledge if a request comes from shelter, and if so use it's storage_path

### Legal Boilerplate

### DIFF
--- a/upload/views/bundle_analysis.py
+++ b/upload/views/bundle_analysis.py
@@ -116,12 +116,15 @@ class BundleAnalysisView(APIView, ShelterMixin):
             },
         )
 
-        upload_external_id = str(uuid.uuid4())
-        storage_path = StoragePaths.upload.path(upload_key=upload_external_id)
-        archive_service = ArchiveService(repo)
-        url = archive_service.storage.create_presigned_put(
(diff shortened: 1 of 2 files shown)


upload/views/bundle_analysis.py
```diff
@@ -116,12 +117,15 @@ def post(self, request: HttpRequest) -> Response:
...
-        upload_external_id = str(uuid.uuid4())
-        storage_path = StoragePaths.upload.path(upload_key=upload_external_id)
-        archive_service = ArchiveService(repo)
-        url = archive_service.storage.create_presigned_put(
-            get_bucket_name(), storage_path, 30
-        )
+        storage_path = data.get("storage_path", None)
+        url = None
+        if storage_path is None or not self.is_shelter_request():
+            upload_external_id = str(uuid.uuid4())
```
JerrySentry, line 123: should pull this var out as it is used in the `reportid` task arguments later on

JerrySentry (approved):

nora-codecov (approved):

PR #894, opened 2024-10-16
Stripe Auto-refund cancellation within Grace Period

calvin-codecov:
### Purpose/Motivation
Adding auto-refund logic for when a user cancels their subscription within 24 hours (monthly charge) or 72 hours (yearly charge) of the period starting and being charged.

Open to any and all feedback including styling and syntax recommendations as this is one of my first python/codecov-api PRs. Should I wrap Stripe API calls in a try/catch?

### Links to relevant tickets
Closes https://github.com/codecov/engineering-team/issues/2508
- Follows ideas at https://github.com/codecov/engineering-team/issues/1813#issuecomment-2290200974
- Follows https://docs.stripe.com/billing/subscriptions/cancel with https://docs.stripe.com/api/subscriptions/cancel
- Last line of https://docs.stripe.com/billing/subscriptions/cancel#prorate-for-usage-based-billing
  - To [refund your customer](https://support.stripe.com/questions/refunding-credit-balance-to-customer-after-subscription-downgrade-or-cancellation), issue [refunds](https://docs.stripe.com/refunds#issuing) and then [adjust their account balance](https://docs.stripe.com/billing/customer/balance#modifying) back to zero.
- https://support.stripe.com/questions/refunding-credit-balance-to-customer-after-subscription-downgrade-or-cancellation

### What does this PR do?
Our current logic is that when someone cancels, since they have already paid for their current period (either a month or a year), we just "modify" there subscription to cancel at period end. We will still keep this logic for when they cancel out of the grace period but within the grace period, we will cancel immediately and refund them their money.

Doing now:
- grabs the current period's start timestamp and compares it to current time
- checks to see if it is within 1 day if monthly or 3 days if yearly
- if it is, check if they have exceeded their limit of 2 autorefund instances
    - we cancel the current subscription instead of modifying it
- look through the invoices on the subscription from within the last period and create refunds for each that is paid and has a created date before the start of the period
- modify the customer to have a balance of 0
- if not within the grace period or they've exceeded their limit, we modify the subscription to end at period end like we currently do

### Notable change ###
- we are managing the limit of autorefunds through stripe Customer object's metadata field `{"autorefunds_remaining: int"}` as opposed to making a new column in the DB `Owners` table
- this allows anyone with Stripe access to modify for a customer if needed
- Drawback: a bad actor could theoretically keep making new Stripe accounts to get around this and get free Codecov if they cancelled every 3 days as this limit is imposed on a stripe Customer as opposed to a Codecov owner. We will be logging each time an autorefund is fired to get a sense of how often it is happened and if users are dropping down to 0 to decide if this is a problem.

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -148,7 +148,6 @@ class StripeService(AbstractPaymentService):
         )
         return list(invoices_filtered_by_status_and_total)
 
-    # cancels a Stripe customer subscription immediately and attempts to refund their payments for the current period
     def cancel_and_refund(
         self,
         owner,
@@ -156,6 +155,7 @@ class StripeService(AbstractPaymentService):
         subscription_plan_interval,
         autorefunds_remaining,
     ):
+        # cancels a Stripe customer subscription immediately and attempts to refund their payments for the current period
         stripe.Subscription.cancel(owner.stripe_subscription_id)
 
         start_of_last_period = current_subscription_datetime - relativedelta(months=1)
@@ -180,12 +180,15 @@ class StripeService(AbstractPaymentService):
             },
         )
 
-        # we only want to refund the invoices for the latest, current period
+        # we only want to refund the invoices PAID recently for the latest, current period. "invoices_list" gives us any invoice
+        # created over the last month/year based on what period length they are on but the customer could have possibly
+        # switched from monthly to yearly recently. 
         recently_paid_invoices_list = [
             invoice
             for invoice in invoices_list["data"]
             if invoice["status_transitions"]["paid_at"] is not None
-            and invoice["status_transitions"]["paid_at"] >= int(invoice_grace_period_start.timestamp())
+            and invoice["status_transitions"]["paid_at"]
+            >= int(invoice_grace_period_start.timestamp())
         ]
 
         created_refund = False
@@ -215,7 +218,6 @@ class StripeService(AbstractPaymentService):
                 ),
             )
         else:
-            # log that we created no refunds but did cancel them
             log.info(
                 "Grace period cancelled a subscription but did not find any appropriate invoices to autorefund",
                 extra=dict(
--- a/services/billing.py
+++ b/services/billing.py
@@ -171,10 +171,22 @@ class StripeService(AbstractPaymentService):
                 "created.lt": int(current_subscription_datetime.timestamp()),
             },
         )
+        invoice_grace_period_start = (
+            current_subscription_datetime - relativedelta(days=1)
+            if subscription_plan_interval == "month"
+            else current_subscription_datetime - relativedelta(days=3)
+        )
+        # we only want to refund the invoices for the latest, current period
+        recently_paid_invoices_list = [
+            invoice
+            for invoice in invoices_list["data"]
+            if invoice["status_transitions"]["paid_at"] is not None
+            and invoice["status_transitions"]["paid_at"] >= invoice_grace_period_start
+        ]
 
         created_refund = False
         # there could be multiple invoices that need to be refunded such as if the user increased seats within the grace period
-        for invoice in invoices_list["data"]:
+        for invoice in recently_paid_invoices_list:
             # refund if the invoice has a charge and it has been fully paid
             if invoice["charge"] is not None and invoice["amount_remaining"] == 0:
                 stripe.Refund.create(invoice["charge"])
--- a/services/billing.py
+++ b/services/billing.py
@@ -182,7 +182,7 @@ class StripeService(AbstractPaymentService):
 
         # we only want to refund the invoices PAID recently for the latest, current period. "invoices_list" gives us any invoice
         # created over the last month/year based on what period length they are on but the customer could have possibly
-        # switched from monthly to yearly recently. 
+        # switched from monthly to yearly recently.
         recently_paid_invoices_list = [
             invoice
             for invoice in invoices_list["data"]
--- a/services/billing.py
+++ b/services/billing.py
@@ -148,9 +148,14 @@ class StripeService(AbstractPaymentService):
         )
         return list(invoices_filtered_by_status_and_total)
 
-    def cancel_and_refund(self, owner, current_subscription_datetime, subscription_plan_interval, autorefunds_remaining):
-        # maybe have a docstring about why the stripe operation for the customer is different
-
+    # cancels a Stripe customer subscription immediately and attempts to refund their payments for the current period
+    def cancel_and_refund(
+        self,
+        owner,
+        current_subscription_datetime,
+        subscription_plan_interval,
+        autorefunds_remaining,
+    ):
         stripe.Subscription.cancel(owner.stripe_subscription_id)
 
         start_of_last_period = (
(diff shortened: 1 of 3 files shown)


services/billing.py
```diff
@@ -155,18 +157,64 @@ def delete_subscription(self, owner: Owner):
...
                 extra=dict(ownerid=owner.ownerid),
             )
             stripe.SubscriptionSchedule.release(subscription_schedule_id)
 
-        stripe.Subscription.modify(
-            owner.stripe_subscription_id,
-            cancel_at_period_end=True,
-            proration_behavior="none",
+        # we give an auto-refund grace period of 24 hours for a monthly subscription or 72 hours for a yearly subscription
+        # current_subscription_timestamp = subscription["current_period_start"]
```
nora-codecov, line 169: did you mean to leave this?

services/billing.py
```diff
@@ -155,18 +157,64 @@ def delete_subscription(self, owner: Owner):
...
+        current_subscription_datetime = datetime.fromtimestamp(
+            subscription["current_period_start"]
         )
+        differenceFromNow = datetime.now() - current_subscription_datetime
```
nora-codecov, line 173: snake pls!

services/billing.py
```diff
@@ -155,18 +157,64 @@ def delete_subscription(self, owner: Owner):
...
+
+        subscription_plan_interval = (
+            subscription.plan.interval if subscription.plan is not None else None
+        )
+        within_refund_grace_period = (
+            subscription_plan_interval == "month" and differenceFromNow.days < 1
+        ) or (subscription_plan_interval == "year" and differenceFromNow.days < 3)
```
nora-codecov, line 180: does `.days` operate the way you expect? Did you check for weird rounding errors in how it defines a day?

services/billing.py
```diff
@@ -155,18 +157,64 @@ def delete_subscription(self, owner: Owner):
...
+                    and invoice_created_datetime < current_subscription_datetime
+                    and invoice_created_datetime >= start_of_last_period
+                ):
+                    stripe.Refund.create(invoice["charge"])
+                    created_refund = True
+            if created_refund == True:
+                stripe.Customer.modify(
+                    owner.stripe_customer_id,
+                    balance=0,
+                )
```
nora-codecov, line 210: really good job finding all the edge cases with invoicing here - are you sure this isn't something we can have stripe handle within their `Subscription.cancellation()` logic?

nora-codecov (commented):
left some comments and slacked you with a few additional thoughts 👍

services/billing.py
calvin-codecov: oops, haha habit

services/billing.py
calvin-codecov: yeah, for some descriptuon

services/billing.py
```diff
@@ -155,18 +157,64 @@ def delete_subscription(self, owner: Owner):
...
-            cancel_at_period_end=True,
-            proration_behavior="none",
+        # we give an auto-refund grace period of 24 hours for a monthly subscription or 72 hours for a yearly subscription
+        # current_subscription_timestamp = subscription["current_period_start"]
+        current_subscription_datetime = datetime.fromtimestamp(
+            subscription["current_period_start"]
         )
+        differenceFromNow = datetime.now() - current_subscription_datetime
+
+        subscription_plan_interval = (
```
ajay-sentry, line 175: This is equivalent to subscription_plan_interval = getattr(subscription.plan, "interval", None) I believe

services/billing.py
```diff
@@ -155,18 +157,64 @@ def delete_subscription(self, owner: Owner):
...
+                invoice_created_datetime = datetime.fromtimestamp(invoice["created"])
+                if (
+                    invoice["charge"] is not None
+                    and invoice["amount_remaining"] == 0
+                    and invoice_created_datetime < current_subscription_datetime
+                    and invoice_created_datetime >= start_of_last_period
+                ):
+                    stripe.Refund.create(invoice["charge"])
+                    created_refund = True
+            if created_refund == True:
```
ajay-sentry, line 206: can get rid of this secondary if (and remove created_refund) and have this clause be in the top level statement with what you have in the else as an else for the conditional starting on 198, right?

 `stripe.Customer.modify(
                    owner.stripe_customer_id,
                    balance=0,
                )`

services/billing.py
calvin-codecov: not the second line. oops

services/billing.py
calvin-codecov: .days is defined as full days so it will be an integer

services/billing.py
```diff
@@ -155,18 +157,75 @@ def delete_subscription(self, owner: Owner):
...
         )
+        within_refund_grace_period = (
+            subscription_plan_interval == "month" and difference_from_now.days < 1
+        ) or (subscription_plan_interval == "year" and difference_from_now.days < 3)
+
+        customer = stripe.Customer.retrieve(owner.stripe_customer_id)
+        # we are giving customers 2 autorefund instances
+        autorefunds_remaining = int(
+            customer["metadata"].get("autorefunds_remaining", "2")
+        )
```
nora-codecov, line 185: I want this logic to be inside the `if within_refund_grace_period` block so it's only called when they are within the cancellation window

services/billing.py
```diff
@@ -155,18 +157,75 @@ def delete_subscription(self, owner: Owner):
...
                 extra=dict(ownerid=owner.ownerid),
             )
             stripe.SubscriptionSchedule.release(subscription_schedule_id)
 
-        stripe.Subscription.modify(
-            owner.stripe_subscription_id,
-            cancel_at_period_end=True,
-            proration_behavior="none",
+        # we give an auto-refund grace period of 24 hours for a monthly subscription or 72 hours for a yearly subscription
+        current_subscription_datetime = datetime.fromtimestamp(
```
ajay-sentry, line 169: Just to confirm, are the timezones for both datetime.now() and datetime.fromtimestamp() the same?

services/billing.py
calvin-codecov: That makes sense. Any ideas about how to prevent another level of `if` statements or in your opinion, does that seem fine? Do we have a general rule of thumb for how many levels we think is acceptable?

services/billing.py
calvin-codecov: That's a great question. I just checked and both are in machine local so since we only care about the difference, we're good.

ajay-sentry: great! Thanks for checking

services/billing.py
```diff
@@ -155,18 +157,75 @@ def delete_subscription(self, owner: Owner):
...
+        customer = stripe.Customer.retrieve(owner.stripe_customer_id)
+        # we are giving customers 2 autorefund instances
+        autorefunds_remaining = int(
+            customer["metadata"].get("autorefunds_remaining", "2")
+        )
+
+        if within_refund_grace_period and autorefunds_remaining > 0:
+            stripe.Subscription.cancel(owner.stripe_subscription_id)
+
+            invoices_list = stripe.Invoice.list(
```
ajay-sentry, line 190: it looks like you can add a created parameter to this call too and maybe shorten the amount of results to filter through as well. Not sure how much of an optimization... maybe to the point of not needing this stuff:

and invoice_created_datetime < current_subscription_datetime
and invoice_created_datetime >= start_of_last_period

ajay-sentry: https://docs.stripe.com/api/invoices/list#list_invoices-created

ajay-sentry: That would allow you to this variable above and condense this for loop as well

start_of_last_period = (
                    current_subscription_datetime - relativedelta(months=1)
                    if subscription_plan_interval == "month"
                    else current_subscription_datetime - relativedelta(years=1)
                )

services/billing.py
```diff
@@ -155,18 +157,75 @@ def delete_subscription(self, owner: Owner):
...
+                ):
+                    stripe.Refund.create(invoice["charge"])
+                    created_refund = True
+            if created_refund == True:
+                stripe.Customer.modify(
+                    owner.stripe_customer_id,
+                    balance=0,
+                    metadata={"autorefunds_remaining": str(autorefunds_remaining - 1)},
+                )
+                log.info(
```
ajay-sentry, line 219: any other log info that might be useful here? Maybe customer id and sub id off the top of my head, we can add them with the extra=dict() param

services/billing.py
calvin-codecov: I'm converting these to utc now actually because I realized depending on what the local time of the machine is, it could change the date. Like if february 29th was converted to march 1st, subtracting 1 month would give february 1st instead of january 29th

services/billing.py
```diff
@@ -146,6 +148,69 @@ def list_filtered_invoices(self, owner: Owner, limit=10):
...
         return list(invoices_filtered_by_status_and_total)
 
+    # cancels a Stripe customer subscription immediately and attempts to refund their payments for the current period
+    def cancel_and_refund(
+        self,
+        owner,
+        current_subscription_datetime,
+        subscription_plan_interval,
+        autorefunds_remaining,
+    ):
```
nora-codecov, line 158:
```suggestion
    def cancel_and_refund(
        self,
        owner,
        current_subscription_datetime,
        subscription_plan_interval,
        autorefunds_remaining,
    ):
    # cancels a Stripe customer subscription immediately and attempts to refund their payments for the current period
```

services/billing.py
```diff
@@ -146,6 +148,69 @@ def list_filtered_invoices(self, owner: Owner, limit=10):
...
+        # there could be multiple invoices that need to be refunded such as if the user increased seats within the grace period
+        for invoice in invoices_list["data"]:
+            # refund if the invoice has a charge and it has been fully paid
+            if invoice["charge"] is not None and invoice["amount_remaining"] == 0:
+                stripe.Refund.create(invoice["charge"])
+                created_refund = True
+
+        if created_refund:
+            # update the customer's balance back to 0 in accordance to
+            # https://support.stripe.com/questions/refunding-credit-balance-to-customer-after-subscription-downgrade-or-cancellation
```
nora-codecov, line 185: ty for the documentation!!

services/billing.py
```diff
@@ -146,6 +148,69 @@ def list_filtered_invoices(self, owner: Owner, limit=10):
...
+                "Grace period cancelled a subscription but did not find any appropriate invoices to autorefund",
+                extra=dict(
+                    owner_id=owner.ownerid,
+                    user_id=self.requesting_user.ownerid,
+                    subscription_id=owner.stripe_subscription_id,
+                    customer_id=owner.stripe_customer_id,
+                    autorefunds_remaining=autorefunds_remaining,
+                ),
+            )
+
```
nora-codecov, line 213: great logs!

nora-codecov (approved):
left 1 suggested change - the comment should be inside the func, other than that I think it's ready to go! 🚀

ajay-sentry (approved):
very nice!

PR #895, opened 2024-10-16
Add support for saving comments in global YAML file

JerrySentry:
Add a key to the yaml JSON field to store the comments included YAML str in its original form.
Things considered/validated:
1. Does not affect DB trigger functions for the yaml column
2. Does not affect `UserYaml` object, values are fetched by specific keys
3. Does not affect `CodecovYamlValidator`, when JSON is stored into the column it is already validated, however will add a followup PR in shared to future proof to protect the key from being used in the codecov YAML schema.

### Legal Boilerplate


codecov_auth/commands/owner/interactors/set_yaml_on_owner.py
```diff
@@ -49,10 +50,12 @@ def convert_yaml_to_dict(self, yaml_input):
             raise ValidationError(message)
 
     @sync_to_async
-    def execute(self, username, yaml_input):
+    def execute(self, username: str, yaml_input: str) -> Owner:
         self.validate()
         self.owner = self.get_owner(username)
         self.authorize()
         self.owner.yaml = self.convert_yaml_to_dict(yaml_input)
```
JerrySentry, line 57: One of two of the actual changes.

services/tests/test_yaml.py
```diff
@@ -56,3 +56,15 @@ def test_when_commit_has_yaml_with_owner(self, mock_fetch_yaml):
...
+
+    @patch("services.yaml.fetch_current_yaml_from_provider_via_reference")
+    def test_when_commit_has_reserved_to_string_key(self, mock_fetch_yaml):
+        mock_fetch_yaml.return_value = """
+        codecov:
+          notify:
+            require_ci_to_pass: no
+        to_string: hello
+        """
+        config = yaml.final_commit_yaml(self.commit, self.org)
```
suejung-sentry, line 68: In case you didn't see this, I wanted to see what `final_commit_yaml` does - there is this essay about the logic 😂 
https://github.com/codecov/shared/blob/f40874907d3ddbe6993c8016c3f09306cc14a9f8/shared/yaml/user_yaml.py#L128

I think it's not a problem because your change only deals with the "ownerYaml".

I also checked what other clients than the graphql resolver for the owner yaml in this PR may be accessing the revised owner.yaml field. Searching "owner.yaml" shows nothing in shared and a couple spots in `worker` that all point back to the same `final_commit_yaml` function above. That function returns a dictionary that doesn't appear to matter if it has the extra `to_string` on it from what I can tell. So no additional work there needed from what I can tell.

codecov_auth/commands/owner/interactors/set_yaml_on_owner.py
```diff
@@ -49,10 +50,12 @@ def convert_yaml_to_dict(self, yaml_input):
             raise ValidationError(message)
```
suejung-sentry, line 50: I'd guess in a perfect world we would want `validate_yaml` (in [shared](https://github.com/codecov/shared/blob/fffdcf08be320e5feb14111d2cea4e5e8bd54ad0/shared/yaml/validation.py#L16)) to reject if any reserved keys (like `to_string`) are in it. It's probably a lot of work for something unlikely, though so can let it be

suejung-sentry (approved):
Nice! I like how simple it is! 
Probably worth end to end testing in staging to confirm as much as possible, as I could see some of the parsing ending up weird, or any edge cases around escaping. Also not sure how to test that the other clients than the graphql resolvers are happy with the change than smoke testing user flows in staging

codecov_auth/commands/owner/interactors/set_yaml_on_owner.py
suejung-sentry: Oh never mind I had missed the note in the PR description! Sounds like that will get covered later - nice!

PR #900, opened 2024-10-17
fix: Remove ordering on account organizations list

spalmurray-codecov:
Removes the ordering argument on the account organizations list because it turns out that the paginator can't handle/isn't intended to do sorting by count of a field. Was throwing errors.

Additionally, adds a new resolver decorator to allow users of an account to see the activatedUserCount of orgs in that same account. This was a necessary change to support the desired UI, otherwise we'd only show activated user counts for orgs that the user is a member of.

### DIFF
--- a/graphql_api/helpers/mutation.py
+++ b/graphql_api/helpers/mutation.py
@@ -78,8 +78,6 @@ def require_part_of_org(resolver):
 def require_shared_account_or_part_of_org(resolver):
     def authenticated_resolver(queried_owner, info, *args, **kwargs):
         current_user = info.context["request"].user
-        current_owner = info.context["request"].current_owner
-
         if (
             current_user
             and current_user.is_authenticated
@@ -89,15 +87,7 @@ def require_shared_account_or_part_of_org(resolver):
         ):
             return resolver(queried_owner, info, *args, **kwargs)
 
-        if (
-            not current_user
-            or not current_user.is_authenticated
-            or not current_owner
-            or not current_user_part_of_org(current_owner, queried_owner)
-        ):
-            return None
-
-        return resolver(queried_owner, info, *args, **kwargs)
+        return require_part_of_org(resolver)(queried_owner, info, *args, **kwargs)
 
     return authenticated_resolver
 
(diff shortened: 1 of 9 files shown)


graphql_api/tests/test_account.py
```diff
@@ -318,7 +226,7 @@ def test_fetch_organizations_pagination(self) -> None:
             for node in result["owner"]["account"]["organizations"]["edges"]
         ]
 
-        assert orgs == ["owner-2", "owner-1"]
```
ajay-sentry, line 321: how come this changed?

nora-codecov: I think Spencer reversed the ordering from desc to asc and it's limited to `first: 2`

graphql_api/helpers/mutation.py
```diff
@@ -75,5 +75,22 @@ def authenticated_resolver(queried_owner, info, *args, **kwargs):
...
 
+def require_shared_account_or_part_of_org(resolver):
+    def authenticated_resolver(queried_owner, info, *args, **kwargs):
+        current_user = info.context["request"].user
+        if (
+            current_user
+            and current_user.is_authenticated
+            and queried_owner
+            and queried_owner.account
+            and current_user in queried_owner.account.users.all()
```
nora-codecov, line 86: 💯 💯 💯  
LMK if this is being slow, can optimize with 
- a prefetch for the Account and Users objs when you hit the db for the Owner obj
- doing `current_user in queried_owner.account.users.all()` in a different way

nora-codecov: `queried_owner` is the OwnerOrg that the `current_user` is checking permissions for right? Not the `current_user`'s OwnerUser object 🫠

nora-codecov (approved):

ajay-sentry (approved):
send it!

graphql_api/tests/test_account.py
spalmurray-codecov: yep that

graphql_api/helpers/mutation.py
spalmurray-codecov: @nora-codecov Yes. Specifically this decorator should allow users to see fields on owners they share an account with. For example, I'm in codecov org and not sentry org. Since we share an Account, this decorator will let me see certain things on Sentry that an unrelated user could not see.

PR #904, opened 2024-10-17
feat: Use env var for GQL rate limit instead of hard code val

ajay-sentry:
### Purpose/Motivation

This PR aims to move the hard coded gql rate limit value to an env variable that can be set per environment

In the same PR I've updated the staging limit to be 1000 from 300

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/codecov/settings_base.py
+++ b/codecov/settings_base.py
@@ -92,12 +92,19 @@ TEMPLATES = [
 
 WSGI_APPLICATION = "codecov.wsgi.application"
 
-# Database
-# https://docs.djangoproject.com/en/2.1/ref/settings/#databases
+
+# GraphQL
 
 GRAPHQL_QUERY_COST_THRESHOLD = get_config(
     "setup", "graphql", "query_cost_threshold", default=10000
 )
+
+GRAPHQL_RATE_LIMIT_VALUE = get_config(
+    "setup", "graphql", "rate_limit_value", default=300
+)
+
+# Database
+# https://docs.djangoproject.com/en/2.1/ref/settings/#databases
 
 DATABASE_ROUTERS = ["codecov.db.DatabaseRouter"]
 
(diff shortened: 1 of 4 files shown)


graphql_api/views.py
```diff
@@ -320,7 +320,7 @@ def _check_ratelimit(self, request):
             user_ip = self.get_client_ip(request)
             key = f"rl-ip:{user_ip}"
 
-        limit = 300
+        limit = settings.GRAPHQL_RATE_LIMIT_VALUE
         window = 60  # in seconds
```
suejung-sentry, line 324: May be nice to have an `GRAPHQL_RATE_LIMIT_ENABLED` knob as well if not too annoying to add in. May be easier for DEC/self-hosted to use than trying to guess a large number for their use case :shrug:

codecov/settings_base.py
```diff
@@ -92,13 +92,20 @@
...
-# Database
-# https://docs.djangoproject.com/en/2.1/ref/settings/#databases
+
+# GraphQL
 
 GRAPHQL_QUERY_COST_THRESHOLD = get_config(
     "setup", "graphql", "query_cost_threshold", default=10000
 )
 
+GRAPHQL_RATE_LIMIT_VALUE = get_config(
```
suejung-sentry, line 102: Thoughts on a name that includes the units? Like GRAPHQL_RATE_LIMIT_RPM or similar?

suejung-sentry (approved):
approved! left optional comments

codecov/settings_base.py
ajay-sentry: Oooo I like that, I was straining my brain thinking of a name besides "Value" lol

graphql_api/views.py
ajay-sentry: Fair point! I like it

suejung-sentry (approved):
nice!!

PR #909, opened 2024-10-21
fix: Improve branch search speed

suejung-sentry:
We run branch search by name on the Configuration page of a repo - it's currently slow (~50 seconds in worst case, median ~10+seconds - Sentry [board](https://codecov.sentry.io/issues/5001374062/events/?project=5215654&selected_replay_index=4&sort=-transaction.duration)). 

The branch name search is the `GetBranches` graphQL query and the below db query. 
```
SELECT branches.branch, branches.repoid, branches.authors, branches.head, branches.base,
  branches.updatestamp
FROM branches
WHERE (
  branches.repoid = 111 AND UPPER(branches.branch::text) LIKE UPPER('%codecov%')
)
ORDER BY branches.updatestamp DESC
LIMIT 21;
```

I ran some tests against postgres prod and the `EXPLAIN ANALYZE` shows significant difference between running `ILIKE` vs. `LIKE` for the search (~10 ms vs. 5 seconds). 
There is an index on `branches.repoid` and one on `branches.repoid, branches.branch`. From the query plans it looks like running with `LIKE` is running a custom columnar scan (full table scan on all 27 million rows of `branches` table) before then using the repoid index and filtering those rows by name. The query with `ILIKE` just uses the index then filters, so it's much faster.
explain analyze with `LIKE`:
Screenshot 2024-10-21 at 1 09 29 PM
explain analyze with `ILIKE`:
Screenshot 2024-10-21 at 1 08 10 PM

So this PR moves over to using `ILIKE` in the query. I suspect the reason for the difference between running with LIKE vs. ILIKE has to do with our collation settings for the table but those are risky to alter so I'd rather move to using ILIKE than trying to fix LIKE.

Other stuff:
* Django's docs say the existing `icontains` methods translates to a query of `ILIKE`, but it clearly doesn't. Some troubleshooting says maybe it's because our DATABASE ENGINE settings.py is not set to postgres, but everything I checked seems to imply that is done already.
* [This from the Django forums](https://forum.djangoproject.com/t/case-insensitive-pattern-lookups-icontains-istartswith-and-iendswith-in-postgres-dont-use-ilike-and-thereby-miss-out-on-potential-performance-boosts-from-gin-indexes/35577) seems to confirm their doc is just wrong. The Django folks recommend redefining the ORM hooks, but I gave it a shot and didn't do what I expected. Also seems risky to mess with it at that level.
* So the alternative I went with here is the `.extra` api that gets us what we need. Technically the docs say to use it as a last resort, but it seems to be the best solution for this now without restructuring our ORM in bigger ways.
* On the frontend we have a debounce (500 ms) for the query already, so unfortunately no further gains that way

Closes https://github.com/codecov/engineering-team/issues/2537


suejung-sentry:
Very sad but I tried deploying [this](https://github.com/codecov/codecov-api/pull/921/files) to staging but the ILike field register lookup didn't get picked up.
Screenshot 2024-10-22 at 5 37 23 PM

`FetchRepoBranchesInteractor with ilike SELECT "branches"."branch", "branches"."repoid", "branches"."authors", "branches"."head", "branches"."base", "branches"."updatestamp" FROM "branches" WHERE ("branches"."repoid" = 1262 AND UPPER("branches"."branch"::text) LIKE UPPER(%asasas%))`

ajay-sentry (approved):
lgtm!

PR #910, opened 2024-10-22
fix: Restrict deactivated enterprise user access

suejung-sentry:
Restrict what deactivated Enterprise users can see when the environment has guest access turned off.
- If guest access if off, deactivated users should not be able to see anything
- If guest access is on, deactivated users should be able to see public repos, but not private repos

Closes https://github.com/codecov/engineering-team/issues/1859


graphql_api/types/query/query.py
```diff
@@ -50,11 +52,17 @@ def resolve_owner(_, info, username):
         if not user or not user.is_authenticated:
             raise UnauthorizedGuestAccess()
 
-    return get_owner(service, username)
+        # per product spec, if guestAccess is off for the environment, the current enterpriseUser
+        # must be "activated" in the given target owner (e.g., "codecov" org) in order to see things
+        target = await get_owner(service, username)
+        if user.ownerid not in target.plan_activated_users:
```
JerrySentry, line 58: you can use this helper function: `current_user_part_of_org`

suejung-sentry: Oh nice thanks!
Looking over [that helper](https://github.com/codecov/codecov-api/blob/68ade431abc69558315b571aced42974d5f0f38d/codecov_auth/helpers.py#L27) - Are we diligent about updating an owner (say `suejung-sentry`)'s `organizations` when we adjust the other source of truth on the organization (say `codecov`)'s `plan_activated_users`?
From what I could tell the term `activated` relates to whether the ownerid in the `plan_activated_users` list, and I haven't dug yet into what the owner.organizations is, but I thought it had to do with `accounts`?

suejung-sentry: sounds like the helper may not be equivalent or what's desired here in this chat

Screenshot 2024-10-23 at 8 40 00 AM

graphql_api/types/query/query.py
```diff
@@ -50,11 +52,17 @@ def resolve_owner(_, info, username):
         if not user or not user.is_authenticated:
             raise UnauthorizedGuestAccess()
 
-    return get_owner(service, username)
+        # per product spec, if guestAccess is off for the environment, the current enterpriseUser
```
ajay-sentry, line 55: nit: I think the comment states something that can be inferred from the code / PR if needed so not sure if there's much value in having it tbh

ajay-sentry (approved):
lgtm!

suejung-sentry:
Tested in local different scenarios with Enterprise + guest access. Also smoke-tested in staging that no impact on the cloud deployment. If guest access for the deployment is on, no change to the existing repo-level check of whether the user has permission to see the repo. If guest access is off, must be a planActivatedUser to do anything

PR #911, opened 2024-10-22
Remove usage of `ReportDetails`/`files_array`

Swatinem:
This removes all usage of `ReportDetails`. The base definitions still exist.

### DIFF
--- a/graphql_api/actions/commits.py
+++ b/graphql_api/actions/commits.py
@@ -80,13 +80,12 @@ def commit_status(
 def repo_commits(
     repository: Repository, filters: Optional[dict] = None
 ) -> QuerySet[Commit]:
-    # prefetch the CommitReport with the ReportLevelTotals and ReportDetails
+    # prefetch the CommitReport with the ReportLevelTotals
     prefetch = Prefetch(
         "reports",
         queryset=CommitReport.objects.coverage_reports()
         .filter(code=None)
-        .select_related("reportleveltotals", "reportdetails")
-        .defer("reportdetails___files_array"),
+        .select_related("reportleveltotals"),
     )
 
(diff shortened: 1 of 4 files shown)


nicholas-codecov (approved):

giovanni-guidini (approved):

trent-codecov (approved):

suejung-sentry (approved):
I double checked and didn't see any impact of removing the `ReportDetails`, `files_array` usages as proposed here on the GraphQL apis and public REST apis. 
Honestly, I'm puzzled why this was here in the first place? Did some digging in the history but didn't turn up anything obvious (my best guess is at some point we moved over from reading the results of the coverage reports in `reports_reportdetails.files_array` and moved to a path at GCS at `commits.report_storage_path`)?

All the places our frontend app [fetches](https://github.com/codecov/codecov-api/blob/90e7eb8a6432fd7b90130c9b3d2ddf6056336058/services/comparison.py#L975) any files data, it appears to [download](https://github.com/codecov/codecov-api/blob/90e7eb8a6432fd7b90130c9b3d2ddf6056336058/services/comparison.py#L1004) from the GCS path at `commits.report_storage_path`, so that's good, unaffected by this change.

I also smoke-tested clicked around the deployment at https://stage-app.codecov.dev/gh/codecov and didn't see anything out of place. So I think we should be good to go with this!

(disclaimer: I only lightly educated myself on the `needs_recompute` logic that was edited in this PR but seems generally reasonable!)

Swatinem:
Yes that was it, the  `ReportDetails.files_array` was intended as a feature flagged replacement for the `Commit.report_json` (which is stored in GCS).

The feature flag never made it, and I wanted to be extra careful rolling back all the changes.

PR #913, opened 2024-10-22
fix: parameter filtering takes branch into account

joseph-sentry:


utils/test_results.py
```diff
@@ -359,16 +371,21 @@ def generate_test_results(
...
         match parameter:
             case GENERATE_TEST_RESULT_PARAM.FLAKY:
-                flakes = Flake.objects.filter(
-                    Q(repository_id=repoid)
-                    & (Q(end_date__date__isnull=True) | Q(end_date__date__gt=since))
-                )
+                totals = get_relevant_totals(repoid, branch, since)
 
-                flaky_test_ids = set([flake.test_id for flake in flakes])
+                flaky_test_ids = (
```
ajay-sentry, line 376: being able to have a generalized pattern for all these makes it so much easier to read

utils/test_results.py
```diff
@@ -359,16 +371,21 @@ def generate_test_results(
     if parameter is not None:
         match parameter:
             case GENERATE_TEST_RESULT_PARAM.FLAKY:
-                flakes = Flake.objects.filter(
-                    Q(repository_id=repoid)
-                    & (Q(end_date__date__isnull=True) | Q(end_date__date__gt=since))
-                )
+                totals = get_relevant_totals(repoid, branch, since)
```
ajay-sentry, line 374: Reading it this way now, we could also bump totals up to line 372 and use it for each case now right?

Vs. having it be the "starting line" in each of our case statements below

ajay-sentry (approved):
couple comments, nothing blocking

PR #914, opened 2024-10-22
bundle analysis: fix unknown asset size compute

JerrySentry:
If looking to get unknown asset type in the filters only retrieve the necessary known asset types to compute it
### Legal Boilerplate

### DIFF
--- a/graphql_api/types/bundle_analysis/base.py
+++ b/graphql_api/types/bundle_analysis/base.py
@@ -295,9 +295,17 @@ def resolve_bundle_report_measurements(
     )
 
     # All measureable names we need to fetch to compute the requested asset types
-    if not asset_types or ASSET_TYPE_UNKNOWN in asset_types:
+    if not asset_types:
         measurables_to_fetch = [
             item for item in list(BundleAnalysisMeasurementsAssetType)
+        ]
+    elif ASSET_TYPE_UNKNOWN in asset_types:
+        measurables_to_fetch = [
+            BundleAnalysisMeasurementsAssetType.REPORT_SIZE,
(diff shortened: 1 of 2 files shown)


graphql_api/types/bundle_analysis/base.py
```diff
@@ -295,10 +295,18 @@ def resolve_bundle_report_measurements(
...
 
     # All measureable names we need to fetch to compute the requested asset types
-    if not asset_types or ASSET_TYPE_UNKNOWN in asset_types:
+    if not asset_types:
         measurables_to_fetch = [
             item for item in list(BundleAnalysisMeasurementsAssetType)
         ]
+    elif ASSET_TYPE_UNKNOWN in asset_types:
+        measurables_to_fetch = [
+            BundleAnalysisMeasurementsAssetType.REPORT_SIZE,
```
ajay-sentry, line 304: I saw ASSET_SIZE is another variable on this type; jw why its not included in this list?

JerrySentry: ASSET_SIZE is for tracking the size of individual javascript assets, so the sum of all asset measurements equals to the JAVASCRIPT_SIZE. And all the other types are aggregate types, notably REPORT_SIZE is the sum of all assets. 
So we can compute the trends line of unknown asset sizes by doing a computation of:
`REPORT_SIZE = JS_SIZE - STYLESHEET_SIZE - FONT_SIZE - IMAGE_SIZE`
and
ASSET_SIZE can be ignored because `JS_SIZE = sum(ASSET_SIZE)`

ajay-sentry: ahhhh okay makes a lot of sense! Thanks

ajay-sentry (approved):

PR #915, opened 2024-10-22
build: update junit xml generation

joseph-sentry:


Swatinem (approved):
rubberstamping.
can you link to documentation that explains what this flag does?

joseph-sentry:
@Swatinem https://docs.pytest.org/en/stable/reference/reference.html#confval-junit_family the format of the outputted JUnit XML is slightly different, notably, the filename attribute is included on testcases which means we get the file name where each test is found which is necessary for our `computed name` feature.

Swatinem:
It feels a bit wrong to *downgrade* in order to get more complete output.
Is it possible to solve this in a different way somehow? It would be really weird if support would need to explain to customers they have to downgrade the exported format version :-D

PR #923, opened 2024-10-24
[feat] Hide Codecov tokens from non admins if YAML setting is enabled

rohitvinnakota-codecov:
### Purpose/Motivation

Relies on https://github.com/codecov/shared/pull/397

Parent issue: https://github.com/codecov/internal-issues/issues/491

Also bumps shared. 

Screenshot 2024-10-24 at 11 49 03 AM


graphql_api/types/repository/repository.py
```diff
@@ -24,6 +25,8 @@
 from services.profiling import CriticalFile, ProfilingSummary
 from services.redis_configuration import get_redis_connection
 
+TOKEN_UNAVAILABLE = "Token Unavailable"
```
spalmurray-codecov, line 28: Is the plan for UI just to show this instead of the token? Or check the value and do something special?

rohitvinnakota-codecov: Just updated the PR and copy. Since this work is just for a single customer, we are just going to render the value directly.

spalmurray-codecov: Cool sounds good. I prefer this approach given this is not an official feature.

spalmurray-codecov (approved):

graphql_api/types/repository/repository.py
```diff
@@ -68,6 +71,13 @@ def resolve_commit(repository: Repository, info: GraphQLResolveInfo, id):
 
 @repository_bindable.field("uploadToken")
 def resolve_upload_token(repository: Repository, info: GraphQLResolveInfo):
+    should_hide_tokens = settings.HIDE_ALL_CODECOV_TOKENS
+
+    current_owner = info.context["request"].current_owner
+    is_current_user_admin = current_owner.is_admin(repository.author)
+
+    if should_hide_tokens and not is_current_user_admin:
+        return TOKEN_UNAVAILABLE
```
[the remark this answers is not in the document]
rohitvinnakota-codecov: wow good point

PR #940, opened 2024-10-29
add signal for Owner

nora-codecov:
### Purpose/Motivation
`Shelter` needs to know some things about `Owner`.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2299

Goes with this piece https://github.com/codecov/shelter/pull/203
Depends on https://github.com/codecov/shared/pull/411

### What does this PR do?
This new signal will trigger `Shelter` to pick up the object and store the relevant fields.

### DIFF
--- a/codecov_auth/signals.py
+++ b/codecov_auth/signals.py
@@ -1,4 +1,5 @@
 import json
+import logging
 
 from django.conf import settings
 from django.db.models.signals import post_save
@@ -6,6 +7,8 @@ from django.dispatch import receiver
 from google.cloud import pubsub_v1
 
 from codecov_auth.models import OrganizationLevelToken, Owner, OwnerProfile
+
+log = logging.getLogger(__name__)
 
 
 @receiver(post_save, sender=Owner)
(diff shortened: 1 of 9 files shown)


codecov_auth/signals.py
```diff
@@ -45,3 +48,38 @@ def update_org_token(sender, instance: OrganizationLevelToken, **kwargs):
...
+    Shelter tracks a limited set of Owner fields - only update if those fields have changed.
+    """
+    created = kwargs["created"]
+    tracked_fields = [
+        "upload_token_required_for_public_repos",
+        "username",
+        "service",
+    ]
+    if created or any(instance.tracker.has_changed(field) for field in tracked_fields):
+        pubsub_project_id = settings.SHELTER_PUBSUB_PROJECT_ID
```
adrian-codecov, line 65: Optional: since this is used in a couple of places, would encourage to move this into its own fn and use that in every signal for readability!

adrian-codecov (approved):

codecov_auth/signals.py
nora-codecov: I agree - I added one, check it out!

codecov_auth/signals.py
```diff
@@ -7,41 +9,86 @@
...
+        # topic_id has REPO in the name but it is used for all types of objects
+        topic_id: str = settings.SHELTER_PUBSUB_SYNC_REPO_TOPIC_ID
+        self.topic_path = self.pubsub_publisher.topic_path(pubsub_project_id, topic_id)
 
-def _get_pubsub_publisher():
-    global _pubsub_publisher
-    if not _pubsub_publisher:
-        _pubsub_publisher = pubsub_v1.PublisherClient()
-    return _pubsub_publisher
+    def publish(self, data: Dict[str, Any]) -> None:
```
adrian-codecov, line 46: Nice! This looks great! A last nitpick, could we move this to somewhere like `utils/shelter.py` in the root folder, or anywhere you see fit? This isn't strictly a codecov_auth responsibility so it makes sense it is its own standalone thing

adrian-codecov (approved):

codecov_auth/signals.py
nora-codecov: agree! moved!

PR #952, opened 2024-11-01
feat: use cached results from GCS in TA GQL

joseph-sentry:
we want the GQL endpoints to consume the cached query results from GCS
and then do some filtering and ordering on top of that, and also do some
extra caching of those results in redis.

this commit also contains a bunch of reorganization and refactoring of
the GQL code


graphql_api/types/flake_aggregates/flake_aggregates.py
```diff
@@ -1,37 +1,88 @@
...
 
 
-class FlakeAggregate(TypedDict):
+@dataclass
+class FlakeAggregates:
     flake_count: int
-    flake_count_percent_change: float | None
     flake_rate: float
-    flake_rate_percent_change: float | None
+    flake_count_percent_change: float | None = None
```
ajay-sentry, line 15: dang you didn't like having it grouped huh

graphql_api/types/flake_aggregates/flake_aggregates.py
```diff
@@ -1,37 +1,88 @@
...
+    flake_rate_percent_change: float | None = None
+
+
+def calculate_flake_aggregates(table: pl.DataFrame) -> pl.DataFrame:
```
ajay-sentry, line 19: is it possible to know what columns exist on the table at this point? Right now it's a little "magic"

graphql_api/types/test_results_aggregates/test_results_aggregates.py
```diff
@@ -1,81 +1,153 @@
...
+                * (pl.col("total_pass_count") + pl.col("total_fail_count"))
+            )
+            .otherwise(0)
+            .top_k(100)
+            .sum()
+            .alias("slowest_tests_duration")
+        ),
+        (pl.col("total_skip_count").sum()).alias("skips"),
+        (pl.col("total_fail_count").sum()).alias("fails"),
+        ((pl.col("avg_duration") >= pl.col("avg_duration").quantile(0.95)).sum()).alias(
```
ajay-sentry, line 46: do we still need to do the limit 100 here?

graphql_api/types/test_results_aggregates/test_results_aggregates.py
```diff
@@ -1,81 +1,153 @@
...
+def test_results_aggregates_with_percentage(
+    curr_results: pl.DataFrame,
+    past_results: pl.DataFrame,
+) -> TestResultsAggregates:
+    curr_aggregates = calculate_aggregates(curr_results)
+    past_aggregates = calculate_aggregates(past_results)
+
+    merged_results: pl.DataFrame = pl.concat([past_aggregates, curr_aggregates])
+
+    merged_results = merged_results.with_columns(
```
ajay-sentry, line 68: It was a bit confusing for me at first to see this since I didn't realize with_columns is essentially an append or upsert depending on the column existing

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -1,24 +1,318 @@
...
+def validate(
+    interval: int,
+    ordering: TestResultsOrderingParameter,
+    ordering_direction: OrderingDirection,
+    after: str | None,
+    before: str | None,
+    first: int | None,
+    last: int | None,
+) -> None:
+    if interval not in {1, 7, 30}:
```
ajay-sentry, line 112: nit: do we have these magic numbers typed out as consts somewhere?

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -1,24 +1,318 @@
...
+
+def encode_cursor(row: TestResultsRow, ordering: TestResultsOrderingParameter) -> str:
+    return b64encode(
+        DELIMITER.join([str(getattr(row, ordering.value)), str(row.name)]).encode(
+            "utf-8"
+        )
+    ).decode("ascii")
+
+
+def validate(
```
ajay-sentry, line 103: Random thought here, does the order of these validations matter? Do we want it to matter?

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -1,24 +1,318 @@
...
+        table = table.filter(pl.col("name").str.starts_with(term))
+
+    if testsuites:
+        table = table.filter(
+            pl.col("testsuite").is_not_null() & pl.col("testsuite").is_in(testsuites)
+        )
+
+    if flags:
+        table = table.filter(
+            pl.col("flags").list.eval(pl.element().is_in(flags)).list.any()
```
ajay-sentry, line 185: Claude was telling me we might be able to add these guards to this check, not sure if you think we need em tho

pl.col("flags").is_not_null() &  # handle null rows
pl.col("flags").list.len() > 0 &  # handle empty lists

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -1,24 +1,318 @@
...
+            )
+        case TestResultsFilterParameter.SLOWEST_TESTS:
+            table = table.filter(
+                pl.col("avg_duration") >= pl.col("avg_duration").quantile(0.95)
+            ).top_k(100, by=pl.col("avg_duration"))
+
+    total_count = table.height
+
+    def ordering_expression(cursor_value: CursorValue, is_forward: bool) -> pl.Expr:
+        if is_forward:
```
ajay-sentry, line 205: what does forward imply here? It's also a little confusing that we use the value of is_forward and then change the value of is_forward in the same fn

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -1,24 +1,318 @@
...
+    if first:
+        page_elements = table.slice(0, first)
+    elif last:
+        page_elements = table.reverse().slice(0, last)
+    else:
+        page_elements = table
+
+    rows = [TestResultsRow(**row) for row in page_elements.rows(named=True)]
+
+    page: list[dict[str, str | TestResultsRow]] = [
```
ajay-sentry, line 254: this should be pages right? More than one

graphql_api/types/test_analytics/test_analytics.py
```diff
@@ -71,10 +360,10 @@ async def resolve_test_results_aggregates(
     repository: Repository,
     info: GraphQLResolveInfo,
     interval: MeasurementInterval | None = None,
-    **_,
-):
+    **_: Any,
+) -> TestResultsAggregates | None:
     return await sync_to_async(generate_test_results_aggregates)(
-        repoid=repository.repoid, interval=convert_interval_to_timedelta(interval)
+        repoid=repository.repoid, interval=interval.value if interval else 30
```
ajay-sentry, line 366: these 30s can be MeasurementInterval.INTERVAL_30_DAY right?

utils/test_results.py
```diff
@@ -1,682 +1,70 @@
...
-		from base_cte
-		group by test_id
-	) as latest_rollups
-    on base_cte.created_at = latest_rollups.created_at
-)
+    branch: str,
+    interval_start: int,
+    interval_end: int | None = None,
+) -> str:
+    key = f"test_results:{repoid}:{branch}:{interval_start}"
```
ajay-sentry, line 15: Since this is an int... does this have a reasonable enough level of precision so we actually have some cache hits sometimes?

Like if the timestamp is 1231232132 vs. 1231232135 it seems like that would trigger a new rollup to be cached/created

graphql_api/types/flake_aggregates/flake_aggregates.py
joseph-sentry: all args with defaults must come after arguments with no defaults 😭

graphql_api/types/test_results_aggregates/test_results_aggregates.py
joseph-sentry: indeed, that's being handled above but not here

graphql_api/types/test_results_aggregates/test_results_aggregates.py
joseph-sentry: i will leave a comment

utils/test_results.py
joseph-sentry: there's no timestamp being use in the key, it's just the range of days that the cached values apply to, for example 1 day, 7 days, or 30 days

graphql_api/types/test_analytics/test_analytics.py
joseph-sentry: nope, and i don't think it should matter (maybe in past iterations it would've(

graphql_api/types/test_analytics/test_analytics.py
joseph-sentry: i think the is not null check makes sense

graphql_api/types/test_analytics/test_analytics.py
joseph-sentry: is_forward is just about determining the ordering of the table, but `ordering_expression` only reads is_forward, it's the outer function that sets the value and invokes `ordering_expression` so i ended up just moving it out

graphql_api/types/test_analytics/test_analytics.py
joseph-sentry: no, this is a single page, since it's after all the ordering and filtering

ajay-sentry (approved):
Approved, reminder to update the bool(after) thing and add a comment on the top_k() thing

PR #955, opened 2024-11-01
fix: Add graphql max depth and aliases limits

suejung-sentry:
Add protections against Denial of Service attacks on the GraphQL API that use high depth, high breadth, or high aliases.

- For high depth attacks (highly nested GraphQL queries) - I added a new validation_rule to ariadne that rejects past a max setting. There were no off-the-shelf with ariadne so this is a custom one created mimicking the one from the Apollo GraphQL SDK linked by the pentesters.
- For high breadth attacks (a ton of fields request for a given level) - we don't really offer in our schema things that wide, so the protection against aliases should cover this issue
- For high aliases attacks (requesting the same field over and over using aliases) - I added a new validation_rule counting these and rejecting if beyond a max setting. See same note on high depth attacks.

Note that we also already have cost validation by ariadne. The additional rules in this PR can catch cases where the cost validation is correct against our day-to-day use cases, but may be too lenient against crafted "attacks", such as the ones composed by the pentesters.

We also already have rate limiting on the GraphQL endpoint so that pentest recommendation is already covered.

Closes https://github.com/codecov/internal-issues/issues/918
Closes https://github.com/codecov/internal-issues/issues/917


graphql_api/validation.py
```diff
@@ -0,0 +1,65 @@
...
+
+def create_max_depth_rule(max_depth: int) -> Type[ValidationRule]:
+    class MaxDepthRule(ValidationRule):
+        def __init__(self, context: ValidationContext) -> None:
+            super().__init__(context)
+            self.operation_depth: int = 1
+            self.max_depth_reached: bool = False
+            self.max_depth: int = max_depth
+
+        def enter_operation_definition(
```
ajay-sentry, line 16: just curious, are these functions base functions you had to override default behavior of?

Similar story with enter_field, leave_field, and enter_document

codecov/settings_base.py
```diff
@@ -106,6 +106,10 @@
 
 GRAPHQL_INTROSPECTION_ENABLED = False
 
+GRAPHQL_MAX_DEPTH = 15
```
ajay-sentry, line 109: Wanted to hear a little more about landing on 15 for both of these values, do you think stage/production/dev should have the same value for each?

graphql_api/views.py
```diff
@@ -201,7 +202,11 @@ def get_validation_rules(
                 maximum_cost=settings.GRAPHQL_QUERY_COST_THRESHOLD,
                 default_cost=1,
                 variables=data.get("variables"),
-            )
+            ),
+            create_max_depth_rule(max_depth=getattr(settings, "GRAPHQL_MAX_DEPTH", 15)),
```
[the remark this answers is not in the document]
suejung-sentry: Yeah, my intention was to force defensive programming there as I feel like I've seen a lot of accessing of dicts and tribal knowledge on when it's a bug.
But thinking about it again, it does seem like burying a default at this level is the wrong practice and since settings is something we control, we would prefer it to fail early and actually throw an error instead of having it "be forgiving" here.
Fixed it!

graphql_api/validation.py
suejung-sentry: Yup! Though I don't think about it so much as "override" as "implement the interface" that is defined [here](https://github.com/graphql-python/graphql-core/blob/46244446c66ca3033f5295f6c796108e4c1f7365/src/graphql/language/visitor.py#L71). The function names `enter_X` and `leave_X` are dynamic per [here](https://github.com/graphql-python/graphql-core/blob/46244446c66ca3033f5295f6c796108e4c1f7365/src/graphql/language/visitor.py#L147C12-L150C60). And for any method that's not implemented explicitly, it behaves as a no-op ([here](https://github.com/graphql-python/graphql-core/blob/46244446c66ca3033f5295f6c796108e4c1f7365/src/graphql/language/visitor.py#L280C13-L281C30)).

this is the stuff I looked at in Ariadne [doc](https://ariadnegraphql.org/docs/query-validators#implementing-custom-validator) & this [example implementation](https://github.com/mirumee/ariadne/blob/af244283987ef57c186a512a1186f819f4dd6e20/ariadne/validation/query_cost.py#L44)

codecov/settings_base.py
suejung-sentry: I skimmed the existing queries and the max depth was around ~10, so I thought it would be a good buffer. Could potentially do like 20 if we want more wiggle room. I think keeping it the same in all envs makes sense since honestly I think it's just a set-and-forget kind of thing. Also, we are the main consumers of the graphql api so I'd want to catch any queries that would error in prod to be caught first in the lower envs at the same settings.

For aliases, I don't think we ever use more than a handful (<5?) at a time, so thought a reasonable use case wouldn't go beyond say 15 anyway (and anything above that seems to be a malicious actor).

Open to thoughts on those. If any become a problem, it's as simple as bumping this up & we would catch the issue during development of some new feature anyway

graphql_api/validation.py
ajay-sentry: Very cool! Thanks for linking all those docs, that was a fun set of reads

graphql_api/views.py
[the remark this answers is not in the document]
ajay-sentry: Haha yep totally agreed! Thanks for making the update

codecov/settings_base.py
ajay-sentry: Totally makes sense! I agree with you having it ~15 for the depth makes a lot of sense. Could probably lower the alias one if we really wanted, though not a deal breaker

ajay-sentry (approved):
Looks good to me!

suejung-sentry:
Some other things tested - 
smoke tested around staging app that no pages had any regression as a result of these rules
tested in staging a query with many aliases above and below the threshold and behaved as expected
will track any latency delta in our request latency histogram

PR #956, opened 2024-11-04
fix: Don't show detailed errors for anonymous users

RulaKhaled:
### Purpose/Motivation
Screenshot 2024-11-04 at 6 05 13 PM

### Links to relevant tickets
https://github.com/codecov/internal-issues/issues/915

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/views.py
+++ b/graphql_api/views.py
@@ -292,6 +292,8 @@ class AsyncGraphqlView(GraphQLAsyncView):
 
     def context_value(self, request, *_):
         request_body = json.loads(request.body.decode("utf-8")) if request.body else {}
+        self.request = request
+
         return {
             "request": request,
             "service": request.resolver_match.kwargs["service"],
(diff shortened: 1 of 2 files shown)


graphql_api/views.py
```diff
@@ -300,9 +302,11 @@ def context_value(self, request, *_):
         }
 
     def error_formatter(self, error, debug=False):
+        user = self.request.user
```
ajay-sentry, line 305: if user is a "none" value wouldn't line 306 return a can't access on undefined error? 

maybe we could do a self.request.get("user",{})

RulaKhaled: Not really! request.user is automatically populated by the AuthenticationMiddleware middleware, it exists in all requests as an instance of User or AnonymousUser, and fwiw we're not checking for user anywhere else in the view

ajay-sentry (approved):

graphql_api/views.py
ajay-sentry: got it! Thanks for clarifying

PR #958, opened 2024-11-04
add true tokenless to all upload endpoints, remove from TR

nora-codecov:
### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/2300

### What does this PR do?
Auth stack on upload endpoints needs to match what's on `shelter`, which treats all upload endpoints equally.

### DIFF
--- a/upload/views/test_results.py
+++ b/upload/views/test_results.py
@@ -14,7 +14,6 @@ from codecov_auth.authentication.repo_auth import (
     OrgLevelTokenAuthentication,
     RepositoryLegacyTokenAuthentication,
     TokenlessAuthentication,
-    UploadTokenRequiredAuthenticationCheck,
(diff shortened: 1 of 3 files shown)


upload/views/test_results.py
```diff
@@ -59,7 +58,6 @@ class TestResultsView(
 ):
     permission_classes = [UploadTestResultsPermission]
     authentication_classes = [
-        UploadTokenRequiredAuthenticationCheck,
```
nora-codecov, line 62: TR needs a custom method, this is in an upcoming pr. We can leave this auth method in the stack, but it would never evaluate to True.

Swatinem (approved):

adrian-codecov (approved):

PR #959, opened 2024-11-04
fix: Race condition in subscription creation webhook handlers

spalmurray-codecov:
Fixes weird bug in subscription creation webhook flow that caused customer's default payment method to not be set.


ajay-sentry (approved):
dank

PR #962, opened 2024-11-05
feat: Create coverage single upload endpoint

tony-codecov:
### Purpose/Motivation
Closes https://github.com/codecov/engineering-team/issues/2536.

This PR:
1. Extracts logic from CommitViews, ReportViews and UploadViews into functions.

2. Creates a new CombinedUploadViews that handles the full upload flow in a single endpoint, utilizing functions extracted from the views previously.

3. Avoids code duplication in Combined Upload view

4. Updates URL routing for new `combined-upload` endpoint

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate


upload/views/combined_upload.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+
+        inc_counter(
+            API_UPLOAD_COUNTER,
+            labels=generate_upload_prometheus_metrics_labels(
+                action="coverage",
+                endpoint="combined_upload",
+                request=self.request,
+                is_shelter_request=self.is_shelter_request(),
+                position="create_upload",
```
Swatinem, line 137: these metrics take up half the of the function, and they only differ in `position`.

Can you extract these into a separate function `emit_metric(self, position: str)`, so that the actual function logic is more visible?

upload/views/combined_upload.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+            return Response(
+                {
+                    "url": url,
+                    "raw_upload_location": raw_upload_location,
+                },
+                status=status.HTTP_201_CREATED,
+            )
+        else:
+            return Response(
```
Swatinem, line 169: I would early-return using `if not upload`, similar to other places.

Swatinem (commented):
the tests so far only cover error cases. it would be nice to actually test the whole logic as well :-)

otherwise, I personally dislike mixins, but if that fits well within the codestyle of the repo, thats fine as well.

reducing the metrics signal/noise ratio would also be nice to make the actual code more readable.

upload/tests/views/test_combined_upload.py
```diff
@@ -0,0 +1,84 @@
+from unittest.mock import patch
+
+from django.urls import reverse
+from rest_framework.test import APIClient
+from shared.django_apps.core.tests.factories import (
+    RepositoryFactory,
+)
+
+
+class TestCombinedUpload:
```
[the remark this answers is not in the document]
tony-codecov: That makes sense, this should be fixed now.

upload/views/combined_upload.py
tony-codecov: Good point! This should also be fixed, there is now an `emit_metrics` function to handle this.

upload/views/combined_upload.py
tony-codecov: Done!

upload/views/commits.py
```diff
@@ -26,7 +26,14 @@
 log = logging.getLogger(__name__)
 
 
-class CommitViews(ListCreateAPIView, GetterMixin):
+class CommitLogicMixin:
+    def create_commit(self, serializer, repository):
```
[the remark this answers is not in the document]
tony-codecov: I've implemented the logic using functions instead of class method via mixins. Only challenge I had with doing this was the `get_token_for_analytics` function, which uses `self.request`. To avoid passing `self.request` through 4-5 functions, I passed to `get_token_for_analytics` first to get a `token`, and only pass this `token` to `create_upload` instead of passing the entire `self.request` to `create_upload`.

upload/views/uploads.py
```diff
@@ -39,6 +39,131 @@
...
+            "activated",
+            "active",
+            "deleted",
+            "coverage_enabled",
+            "updatestamp",
+        ]
+    )
+
+
+def send_analytics_data(commit: Commit, upload: ReportSession, version, token):
```
Swatinem, line 129: calling this `analytics_token` will hopefully make it more obvious what its purpose is, and that it is only used for analytics.

upload/tests/views/test_combined_upload.py
```diff
@@ -0,0 +1,289 @@
...
+    flag2 = RepositoryFlag.objects.filter(
+        repository_id=repository.repoid, flag_name="flag2"
+    ).first()
+    assert UploadFlagMembership.objects.filter(
+        report_session_id=upload.id, flag_id=flag1.id
+    ).exists()
+    assert UploadFlagMembership.objects.filter(
+        report_session_id=upload.id, flag_id=flag2.id
+    ).exists()
+    assert [flag for flag in upload.flags.all()] == [flag1, flag2]
```
Swatinem, line 182: instead of these exists calls, I would just assert that the flags are properly attached:

```
assert {flag.name for flag in upload.flags.all()} == {"flag1", "flag2"}
```

We shouldn’t care about the implementation details of how the database manages its relations and joins. We care that our upload has the proper flags.

Swatinem (approved):
nice, this looks good, though probably someone else from @codecov/api should also review/approve this.

I have two smaller improvements, otherwise this could sure use some type annotations. And the PR description still mentions mixins, so you might want to update that.

upload/views/uploads.py
tony-codecov: Done!

PR #968, opened 2024-11-11
chore: Add async linting

suejung-sentry:
The Django postgres ORM is synchronous but the GraphQL resolvers are asynchronous. When any ORM commands are attempted to run directly within a resolver, this causes a runtime error `You cannot call this from an async context - use a thread or sync_to_async.`

I investigated different ways we could potentially use lint or type-check tools or even just a unit test that could catch these errors before they hit prod. Unfortunately I had trouble getting anything that could work reliably, and plug in with our existing tools (ruff, mypy, pytest). I explored the below:
* mypy django-stubs plugin, with enable-async-orm-checks = True - https://github.com/typeddjango/django-stubs - running this plugin pointed at our configurations could not resolve our django settings 
* pylint custom plugin (ruff doesn't allow custom plugin) - this required adding support for pylint on top of ruff to get custom plugin functionality - but even trying one out I couldn't get a simple test case to work for our needs
* unit test that traverses resolver functions - this resulted in too many false positives

So in the end I just added in the off-the-shelf async linter offered by flake8 that can catch common errors with async. Unfortunately it doesn't solve the original reason for this investigation re: Django ORM, but can at least be a backstop for other async related pitfalls.

I've also added this "[runbook](https://www.notion.so/sentry/How-to-Fix-sync-to-async-api-errors-6ce4124f7d634e31a34a0811b7ce9a83?pvs=4)" for on-call or debugging response related to graphQL/async in `api`

Closes https://github.com/codecov/engineering-team/issues/2473


ruff.toml
```diff
@@ -36,9 +36,16 @@ indent-width = 4
...
 [lint]
-# Currently only enabled for F (Pyflakes), I (isort), E,W (pycodestyle:Error/Warning), PLC/PLE (Pylint:Convention/Error) 
-# and PERF (Perflint) rules: https://docs.astral.sh/ruff/rules/
-select = ["F", "I", "E", "W", "PLC", "PLE", "PERF"]
+select = [
+    "ASYNC", # flake8-async - async checks
+    "E", # pycodestyle - error rules
+    "F", # pyflakes - general Python errors, undefined names
+    "I", # isort - import sorting
+    "PERF", # perflint - https://docs.astral.sh/ruff/rules/
```
ajay-sentry, line 44: nit: maybe this should be "performance anti-pattern rules" and the rules list should be above

suejung-sentry: oh woops, I misread what the original comment was trying to convey haha. Fixed!

ajay-sentry (approved):
lgtm! Happy to see we didn't have any errors for this guy either 😅

PR #988, opened 2024-11-18
feat: Use uv package manager

suejung-sentry:
Update `api` to use a more modern package manager (`uv`). This will let us more easily convert to monolith in the future.

Shared is getting converted in here: https://github.com/codecov/shared/pull/423
Worker is getting converted in here: https://github.com/codecov/worker/pull/896

### Notable changes
* I moved over existing `requirements.in` 1:1 into new pyproject.toml to use uv-tools for build instead of pip-compile. Any deps that appeared to be used for dev only were added to `dev-dependencies`. The new lockfile will be `uv.lock`
* I updated the existing use of `sentry-sdk[celery]` to fix its build error. As long as we use `CeleryIntegration`, can remove the custom wrapping with `_wrap_async`
* Docker and Makefile stuff was updated to match that done for worker in above

Closes https://github.com/codecov/engineering-team/issues/2779


docker/Dockerfile
```diff
@@ -57,6 +56,7 @@ FROM self-hosted as self-hosted-runtime
 USER root
 ARG EXTERNAL_DEPS_FOLDER=./external_deps
 RUN mkdir $EXTERNAL_DEPS_FOLDER
+# TODO - replace the pip install here?
```
suejung-sentry, line 59: TODO - look into the build for self-hosted

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[tool.setuptools]
+py-modules = []
+
+[tool.uv]
+dev-dependencies = [
```
suejung-sentry, line 70: Pulled from existing requirements.in the ones that looked like only used for development

pyproject.toml
```diff
@@ -0,0 +1,86 @@
+[project]
+name = "api"
+version = "0.1.0"
+description = "Codecov's API layer"
+readme = "README.md"
+requires-python = ">=3.12"
+dependencies = [
```
suejung-sentry, line 7: TODO - these are ported over to align with existing requirements.in. Go through and pin these to max versions in existing requirements.txt so not introducing any regressions by using an old version of deps

PR #989, opened 2024-11-18
Clean up unused `minio.periodic_callback_ms` field

Swatinem:


adrian-codecov (approved):
What was this used for before?

Swatinem:
https://docs.codecov.com/v4.6/docs/archiving-reports#general-instructions mentions this flag.
But otherwise I couldn’t find any reference where this is actually *used*.

PR #997, opened 2024-11-21
fix: Use new func name in sentry-sdk celery config

suejung-sentry:
It appears we have custom configuration of sentry-sdk to address this issue with celery - https://github.com/getsentry/sentry-python/issues/1416

When bumping to the latest [version](https://github.com/getsentry/sentry-python/commit/0fb9606eca582f44897253ed1dda426161c5b3e6) of sentry-sdk, we need to rename the function `_wrap_apply_async` --> `_wrap_task_run`

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -20,7 +20,7 @@ factory-boy
 fakeredis
 freezegun
 https://github.com/codecov/opentelem-python/archive/refs/tags/v0.0.4a1.tar.gz#egg=codecovopentelem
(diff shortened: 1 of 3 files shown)


adrian-codecov (approved):

suejung-sentry:
This PR requires some whack-a-mole with deps versions. We'll handle downgrading shared to be compatible with api instead of the other way around with this pr - https://github.com/codecov/shared/pull/432

PR #1007, opened 2024-11-28
v2 API: add endpoint to activate and deactivate users

JerrySentry:
Introduce endpoint
`PATCH /api/v2/{service}/{owner_username}/users/{user_username_or_ownerid}/`
to activate or deactivate the specified user via request body:
`{activated: boolean}`

closes: https://github.com/codecov/feedback/issues/559

Screenshot 2024-11-28 at 2 58 42 PM

### Legal Boilerplate


suejung-sentry (commented):
I'm probably missing it but where do we do the check that only an admin is allowed to perform this action?

suejung-sentry (approved):

Issue #833, opened 2024-09-23
Implement fake pagination for array based data 


RulaKhaled:
Since we cannot paginate the data at the database level, we’ll implement fake pagination. This approach will simulate pagination by chunking the data on the backend and rendering it incrementally on the frontend.

More context can be found here https://github.com/codecov/internal-issues/issues/656#issuecomment-2358606170 

follow up to https://github.com/codecov/internal-issues/issues/656 investigation

PR #1009, opened 2024-11-29
feat: Create new array pagination, and apply it to path contents 

RulaKhaled:
### Purpose/Motivation
Since we cannot paginate the data at the database level, we’ll implement fake pagination. This approach will simulate pagination by chunking the data on the backend and rendering it incrementally on the frontend.

More context can be found here https://github.com/codecov/internal-issues/issues/656#issuecomment-2358606170

follow up to https://github.com/codecov/internal-issues/issues/656 investigation

### Links to relevant tickets
[#833](https://github.com/codecov/codecov-api/issues/833)

### What does this PR do?
- Create new array pagination 
- Create new deprecated path contents, which we will get rid of once frontend is migrated to using new functionality 

### Notes to Reviewer

### Legal Boilerplate


graphql_api/helpers/connection.py
```diff
@@ -68,6 +69,112 @@ def page_info(self, *args, **kwargs):
         }
 
 
+class ArrayPaginator:
+    """Cursor-based paginator for in-memory arrays."""
```
RulaKhaled, line 73: I was debating between using stringified random cursors and numeric ones for pagination. I decided to go with numeric cursors because they are simple, efficient, and directly map to the array indices (we have cases where customers have over 1,000 files), making pagination faster and easier to debug. Since the dataset is static, there’s no need for the added complexity or security of random strings. If you have other perspectives, lmk

graphql_api/helpers/connection.py
```diff
@@ -68,6 +69,110 @@ def page_info(self, *args, **kwargs):
...
+        last: Optional[int] = None,
+        after: Optional[str] = None,
+        before: Optional[str] = None,
+    ):
+        self.data = data
+        self.start_index = 0
+        self.end_index = len(data)
+
+        if first and last:
+            raise ValueError("Cannot provide both 'first' and 'last'")
```
JerrySentry, line 88: use `ValidationError` instead so it can be handled well in the response, I have a feeling this will be an internal server error at worst or something ugly in the response for the user.

graphql_api/helpers/connection.py
```diff
@@ -68,6 +69,110 @@ def page_info(self, *args, **kwargs):
...
+            self.end_index = min(self.start_index + first, len(data))
+
+        if last is not None:
+            range_length = self.end_index - self.start_index
+            if range_length > last:
+                self.start_index = self.end_index - last
+
+        # Ensure bounds remain valid
+        self.start_index = max(self.start_index, 0)
+        self.end_index = min(self.end_index, len(data))
```
JerrySentry, line 110: This final safe guard is great!

graphql_api/helpers/connection.py
```diff
@@ -68,6 +69,110 @@ def page_info(self, *args, **kwargs):
...
+    @property
+    def has_previous(self) -> bool:
+        """Check if there's a previous page."""
+        return self.start_index > 0
+
+
+class ArrayConnection:
+    """Connection wrapper for array pagination."""
+
+    def __init__(self, data: List[Any], paginator: ArrayPaginator, page: List[Any]):
```
JerrySentry, line 135: You don't actually need `data` and `page` right?
Total count will be `len(self.paginator.data)` and references of `self.page` will be `self.paginator.page`.
IMO adding these two params to the `ArrayConnection` class just adds more confusion

RulaKhaled: You're right, I was trying to decouple the two to some extent, but it's not necessary at this point, as I doubt we'll use it anywhere besides the `query_to_connection` function

JerrySentry (approved):
Looks good!
Codecov is just asking for a test case for `raise ValidationError("'after' cursor must be an integer")`

PR #1037, opened 2024-12-09
[fix] Only count directly changed lines as misses when intended changes is specified

rohitvinnakota-codecov:
Closes https://github.com/orgs/codecov/projects/31/views/11?pane=issue&itemId=88922866&issue=codecov%7Cengineering-team%7C3030

We choose segments with indirectly changed lines even if a user queries for intended changes only if they have diff changes. With this PR, we will exclude those lines from each segment.

### DIFF
--- a/services/comparison.py
+++ b/services/comparison.py
@@ -480,6 +480,15 @@ class Segment:
             if not (line.added or line.removed) and (base_coverage != head_coverage):
                 return True
         return False
+    
+    def remove_unintended_changes(self):
+        filtered = []
+        for line in self._lines:
+            base_cov = line.coverage["base"]
+            head_cov = line.coverage["head"]
+            if (line.added or line.removed) or (base_cov == head_cov):
+                filtered.append(line)
+        self._lines = filtered
 
 
 class FileComparison:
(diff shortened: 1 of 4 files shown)


services/comparison.py
```diff
@@ -481,6 +481,15 @@ def has_unintended_changes(self):
                 return True
         return False
 
+    def remove_unintended_changes(self):
+        filtered = []
+        for line in self._lines:
+            base_cov = line.coverage["base"]
+            head_cov = line.coverage["head"]
+            if (line.added or line.removed) or (base_cov == head_cov):
```
spalmurray-codecov, line 489: This logic says if line is in the diff or coverage on the line has not changed, then it's unintended (coverage change on non-diff line) and we should filter it out. Correct?

services/tests/test_comparison.py
spalmurray-codecov, line 1: Looks good. Just want to confirm that these indirect changes will still show in the indirect changes tab?

spalmurray-codecov (approved):

services/tests/test_comparison.py
rohitvinnakota-codecov: yep

services/comparison.py
rohitvinnakota-codecov: Correct yep

services/tests/test_comparison.py
spalmurray-codecov: Solid lgtm

PR #1039, opened 2024-12-10
feat: Add is___plan fields to plan resolver

ajay-sentry:
### Purpose/Motivation

As part of Milestone 2 for making the developer plan like the team plan, we want to move all the business logic from Gazebo to API; making Gazebo the presentation layer it should've been in the first place.

Currently Gazebo has these helper functions:
- isFreePlan
- isTeamPlan
- isEnterprisePlan
- isBasicPlan
- isPaidPlan -- This will be removed in favor of a check for the "billingRate" prop being none
- isMonthlyPlan -- This will be removed in favor of checking the billingRate prop being "Monthly"
- isAnnualPlan -- This will be removed in favor of checking the billingRate prop being "Annual"
- isSentryPlan
- isCodecovProPlan -- This will be removed outright
- isProPlan
- isTrialPlan
- useProPlans -- this is a copy of findProPlans now that the enterprise flag has been removed lol
- findProPlans -- next PR
- findSentryPlans -- next PR
- findTeamPlans -- next PR

Each of which needs to be updated / modified in tandem with any plan changes that occur on the API side. We basically have 2 sources of Truth for this information, and anytime we want to update that information on one end we end up having to "remember" to update it on the other to avoid breaking something. 

The goal of this PR is to tackle creating all the new resolvers and graphQL types we need for the is____Plan helpers. The find___Plan helpers will be tackled in a separate PR.

### Links to relevant tickets

Related ticket -- [#2996](https://github.com/codecov/engineering-team/issues/2996)

### What does this PR do?

So in this PR, I declare a couple extra properties on the plan service, and just do a check for the current plan name being in any of the existing representations. In the future, when we create the new plan configuration table, we may need to update these functions one more time to just pull the properties from the table directly (depending on how the columns end up shaking up there)

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/plan/plan.graphql
+++ b/graphql_api/types/plan/plan.graphql
@@ -1,7 +1,7 @@
 type Plan {
   baseUnitPrice: Int!
   benefits: [String!]!
-  billingRate: PlanBillingRate!
+  billingRate: String
   hasSeatsLeft: Boolean!
   isEnterprisePlan: Boolean!
   isFreePlan: Boolean!
--- a/graphql_api/types/plan/plan.graphql
+++ b/graphql_api/types/plan/plan.graphql
@@ -1,20 +1,26 @@
 type Plan {
-  trialStatus: TrialStatus!
-  trialStartDate: DateTime
-  trialEndDate: DateTime
-  trialTotalDays: Int
-  pretrialUsersCount: Int
+  baseUnitPrice: Int!
+  benefits: [String!]!
+  billingRate: PlanBillingRate!
+  hasSeatsLeft: Boolean!
+  isEnterprisePlan: Boolean!
+  isFreePlan: Boolean!
+  isProPlan: Boolean!
+  isSentryPlan: Boolean!
+  isTeamPlan: Boolean!
+  isTrialPlan: Boolean!
   marketingName: String!
+  monthlyUploadLimit: Int
   planName: String!
     @deprecated(
       reason: "Plan representations have used `value` for a while, making the frontend code hard to change"
     )
-  value: String!
-  tierName: String!
-  billingRate: String
-  baseUnitPrice: Int!
-  benefits: [String!]!
-  monthlyUploadLimit: Int
   planUserCount: Int
-  hasSeatsLeft: Boolean!
+  pretrialUsersCount: Int
+  tierName: String!
+  trialEndDate: DateTime
+  trialStartDate: DateTime
+  trialStatus: TrialStatus!
+  trialTotalDays: Int
+  value: String!
 }
--- a/graphql_api/types/plan/plan.py
+++ b/graphql_api/types/plan/plan.py
@@ -2,9 +2,7 @@ from datetime import datetime
 from typing import List, Optional
 
 from ariadne import ObjectType
-from shared.plan.constants import (
-    TrialStatus,
-)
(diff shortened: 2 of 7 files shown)


graphql_api/types/plan/plan.py
```diff
@@ -98,3 +96,39 @@ def resolve_plan_user_count(plan_service: PlanService, info) -> int:
 @sync_to_async
 def resolve_has_seats_left(plan_service: PlanService, info) -> bool:
     return plan_service.has_seats_left
+
+
+@plan_bindable.field("isEnterprisePlan")
+@sync_to_async
+def resolve_is_enterprise_plan(plan_service: PlanService, info) -> bool:
+    return plan_service.is_enterprise_plan
```
JerrySentry, line 104: where are these `PlanService.is_*` attributes defined, I couldn't find them in shared.

graphql_api/types/plan/plan.graphql
```diff
@@ -1,20 +1,26 @@
...
+  baseUnitPrice: Int!
+  benefits: [String!]!
+  billingRate: PlanBillingRate!
+  hasSeatsLeft: Boolean!
+  isEnterprisePlan: Boolean!
+  isFreePlan: Boolean!
+  isProPlan: Boolean!
+  isSentryPlan: Boolean!
+  isTeamPlan: Boolean!
+  isTrialPlan: Boolean!
```
JerrySentry, line 11: How are all these boolean fields different from using a enum for plan name?

graphql_api/types/plan/plan.py
ajay-sentry: LOL I totally forgot that we swapped over that stuff to shared, let me open up a new PR for that part. It worked on my local because I modified shared directly

graphql_api/types/plan/plan.graphql
ajay-sentry: A couple aspects to it; first being there are multiple pro / sentry / team / enterprise / free plans so we can't rely on just the name of the plan to uniquely identify the plan type

Secondly, we _could_ use the 'tier name' attribute, and I was thinking about that route as well, but then in the gazebo code instead of having something like 

**BEFORE**
`isFreePlan(accountDetails?.plan?.value)`

**AFTER**
accountDetails?.plan?.isFreePlan

**POTENTIAL AFTER**
accountDetails?.plan?.tier == 'free'

which then kind of leaves us in a similar situation as currently where we have some business logic in the FE. I could go either way tbh, I was just slightly leaning toward keeping gazebo more presentational

JerrySentry: Ok got it
> there are multiple pro / sentry / team / enterprise / free plans
does this mean multiple of these `is_*` booleans can be true for a given org?

JerrySentry (approved):

graphql_api/types/plan/plan.graphql
```diff
@@ -1,20 +1,26 @@
 type Plan {
-  trialStatus: TrialStatus!
-  trialStartDate: DateTime
-  trialEndDate: DateTime
-  trialTotalDays: Int
-  pretrialUsersCount: Int
+  baseUnitPrice: Int!
```
ajay-sentry, line 2: just organizing this in alphabetical

PR #1040, opened 2024-12-10
API Improvements: Fix typings for various files pt1

JerrySentry:
### Legal Boilerplate


billing/views.py
```diff
@@ -410,7 +410,7 @@ def checkout_session_completed(
 
         self._log_updated([owner])
 
-    def post(self, request: HttpRequest, *args, **kwargs) -> Response:
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> Response:
```
adrian-codecov, line 413: Could we know what the args/kwargs are from stripe?

services/profiling.py
```diff
@@ -79,7 +79,12 @@ def _get_critical_files_from_yaml(
             "critical_files_paths"
         ):
             return []
-        commit_sha = self.commit_sha or profiling_commit.commit_sha
+
+        if profiling_commit is not None:
```
adrian-codecov, line 83: Was there something wrong with the one on the left? Or is this mostly preference?

adrian-codecov (approved):
Left comments on the ones I had questions on. Tyvm, this is a great improvement

services/profiling.py
JerrySentry: Hmm good point this change is actually not equivalent, old code gives `self.commit_sha` preference whereas as this gives `profiling_commit.commit_sha` preference. I need to do it like this:
```
        commit_sha = None
        if self.commit_sha:
            commit_sha = self.commit_sha
        elif profiling_commit:
            commit_sha = profiling_commit.commit_sha
```
mypy didn't like the old style because `profiling_commit` might be null so it can produce an error

billing/views.py
JerrySentry: The code doesn't use args/kwargs at all, I think we just add it in as part of the DRF APIView interface.

PR #1065, opened 2024-12-19
feat: Add success after failed payment email (and fix card in failed-payment email)

spalmurray-codecov:
This PR adds the logic for sending the new success-after-failed-payment email template. It also addresses a bug where the card wasn't being properly populated on the other failed-payment email.

Closes https://github.com/codecov/engineering-team/issues/2503
Closes https://github.com/codecov/engineering-team/issues/3119

### DIFF
--- a/billing/views.py
+++ b/billing/views.py
@@ -46,10 +46,41 @@ class StripeWebhookHandler(APIView):
         owners: QuerySet[Owner] = Owner.objects.filter(
             stripe_customer_id=invoice.customer,
             stripe_subscription_id=invoice.subscription,
+            delinquent=True,
         )
-        owners.update(delinquent=False)
 
+        if not owners.exists():
+            return
+
+        admins = get_all_admins_for_owners(owners)
+        owners.update(delinquent=False)
         self._log_updated(list(owners))
+
+        # Send a success email to all admins
+
+        task_service = TaskService()
+        template_vars = {
+            "amount": invoice.total / 100,
+            "date": datetime.now().strftime("%B %-d, %Y"),
+            "cta_link": invoice.hosted_invoice_url,
+        }
+
+        for admin in admins:
+            if admin.email:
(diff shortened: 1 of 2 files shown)


billing/views.py
```diff
@@ -70,9 +101,13 @@ def invoice_payment_failed(self, invoice: stripe.Invoice) -> None:
         admins = get_all_admins_for_owners(owners)
 
         task_service = TaskService()
+        payment_intent = stripe.PaymentIntent.retrieve(
```
spalmurray-codecov, line 104: Turns out you gotta go get the card in the PaymentIntent explicitly. Annoying that it's another fetch, but oh well, this code isn't time critical enough for that to matter.

billing/views.py
```diff
@@ -46,11 +46,42 @@ def invoice_payment_succeeded(self, invoice: stripe.Invoice) -> None:
...
+        task_service = TaskService()
+        template_vars = {
+            "amount": invoice.total / 100,
+            "date": datetime.now().strftime("%B %-d, %Y"),
+            "cta_link": invoice.hosted_invoice_url,
+        }
+
+        for admin in admins:
+            if admin.email:
+                task_service.send_email(
```
suejung-sentry, line 70: Can consider putting this in a try/catch with `continue` so if there is one bad email, it doesn't prevent the rest from going out

suejung-sentry (approved):

billing/views.py
spalmurray-codecov: @suejung-sentry as far as I understand it, the `send_email` call just creates a celery task that is later processed, totally async to this code. So, given that it shouldn't be possible for what you're describing to occur

suejung-sentry: Oh nice! Missed that. Yeah looks good to go then!

spalmurray-codecov: Thanks!

PR #1068, opened 2024-12-20
ref: Update shared + Add missing fields for plan representation 

RulaKhaled:
### Purpose/Motivation
Still need the update version of shared after merging this PR: https://github.com/codecov/gazebo/pull/3602  to pass the tests 

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/plan_representation/plan_representation.py
+++ b/graphql_api/types/plan_representation/plan_representation.py
@@ -14,27 +14,27 @@ plan_representation_bindable = ObjectType("PlanRepresentation")
 
 @plan_representation_bindable.field("marketingName")
 def resolve_marketing_name(plan_data: PlanData, info) -> str:
-    return plan_data.marketing_name
+    return plan_data["marketing_name"]
 
 
 @plan_representation_bindable.field("planName")
 def resolve_plan_name(plan_data: PlanData, info) -> str:
-    return plan_data.value
+    return plan_data["value"]
 
 
 @plan_representation_bindable.field("value")
 def resolve_plan_value(plan_data: PlanData, info) -> str:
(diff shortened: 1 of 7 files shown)


graphql_api/types/plan_representation/plan_representation.py
```diff
@@ -14,27 +14,27 @@
 
 @plan_representation_bindable.field("marketingName")
 def resolve_marketing_name(plan_data: PlanData, info) -> str:
```
ajay-sentry, line 16: Did we update the PlanData type somewhere to reflect this? Can't remember if that is included in shared

RulaKhaled: oh yes! shared changes included this, we added convet_to_DTO fn to the type

ajay-sentry: Right but that's just a return type of that function right? The PlanData type still reflects the explicit properties of the dataclass iirc

ajay-sentry: I don't remember if we actually created an interface for convert_to_DTO

RulaKhaled: hmmm yah it's a bit confusing because we're kinda calculating it on the fly within the class and actually returning a dict type after all, we could enhance this at some point to make the typing clearer, but it's not gonna effect things atm

ajay-sentry (approved):

ajay-sentry:
Tested this locally by pulling branch, confirming bug I was seeing previously no longer exists. Will merge this on Monday

PR #1072, opened 2024-12-26
dev: Add Ruff C4 rules

ajay-sentry:
### Purpose/Motivation

Was looking through the ruff rules list and saw these comprehension style rules which I kinda liked. They all had unsafe-auto fixes available but I was able to go through each one and do a manual replace for all of them

If we like it we can do a similar exercise for Shared / Worker, I think the notation is a lot clearer for me but maybe folks think otherwise and we should scrap this idea.

Closes https://github.com/codecov/engineering-team/issues/3151

Rule fixes were:
[C401](https://docs.astral.sh/ruff/rules/unnecessary-generator-set/)
[C403](https://docs.astral.sh/ruff/rules/unnecessary-list-comprehension-set/)
[C405](https://docs.astral.sh/ruff/rules/unnecessary-literal-set/)
[C416](https://docs.astral.sh/ruff/rules/unnecessary-comprehension/)
[C417](https://docs.astral.sh/ruff/rules/unnecessary-map/)

I also added "C408" to the ignore list because there were like 400 errors for that and they were kinda annoying, particularly for logs. Maybe there were a few that were worth but overall juice wasn't worth the squeeze. Example in the screenshot below

Screenshot 2024-12-26 at 10 08 25 AM

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/api/internal/commit/serializers.py
+++ b/api/internal/commit/serializers.py
@@ -41,7 +41,7 @@ class CommitWithFileLevelReportSerializer(CommitSerializer):
         for filename in report.files:
             file_report = report.get(filename)
             file_totals = CommitTotalsSerializer(
-                {key: val for key, val in zip(TOTALS_MAP, file_report.totals)}
+                dict(zip(TOTALS_MAP, file_report.totals))
             )
             files.append(
                 {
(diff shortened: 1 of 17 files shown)


joseph-sentry (approved):
408 is probably worth adding at some point since it's: "[...] slower because the name dict must be looked up in the global scope in case it has been rebound"

nora-codecov (approved):

Swatinem:
I also believe C408 should be trivially auto-fixable.

PR #1073, opened 2024-12-26
dev: Swap Transaction Test case for Test Case in api/internal tests

ajay-sentry:
### Purpose/Motivation

I was looking into ways to speed up our test suite and stumbled upon a comment on the "TestCase" class in Pytest where it mentioned it being quicker to execute than TransactionTestCase

Screenshot 2024-12-26 at 10 57 49 AM

I got to tinkering a bit and saw that with a simple drop in replacement of TestCase from TransactionTestCase we were able to reduce the runtime of our api/internal pathed test suites from ~38 seconds to under 15 🤯 The results honestly astonished me, and this might end up being a better way to tackle the speed problem prior to doing the parallelization effort, which seemed kinda tough due to needing to understand all our test fixtures anyway.

I may end up creating an investigation ticket at some point to try and migrate some more of these over, or hand off to someone else, but this seems really big! Even after this initial set of updates we have 115 files (probably even more due to classes implementing TransactionTestCase and being used throughout) where we can try and make similar modifications.

**PRE**

Screenshot 2024-12-26 at 10 55 17 AM

**POST**

Screenshot 2024-12-26 at 10 55 06 AM

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/services/tests/test_repo_providers.py
+++ b/services/tests/test_repo_providers.py
@@ -3,11 +3,11 @@ from unittest.mock import patch
 
 import pytest
 from django.conf import settings
+from django.test import TransactionTestCase
 from shared.django_apps.core.tests.factories import OwnerFactory, RepositoryFactory
 from shared.torngit import Bitbucket, Github, Gitlab
 
 from codecov.db import sync_to_async
-from codecov.tests.base_test import InternalAPITest
 from codecov_auth.models import (
     GITHUB_APP_INSTALLATION_DEFAULT_NAME,
     GithubAppInstallation,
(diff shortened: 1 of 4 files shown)


services/tests/test_repo_providers.py
```diff
@@ -114,7 +114,7 @@ def test_token_refresh_callback_none_cases(should_have_owner, service, db):
     assert get_token_refresh_callback(owner, service) is None
 
 
-class TestRepoProviderService(InternalAPITest):
```
ajay-sentry, line 117: keeping this one as is due to test failures when not using TransactionTestCase

nora-codecov (approved):
nice find! let's switch as many as we can

PR #1080, opened 2025-01-02
Add filter condition to update less often

adrian-codecov:
### Purpose/Motivation
We're getting hammered by lock contention and trying to minimize calls that lock the DB. This was one of them, so made adjustments for it.

### Links to relevant tickets
https://github.com/codecov/engineering-team/issues/3173
### Legal Boilerplate


webhook_handlers/tests/test_github.py
```diff
@@ -275,8 +275,12 @@ def test_push_updates_only_unmerged_commits_with_branch_name(self):
 
     @patch("redis.Redis.sismember", lambda x, y, z: False)
     def test_push_updates_commit_on_default_branch(self):
-        commit1 = CommitFactory(merged=False, repository=self.repo)
-        commit2 = CommitFactory(merged=False, repository=self.repo)
+        commit1 = CommitFactory(
```
adrian-codecov, line 278: Had to change branch name for these commits to non-default branch names for the test to make sense - I believe it originally wasn't working as intended

webhook_handlers/views/github.py
```diff
@@ -265,9 +267,11 @@ def push(self, request, *args, **kwargs):
                     commits=[commit.get("id") for commit in commits],
                 ),
             )
+        else:
```
adrian-codecov, line 270: Had to do it this way because after a django update, if your original query is no longer satisfied, which was the case when we updated the branch to pushed_to_branch_name, then your queryset no longer uses records. This is the same logic because both if's do an update, only difference is that the first if also does the merged=True

webhook_handlers/views/github.py
```diff
@@ -267,7 +269,7 @@ def push(self, request, *args, **kwargs):
             )
 
         log.info(
-            f"Branch name updated for commits to {branch_name}",
+            f"Branch name updated for commits to {pushed_to_branch_name}",
             extra=dict(
```
michelletran-codecov, line 273: This will probably also need to be in the if statement now. We're not updating the branch if it's not the default branch.

webhook_handlers/views/github.py
```diff
@@ -239,24 +240,25 @@ def push(self, request, *args, **kwargs):
...
+            ~Q(branch=pushed_to_branch_name),
             repository=repo,
             commitid__in=[commit.get("id") for commit in commits],
             merged=False,
         )
-        commits_queryset.update(branch=branch_name)
-        if branch_name == repo.branch:
-            commits_queryset.update(merged=True)
+
+        if pushed_to_branch_name == repo.branch:
```
michelletran-codecov, line 260: For more optimization, we can probably also call this before querying.

webhook_handlers/views/github.py
adrian-codecov: Pushed and consolidated into 1 log

webhook_handlers/views/github.py
adrian-codecov: donee

PR #1083, opened 2025-01-07
feat: Add ACH payment method

suejung-sentry:
This PR does below to add compatibility for ACH payment method.
1. Adds a Stripe createSetupIntent api (via GraphQL) so our frontend can fetch a clientSecret when it wants to save payment methods using the UI `PaymentElement` (formerly using the `CardElement` that does not require this).
2. Updates to use the Stripe dashboard as the source of truth for the payment method configuration ID in the `CheckoutSession` and the new `PaymentElement`. If not specified, it uses the Default in the stripe dashboard.
3. Adds a `should_propagate` knob on the existing `update_email` api as the ACH payment method requires an associated billing_email (where it was optional on credit card). With our UI designs, we specify email in a separate component, so have that component propagate the info to the default_payment_method using this updated api.
4. Adds the serializers and GraphQL types for the U.S. bank account (ACH) related info for the Views in gazebo UI.

The changes to these apis are all backwards compatible so can go in before any UI changes go live.

Part of this project: https://github.com/codecov/engineering-team/issues/2622
Closes https://github.com/codecov/engineering-team/issues/3115
Closes https://github.com/codecov/engineering-team/issues/3114
Closes https://github.com/codecov/engineering-team/issues/3110


api/internal/owner/views.py
```diff
@@ -80,12 +80,31 @@ def update_payment(self, request, *args, **kwargs):
     @action(detail=False, methods=["patch"])
     @stripe_safe
     def update_email(self, request, *args, **kwargs):
+        """
+        Update the email address associated with the owner's billing account.
+
+        Args:
+            request: The HTTP request object containing:
+                - new_email: The new email address to update to
+                - should_propagate_to_payment_methods: Optional boolean flag to update email on payment methods (default False)
```
ajay-sentry, line 89: nit: can maybe make this param a little less verbose with "apply_to_payment_methods" or similar

codecov/settings_base.py
```diff
@@ -429,6 +429,10 @@
 SHELTER_PUBSUB_PROJECT_ID = get_config("setup", "shelter", "pubsub_project_id")
 SHELTER_PUBSUB_SYNC_REPO_TOPIC_ID = get_config("setup", "shelter", "sync_repo_topic_id")
 
+STRIPE_PAYMENT_METHOD_CONFIGURATION_ID = get_config(
+    "setup", "stripe", "payment_method_configuration", default=None
```
ajay-sentry, line 433: nit: payment_method_configuration_id to match naming convention of above

codecov_auth/commands/owner/interactors/create_stripe_setup_intent.py
```diff
@@ -0,0 +1,39 @@
...
+        if not current_user_part_of_org(self.current_owner, owner_obj):
+            raise Unauthorized()
+
+    def create_setup_intent(self, owner_obj: Owner) -> stripe.SetupIntent:
+        try:
+            billing = BillingService(requesting_user=self.current_owner)
+            return billing.create_setup_intent(owner_obj)
+        except Exception as e:
+            log.error(
+                f"Error getting setup intent for owner {owner_obj.ownerid}",
```
ajay-sentry, line 30: nit: By making the log generic and adding owner_id as an "extra" log param, we could filter on either the owner_id or that particular log message a little easier

services/billing.py
```diff
@@ -595,6 +604,27 @@ def update_email_address(self, owner: Owner, email_address: str):
...
+                )["invoice_settings"]["default_payment_method"]
+
+                stripe.PaymentMethod.modify(
+                    default_payment_method,
+                    billing_details={"email": email_address},
+                )
+                log.info(
+                    f"Stripe successfully updated billing email for payment method {default_payment_method}"
+                )
+            except Exception:
```
ajay-sentry, line 620: is there a set of exceptions we can narrow this down to? And / or logging the exception type

services/billing.py
```diff
@@ -595,6 +604,27 @@ def update_email_address(self, owner: Owner, email_address: str):
             f"Stripe successfully updated email address for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
         )
 
+        if should_propagate_to_payment_methods:
```
ajay-sentry, line 607: It looks like we're only updating a singular payment method, so wondering if we should remove the "s" from payment_methods as well for clarity 🤔

services/billing.py
```diff
@@ -664,6 +694,22 @@ def apply_cancellation_discount(self, owner: Owner):
                 coupon=owner.stripe_coupon_id,
             )
 
+    @_log_stripe_error
+    def create_setup_intent(self, owner: Owner) -> stripe.SetupIntent:
+        log.info(
+            "Stripe create setup intent for owner",
+            extra=dict(
+                owner_id=owner.ownerid,
+                user_id=self.requesting_user.ownerid,
```
ajay-sentry, line 703: nit: this might be more aptly named requesting_owner_id

services/billing.py
```diff
@@ -781,3 +837,9 @@ def update_billing_address(self, owner: Owner, name: str, billing_address):
 
     def apply_cancellation_discount(self, owner: Owner):
         return self.payment_service.apply_cancellation_discount(owner)
+
+    def create_setup_intent(self, owner: Owner):
+        """
+        Creates a SetupIntent for the given owner to securely collect payment details
```
ajay-sentry, line 843: Could we also add the SetupIntent API doc to this comment?

api/internal/owner/views.py
suejung-sentry: Sure, saw your other comment about this too. Updated to `apply_to_default_payment_method`. In theory it may be nicer to have something like passing in the paymentMethodIDs instead of such a specific flag on the api, but I think it's good enough for now, especially if adding in the secondary payment method later causes this to change.

codecov/settings_base.py
suejung-sentry: Good catch - fixed!

codecov_auth/commands/owner/interactors/create_stripe_setup_intent.py
suejung-sentry: Nice good point - fixed!

services/billing.py
suejung-sentry: Good question, I think the errors Stripe can throw are wide but we handle them the same so may just throw generically here (seems others in this Service do the same too)... 
I logged the exception type at least, good callout.

services/billing.py
suejung-sentry: Yeah good point. For now it's just updating the default payment method but in theory when we add more payment methods we can propagate to all of them. Actually I think we may want to update the design to collect an email per payment method instead of the current way we have it. In the meantime I've updated this flag to be `apply_to_default_payment_method`

services/billing.py
suejung-sentry: Ah true. I re-named it requesting_user_id since it seems "requesting_user" is more common as the name of this concept

services/billing.py
suejung-sentry: Good call - added!

codecov_auth/commands/owner/interactors/create_stripe_setup_intent.py
```diff
@@ -0,0 +1,42 @@
...
+                extra={
+                    "ownerid": owner_obj.ownerid,
+                    "error": str(e),
+                },
+            )
+            raise ValidationError("Unable to create setup intent")
+
+    @sync_to_async
+    def execute(self, owner: str) -> stripe.SetupIntent:
+        owner_obj = Owner.objects.filter(username=owner, service=self.service).first()
```
ajay-sentry, line 40: just to confirm, is owner / service a unique pairing?

services/billing.py
```diff
@@ -595,6 +604,28 @@ def update_email_address(self, owner: Owner, email_address: str):
...
+                default_payment_method = stripe.Customer.retrieve(
+                    owner.stripe_customer_id
+                )["invoice_settings"]["default_payment_method"]
+
+                stripe.PaymentMethod.modify(
+                    default_payment_method,
+                    billing_details={"email": email_address},
+                )
+                log.info(
+                    f"Stripe successfully updated billing email for payment method {default_payment_method}"
```
ajay-sentry, line 618: nit: having the stripe customer id in this log too might be beneficial

services/billing.py
```diff
@@ -595,6 +604,28 @@ def update_email_address(self, owner: Owner, email_address: str):
...
+                )
+            except Exception as e:
+                log.error(
+                    "Unable to update billing email for payment method",
+                    extra=dict(
+                        payment_method=default_payment_method,
```
ajay-sentry, line 624: same with this one (and the email too maybe?)

ajay-sentry (approved):
lgtm! Just two more minor nits

services/billing.py
suejung-sentry: fixed that thanks!

services/billing.py
suejung-sentry: Good call, fixed! I omitted email since that's PII technically

codecov_auth/commands/owner/interactors/create_stripe_setup_intent.py
suejung-sentry: I looked around and think I'll just have to leave this one. Otherwise need to rework some things and it seems this pattern is working fine for most else around in this repo. It will be highly unlikely that the provider/owner combo will end up not unique, especially as here where we check that the logged in user is part of the target org. Thanks for bringing this up though

PR #1087, opened 2025-01-09
API Improvements: Fix typings for various files pt5

JerrySentry:
### Legal Boilerplate

### DIFF
--- a/upload/helpers.py
+++ b/upload/helpers.py
@@ -1,7 +1,7 @@
 import logging
 import re
 from json import dumps
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import jwt
 from asgiref.sync import async_to_sync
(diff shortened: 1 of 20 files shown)


upload/helpers.py
```diff
@@ -671,8 +691,8 @@ def parse_headers(headers, upload_params):
     else:
         content_type = (
             "text/plain"
-            if headers.get("X_Content_Type") in (None, "text/html")
-            else headers.get("X_Content_Type")
+            if headers.get("X_Content_Type", "") in (None, "text/html")
+            else headers.get("X_Content_Type", "")
```
suejung-sentry, line 695: This change means that where we used to default to `X_Content_Type: text/plain` when the header is None, now we'd end up with `X_Content_Type: ""` empty string. So I think that means we need to revert to the prior?

Either that or
```
    if headers.get("X_Content_Type", "") in ("", "text/html")
```

suejung-sentry (approved):
This is awesome!! Thank you for doing it!
Ideally we'd have fewer generic Any or Dict, but I know that gets messy, so this is a great step forward for now.

upload/helpers.py
```diff
@@ -437,31 +449,28 @@ def determine_upload_commit_to_use(upload_params, repository):
     # Check if this is a merge commit and, if so, use the commitid of the commit being merged into per the merge commit message.
     # See https://docs.codecov.io/docs/merge-commits for more context.
     service = repository.author.service
+    commitid = upload_params.get("commit", "")
```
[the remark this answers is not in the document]
JerrySentry: From what I can tell it shouldn't, if anything a `None` would actually cause an error in some cases downstream or upstream.

upload/helpers.py
JerrySentry: Ah good point!

adrian-codecov (approved):

PR #1089, opened 2025-01-10
Migrate some `django_apps` to `shared`

Swatinem:
This effects the following: `compare`, `labelanalysis`, `profiling` and `staticanalysis`

The models and factories are re-exported from `shared`, and the `MIGRATION_MODULES` is properly set up so that the migrations are being imported and ran from `shared`, and the local migrations are thus being removed.

---

This depends on https://github.com/codecov/shared/pull/469


adrian-codecov (commented):
https://github.com/codecov/shared/pull/469#pullrequestreview-2542799565, as long as they seem good then I'm 👌 

Also, you don't have to if this is beyond your scope, but are you okay getting rid of the migrations for the other apps that have been moved? Core, codecov_auth, report and legacy_migrations*? Those should also be good to delete.

If it's beyond your scope, could you create a ticket nonetheless and hopefully we'll get to it in the future :]

* I forget if legacy_migrations has some quirks to their migrations, maybe not, but leaving it just in case

adrian-codecov (approved):

JerrySentry (approved):

PR #1090, opened 2025-01-10
Remove `migrations` migrated to `shared`

Swatinem:
All the migrations which have been migrated to `shared` and are being referenced via `MIGRATION_MODULES` can be removed from `api`, as that code is never being imported from here.


adrian-codecov (approved):
Godsend 😂 I had made a comment in another ticket to see if you could do this - thank you. 

Left another comment as well but approved. Reemphasizing to manually test this in some capacity to ensure nothing breaks since it's a lot of files moved

codecov_auth/models.py
```diff
@@ -1,2 +1 @@
 from shared.django_apps.codecov_auth.models import *
-from shared.django_apps.codecov_auth.models import _generate_key  # noqa: F401
```
adrian-codecov, line 2: Is this safe to remove? I forget specifically why this was here in the first place, maybe for some migrations?

Swatinem: searching for this across the code did not yield any results, so 🤷🏻‍♂️

adrian-codecov: Did a quick search and found it here https://github.com/codecov/codecov-api/blob/2f0dbaf5817d46cb5fd5652b1246c3574d8d5b7f/core/migrations/0001_initial.py#L95

So I think by completely removing that migration from API and putting it into Shared we should be able to delete it

Swatinem: Indeed, the migration as it exists in shared already has copied this code:
https://github.com/codecov/shared/blob/95c63cd9647a825355db6b75c370ccc32d0b1704/shared/django_apps/core/migrations/0001_initial.py#L13

JerrySentry (approved):

PR #1094, opened 2025-01-13
feat: Receive user name for  new Terms of Service UI

calvin-codecov:
### Purpose/Motivation

New onboarding changes for pre-PAT appless to make it easier and more clear for users to sign up. We are changing up the required information that a user must fill out to complete onboarding. We will no longer be asking for customer intent and for the user to select a default organization in step that follows this screen.

### Links to relevant tickets

https://github.com/codecov/engineering-team/issues/3174

Changes related to the UI changes designated in this Figma: https://www.figma.com/design/SsoxtY2SB73l0wiLbJ8FhZ/GH-2092?node-id=2288-16704&t=sMCRUMT3J1bwIct1-1

### What does this PR do?
For the onboarding page, we are:
- adding <ins>**name**</ins> input and requiring it
- newly requiring <ins>**email**</ins> input (previously only required when user was opting into marketing updates)
- no longer asking for customer intent (Personal or Business)

### Notes to Reviewer

### Legal Boilerplate


codecov_auth/commands/owner/interactors/save_terms_agreement.py
```diff
@@ -11,30 +11,29 @@
...
     requires_service = False
 
     def validate(self, input: TermsAgreementInput) -> None:
-        valid_customer_intents = ["Business", "BUSINESS", "Personal", "PERSONAL"]
-        if (
-            input.customer_intent
-            and input.customer_intent not in valid_customer_intents
-        ):
-            raise ValidationError("Invalid customer intent provided")
+        print("qwerty input: ", input)
```
JerrySentry, line 24: print

graphql_api/types/user/user.graphql
```diff
@@ -5,5 +5,6 @@ type User {
   student: Boolean!
   studentCreatedAt: DateTime
   studentUpdatedAt: DateTime
+  # this will no longer be updated from the UI with Appless
```
JerrySentry, line 8: we can't delete the `customerIntent: String` line?

graphql_api/types/user/user.py
```diff
@@ -12,37 +13,42 @@
...
+def resolve_student_updated_at(
+    user: Owner, info: GraphQLResolveInfo
+) -> Optional[datetime]:
     return user.student_updated_at
 
 
+# this will no longer be updated from the UI
 @user_bindable.field("customerIntent")
-def resolve_customer_intent(user: Owner, info) -> str:
+def resolve_customer_intent(user: Owner, info: GraphQLResolveInfo) -> Optional[str]:
```
JerrySentry, line 51: similarly should this be deleted?

graphql_api/types/user/user.graphql
JerrySentry: Wait so if you are changing the interface of SaveTermsAgreementInput then would gazebo not break when this change is deployed?

calvin-codecov: I think it would... it sounds like I would need to change this to be a phased rollout in that case. I think I would need to add back `customerIntent` for now and then just loosen the validations on it as we can't expect it to be passed through from the new UI form. Does that sound accurate?

JerrySentry: Yeah sounds about right.
1. Accept either name or customer intent
2. Switch gazebo to use name
3. Remove customer intent from api

JerrySentry (approved):

PR #1102, opened 2025-01-15
[feat] Add endpoint that checks whether an owner has Gen AI consent

rohitvinnakota-codecov:
We want to authenticate 2 ways from Seer <-> Codecov before invoking any of our AI features. This endpoint will be called from Seer to verify that the owner has a valid app installation with Codecov. 

### Legal Boilerplate

### DIFF
--- a/api/internal/gen_ai/serializers.py
+++ b/api/internal/gen_ai/serializers.py
@@ -1,0 +1,15 @@
+from graphql_api.types.owner.owner import AI_FEATURES_GH_APP_ID
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from rest_framework import serializers
+
+from codecov_auth.models import GithubAppInstallation, Owner
+from shared.django_apps.core.models import Repository
+from shared.license import get_current_license
+
+
+class GenAIAuthSerializer(serializers.Serializer):
+    is_valid = serializers.BooleanField()
+    repos = serializers.ListField(child=serializers.CharField(), required=False)
+
--- a/api/internal/gen_ai/urls.py
+++ b/api/internal/gen_ai/urls.py
@@ -1,0 +1,7 @@
+from django.urls import path
+
+from .views import GenAIAuthView
+
+urlpatterns = [
+    path("", GenAIAuthView.as_view(), name="gen-ai-auth"),
+]
--- a/api/internal/gen_ai/views.py
+++ b/api/internal/gen_ai/views.py
@@ -1,0 +1,38 @@
+from graphql_api.types.owner.owner import AI_FEATURES_GH_APP_ID
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+from shared.license import get_current_license
+
+from .serializers import GenAIAuthSerializer, LicenseSerializer
+
+from codecov_auth.models import (
+    GithubAppInstallation,
+    Owner,
+)
+
+class GenAIAuthView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request, *args, **kwargs):
+        owner_id = request.query_params.get("owner_id")
+        owner = Owner.objects.filter(pk=owner_id, service="github").first()
+
+        ai_features_app_install = None
+        repos = []
+
+        if owner:
+            ai_features_app_install = GithubAppInstallation.objects.filter(
+                app_id=AI_FEATURES_GH_APP_ID, owner=owner
+            ).first()
+
+        if ai_features_app_install and ai_features_app_install.repository_service_ids:
+                repos = ai_features_app_install.repository_service_ids
+
+        data = {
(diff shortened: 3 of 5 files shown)


codecov/settings_staging.py
```diff
@@ -4,9 +4,7 @@
 
 DEBUG = False
 THIS_POD_IP = os.environ.get("THIS_POD_IP")
-ALLOWED_HOSTS = get_config(
-    "setup", "api_allowed_hosts", default=["stage-api.codecov.dev"]
-)
+ALLOWED_HOSTS = ["*"]
```
rohitvinnakota-codecov, line 7: undo this

api/gen_ai/views.py
```diff
@@ -0,0 +1,62 @@
...
+    serializer_class = GenAIAuthSerializer
+
+    def validate_signature(self, request):
+        key = get_config(
+            "gen_ai", "auth_secret", default=b"testixik8qdauiab1yiffydimvi72ekq"
+        )
+        if isinstance(key, str):
+            key = key.encode("utf-8")
+        expected_sig = request.headers.get("HTTP-X-GEN-AI-AUTH-SIGNATURE")
+        print(request.headers)
```
mdtro, line 30:
```suggestion
```

api/gen_ai/views.py
```diff
@@ -0,0 +1,62 @@
...
+        computed_sig = (
+            "sha256=" + hmac.new(key, request.body, digestmod=sha256).hexdigest()
+        )
+        if not (expected_sig and constant_time_compare(computed_sig, expected_sig)):
```
mdtro, line 34:
```suggestion
        if not hmac.compare_digest(computed_sig, expected_sig):
```

I'd probably simplify this and just use `hmac`s `compare_digest`. 

I'd also suggest erroring out earlier if `expected_sig` is none with a HTTP 400. This would keep you from spending time calculating the payload signature when there's no signature in the headers anyway.

api/gen_ai/views.py
```diff
@@ -0,0 +1,62 @@
...
+log = logging.getLogger(__name__)
+
+
+class GenAIAuthView(APIView):
+    permission_classes = [AllowAny]
+    serializer_class = GenAIAuthSerializer
+
+    def validate_signature(self, request):
+        key = get_config(
+            "gen_ai", "auth_secret", default=b"testixik8qdauiab1yiffydimvi72ekq"
```
mdtro, line 25: Is it possible to avoid setting a default here and instead error out if it is not set?

api/gen_ai/tests/test_gen_ai.py
```diff
@@ -0,0 +1,120 @@
...
+from codecov_auth.models import GithubAppInstallation
+
+PAYLOAD_SECRET = b"testixik8qdauiab1yiffydimvi72ekq"
+VIEW_URL = reverse("auth")
+
+
+def sign_payload(payload, secret=PAYLOAD_SECRET):
+    data = json.dumps(payload, separators=(",", ":")).encode("utf-8")
+    signature = "sha256=" + hmac.new(secret, data, digestmod=sha256).hexdigest()
+    return signature, data
```
mdtro, line 20:
```suggestion
def sign_payload(payload: bytes, secret=PAYLOAD_SECRET):
    signature = "sha256=" + hmac.new(secret, payload, digestmod=sha256).hexdigest()
    return signature, data
```

I would avoid going from bytes -> string -> bytes with `json.dumps()`. Ordering of the keys isn't guaranteed (afaik), this can result in different signatures. The recommendation when working with payload signatures like this is to always calculate them based on the raw bytes.

rohitvinnakota-codecov: This is just for the tests as I am mimicking a request. The actual code(tested to verify) does not do any conversions like that.

api/gen_ai/views.py
rohitvinnakota-codecov: Our general pattern in this codebase is to do this for testing.

trent-codecov (approved):

api/gen_ai/views.py
mdtro: I believe the extra `expected_sig and` is going to open this up to potential timing attacks since it does not use the constant time compare. The extra check is not needed, if it's `None` the comparison will fail.

api/gen_ai/tests/test_gen_ai.py
mdtro: If the actual code isn't doing this, I would model the `sign_payload` after it's same process. Otherwise you may end up with some unexpected test failures.

api/gen_ai/views.py
mdtro: What about self-hosted users? Will all of those installations end up with this same secret?
Is it possible to use Django's `override_settings` to ensure this is set for tests? 
https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.override_settings

mdtro (approved):

PR #1104, opened 2025-01-16
feat: Add Stripe_id to Django Admin

ajay-sentry:
### Purpose/Motivation

This PR adds the ability to add and update a stripe_id for a plan on django admin. The field is not required, as it's not used for free plans.

This will allow us to remove these things: https://github.com/codecov/codecov-api/blob/c0110f86efd7e22a5b4ec28961c663954d245ea2/codecov/settings_prod.py#L18-L38

Screenshot 2025-01-16 at 2 26 05 PM
Screenshot 2025-01-16 at 2 27 03 PM

### Legal Boilerplate


codecov_auth/admin.py
```diff
@@ -791,12 +792,13 @@ class PlanAdmin(admin.ModelAdmin):
     list_display = (
         "name",
         "marketing_name",
-        "base_unit_price",
         "is_active",
+        "tier",
```
ajay-sentry, line 796: added tier and reordered the "billing" parts of the plan to be next to each other for slightly easier readability

adrian-codecov (approved):

PR #1108, opened 2025-01-27
[feat] Add patch totals to the /pulls endpoint

rohitvinnakota-codecov:
### Purpose/Motivation
Addressing https://github.com/codecov/internal-issues/issues/1144

The `diff` field is deprecated due to its flaky responses so we are going to add a `patch` field for changes in the patch coverage of a pull request 

Screenshot 2025-01-28 at 9 12 59 AM

### Legal Boilerplate


api/public/v2/pull/serializers.py
```diff
@@ -30,5 +33,42 @@ class Meta:
...
+            if pc:
+                hits += pc.hits
+                misses += pc.misses
+                partials += pc.partials
+
+        total_branches = hits + misses + partials
+        if total_branches == 0:
+            return None
+
+        return dict(
```
JerrySentry, line 69: There might be a pre-existing serializer that returns these, o/w we should create one.

api/public/v2/pull/serializers.py
```diff
@@ -30,5 +33,42 @@ class Meta:
...
+        for f in cr.impacted_files:
+            pc = f.patch_coverage
+            if pc:
+                hits += pc.hits
+                misses += pc.misses
+                partials += pc.partials
+
+        total_branches = hits + misses + partials
+        if total_branches == 0:
+            return None
```
JerrySentry, line 67: is it better to return None or {hits: 0, misses: 0, ...} for example their PR only touch config files?

api/public/v2/pull/serializers.py
```diff
@@ -30,5 +33,42 @@ class Meta:
             "state",
             "ci_passed",
             "author",
+            "patch",
         )
         fields = read_only_fields
+
+    def get_patch(self, obj: Pull):
+        # 1) Fetch the CommitComparison for (compared_to, head)
+        comparison_qs = CommitComparison.objects.filter(
```
JerrySentry, line 42: can we put the queries in the `CommitComparisonService` class?

api/public/v2/pull/serializers.py
```diff
@@ -30,5 +33,42 @@ class Meta:
             "state",
             "ci_passed",
             "author",
+            "patch",
         )
         fields = read_only_fields
+
+    def get_patch(self, obj: Pull):
```
JerrySentry, line 40: forgot the return type which is why the mypy job is failing.

JerrySentry (approved):
lgtm, just missing a test for the uncovered line :P

PR #1115, opened 2025-01-30
feat: Add list unverified payment methods api

suejung-sentry:
Another piece of work for the ACH microdeposits delayed payment verification flow - https://github.com/codecov/engineering-team/issues/2622

This PR adds an endpoint to fetch from Stripe any payment methods awaiting delayed payment notification (e.g., ACH 2-day later microdeposits).
It hooks into our GraphQL api at a new `billing` key that we can extend and add to as we cut over stuff from the existing /internal/account-details REST api.

It will be used by gazebo UI to reflect this waiting status to the user in that 2-day period.

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -534,14 +534,42 @@ class StripeService(AbstractPaymentService):
                 "metadata": self._get_checkout_session_and_subscription_metadata(owner),
             },
             tax_id_collection={"enabled": True},
-            customer_update={"name": "auto", "address": "auto"}
-            if owner.stripe_customer_id
-            else None,
+            customer_update=(
+                {"name": "auto", "address": "auto"}
+                if owner.stripe_customer_id
+                else None
+            ),
         )
         log.info(
             f"Stripe Checkout Session created successfully for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
         )
         return session["id"]
+
+    def _is_unverified_payment_method(self, payment_method_id: str) -> bool:
+        payment_method = stripe.PaymentMethod.retrieve(payment_method_id)
+
+        is_us_bank_account = payment_method.type == "us_bank_account" and hasattr(
+            payment_method, "us_bank_account"
+        )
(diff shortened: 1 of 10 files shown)


services/billing.py
```diff
@@ -718,6 +720,59 @@ def create_setup_intent(self, owner: Owner) -> stripe.SetupIntent:
             customer=owner.stripe_customer_id,
         )
 
+    @_log_stripe_error
+    def get_unverified_payment_methods(self, owner):
```
ajay-sentry, line 724: nit: Can we add owner type

services/billing.py
```diff
@@ -718,6 +720,59 @@ def create_setup_intent(self, owner: Owner) -> stripe.SetupIntent:
...
+            ),
+        )
+        if not owner.stripe_customer_id:
+            return []
+
+        unverified_payment_methods = []
+
+        # Check payment intents
+        payment_intents = stripe.PaymentIntent.list(
+            customer=owner.stripe_customer_id, limit=100
```
ajay-sentry, line 738: nit: can probably remove this, default is 10 for limit which seems fine

services/billing.py
```diff
@@ -788,6 +846,9 @@ def get_invoice(self, owner, invoice_id):
     def list_filtered_invoices(self, owner, limit=10):
         return self.payment_service.list_filtered_invoices(owner, limit)
 
+    def get_unverified_payment_methods(self, owner):
```
ajay-sentry, line 849: nit: Can we add Owner type

ajay-sentry (approved):
lgtm!

services/billing.py
suejung-sentry: nice, done

services/billing.py
suejung-sentry: nice, done

services/billing.py
suejung-sentry: fixed per our chat to paginate over stripe's api

PR #1116, opened 2025-01-30
feat: Handle webhooks for ACH microdeposits lifecycle

suejung-sentry:
Handles ACH microdeposits delayed payment verification flow.

This PR handles the below new logical flows:
1. User async verifies the ACH microdeposits for the initial checkout session
   - added webhook listener for `payment_intent.succeeded` 
   - once verified, we set the account as the default payment method on customer, subscription, invoice_settings
   - we also upgrade the plan at this point because stripe fulfills the initial checkout session's subscription from incomplete to active. That updated status is picked up by the customer.subscription.updated webhook handler that saves/updates/upgrades plan data in our database.
   - on Stripe's end the "charges_automatically" initital invoice that was pending gets successfully fulfilled
  
2. User async verifies the ACH microdeposits for subsequent paymentMethodUpdate
   - added webhook listener for `setup_intent.succeeded`
   - once verified, we set the account as the default payment method on customer, subscription, invoice_settings

 3. The initial checkout session by Stripe has a "charges_automatically" that issues an invoice that fails when trying to use the ach microdeposits payment method.
    - ignore this case in the webhook listener for stripe invoice.payment_failed that does other stuff like marking the account delinquent 
 
4. If the user abandons their initial checkout session that has a pending ACH, we delete that abandoned one before offering a way to make a new one
    - adjusted the upgrade_plan logic to handle when we have this incomplete subscription. Deletes the abandoned subscription in stripe but let's the `customer.subscription.deleted` listener be the source of truth of then removing it from the owner object in our postgres

Other things of note:
* Stripe's list payment methods api does not included unverified paymentIntent / setupIntent
* There isn't a combined Stripe endpoint to get a list of both payment intents and setup intents or otherwise a list of all unverified payment methods.

Tested end-to-end flows:
1. Setup initial checkout session
    - Card
    - ACH instant
    - ACH microdeposits
2. Change payment method
    - Card
    - ACH instant
    - ACH microdeposits
3. Abandon ACH microdeposits
More about the end to end testing here: https://www.notion.so/sentry/Suejung-s-notes-1838b10e4b5d8075b7a9cca013de502c?pvs=4#18c8b10e4b5d80a29634f0702e743a90

Epic: https://github.com/codecov/engineering-team/issues/2622


billing/constants.py
```diff
@@ -17,6 +17,8 @@ class StripeWebhookEvents:
         "customer.updated",
         "invoice.payment_failed",
         "invoice.payment_succeeded",
+        "payment_intent.succeeded",
+        "setup_intent.succeeded",
```
suejung-sentry, line 21: TODO - these need to be turned on in Stripe dashboard

services/billing.py
```diff
@@ -538,17 +538,45 @@ def create_checkout_session(self, owner: Owner, desired_plan):
...
+                if owner.stripe_customer_id
+                else None
+            ),
         )
         log.info(
             f"Stripe Checkout Session created successfully for owner {owner.ownerid} by user #{self.requesting_user.ownerid}"
         )
         return session["id"]
 
+    def _is_unverified_payment_method(self, payment_method_id: str) -> bool:
```
ajay-sentry, line 552: nit: could we move this to a separate helpers file

services/billing.py
```diff
@@ -866,3 +910,47 @@ def create_setup_intent(self, owner: Owner):
...
+            except Exception as e:
+                log.error(
+                    "Failed to delete subscription",
+                    extra=dict(
+                        subscription_id=subscription.id,
+                        payment_intent_id=payment_intent.id,
+                        error=str(e),
+                    ),
+                )
+                return None
```
ajay-sentry, line 956: nit: the return at end of fn is implicit

services/billing.py
```diff
@@ -866,3 +910,47 @@ def create_setup_intent(self, owner: Owner):
...
+                extra=dict(
+                    subscription_id=subscription.id,
+                    payment_intent_id=payment_intent.id,
+                    payment_intent_status=payment_intent.status,
+                ),
+            )
+            try:
+                # Delete the subscription, which also removes the
+                # pending payment method and unverified payment intent
+                stripe.Subscription.delete(subscription)
```
ajay-sentry, line 937: should we wrap this in a transaction?

services/billing.py
suejung-sentry: I tried it out but ending up leaving it here for now as I was thinking spinning off a helpers file just because this file is unmanageable seems not great imo. Ideally we can split this mega file out instead.. I think creating a new kitchen sink could make it worse so left it here status quo

services/billing.py
suejung-sentry: made the update we discussed about removing the owner save here and letting it happen in the webhook handler for customer.subscription.deleted

services/billing.py
ajay-sentry: Ah okay fair enough, originally I thought this was the webhooks file but its not. I actually meant to add this comment to the new functions added in views.py

billing/views.py
```diff
@@ -105,9 +130,9 @@ def invoice_payment_failed(self, invoice: stripe.Invoice) -> None:
             invoice["payment_intent"], expand=["payment_method"]
         )
         card = (
```
ajay-sentry, line 132: Really not a fan of this, how about 

 Try:
 
 card = payment_intent.payment_method.card
 
 except NotFound:
 
 card = None

billing/views.py
```diff
@@ -311,7 +360,34 @@ def customer_subscription_created(self, subscription: stripe.Subscription) -> No
...
+        that requires verification (e.g. ACH microdeposits). This indicates that
+        there is an unverified payment method from the initial CheckoutSession.
+        """
+        latest_invoice = stripe.Invoice.retrieve(subscription.latest_invoice)
+        if latest_invoice and latest_invoice.payment_intent:
+            payment_intent = stripe.PaymentIntent.retrieve(
+                latest_invoice.payment_intent
+            )
+            return (
+                payment_intent is not None
```
ajay-sentry, line 377: not: don't need the `is not None`

billing/views.py
```diff
@@ -311,7 +360,34 @@ def customer_subscription_created(self, subscription: stripe.Subscription) -> No
...
+                and payment_intent.status == "requires_action"
+                and payment_intent.next_action is not None
```
ajay-sentry, line 379: same comment here, and I saw in the other reference we used .get() if we want to do the same here

billing/views.py
```diff
@@ -445,6 +531,74 @@ def checkout_session_completed(
...
+    ):
+        """
+        Helper method to handle payment methods that require delayed verification (like ACH).
+        When verification succeeds, this attaches the payment method to the customer and sets
+        it as the default payment method for both the customer and subscription.
+        """
+        owner = Owner.objects.get(stripe_customer_id=customer_id)
+        payment_method = stripe.PaymentMethod.retrieve(payment_method_id)
+
+        is_us_bank_account = payment_method.type == "us_bank_account" and hasattr(
```
ajay-sentry, line 545: jw, why do we need both of these conditions?

services/billing.py
```diff
@@ -546,8 +546,34 @@ def create_checkout_session(self, owner: Owner, desired_plan):
...
 
+    def _is_unverified_payment_method(self, payment_method_id: str) -> bool:
+        payment_method = stripe.PaymentMethod.retrieve(payment_method_id)
+
+        is_us_bank_account = payment_method.type == "us_bank_account" and hasattr(
+            payment_method, "us_bank_account"
+        )
+        if is_us_bank_account:
+            setup_intents = stripe.SetupIntent.list(
+                payment_method=payment_method_id, limit=1
```
ajay-sentry, line 557: does this default to the most recent?

services/billing.py
```diff
@@ -546,8 +546,34 @@ def create_checkout_session(self, owner: Owner, desired_plan):
...
+            )
+            if (
```
ajay-sentry, line 559: I'd try and see if we can get away with an eager try catch here

Maybe something like 

Try

latest_intent = setup_intents.data[0]

if latest_intent.status blah blah
except NotFound:

return False

services/billing.py
suejung-sentry: We should only ever have 1 setup intent to 1 paymentMethodId but the only available api to fetch this from stripe is this list with filter, it seemed. I don't think it's possible on their end for it to be more than 1. I'd guess if it is, it'd take the most recent

services/billing.py
```diff
@@ -940,3 +983,44 @@ def create_setup_intent(self, owner: Owner):
         See https://docs.stripe.com/api/setup_intents/create
         """
         return self.payment_service.create_setup_intent(owner)
+
+    def _cleanup_incomplete_subscription(
+        self, subscription: stripe.Subscription, owner: Owner
+    ):
+        latest_invoice = subscription.get("latest_invoice")
```
ajay-sentry, line 990: Similar thought here with the Try

try:
payment_intent_id = subscription["latest_invoice"]["payment_intent"]
except NotFound:
return None

ajay-sentry (approved):
lgtm

billing/views.py
suejung-sentry: I thought it'd be safer to check both as I haven't read through all scenarios where Stripe may populate one but not the other. When we reach this point in the code we expect both of these to be in this state in order to proceed

PR #1119, opened 2025-01-31
feat: Improve upload storage path generation for reports

trent-codecov:
### Purpose/Motivation

### What does this PR do?

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/upload/views/uploads.py
+++ b/upload/views/uploads.py
@@ -1,4 +1,5 @@
 import logging
+import uuid
 from typing import Any, Callable, Dict
 
 from django.http import HttpRequest, HttpResponseNotAllowed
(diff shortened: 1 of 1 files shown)


upload/views/uploads.py
```diff
@@ -56,6 +57,18 @@ def create_upload(
         else None
     )
     archive_service = ArchiveService(repository)
+    # only Shelter requests are allowed to set their own `storage_path`
+    serializer.validated_data["external_id"] = uuid.uuid4()
```
adrian-codecov, line 61: We should be able to directly do `uploadid=uuid.uuid4()`, otherwise we'd be creating this external ID and not using it most of the time. 

If we need to store this external_id in the DB, then I'd just move this line bw 62 and 63 and the rest would be the same

trent-codecov: Good point. I moved it down. I assume we need it because of how it was being done before.

adrian-codecov (approved):

PR #1120, opened 2025-01-31
dev: swap create to update_or_create when inserting plans/tiers

ajay-sentry:
### Purpose/Motivation

Swapping this to be update_or_create in the event we update CSVs locally and want to use hot-refresh to get those changes to persist. 

Also as part of this change we're removing the django commands from being in "ready" because they would run on every "refresh" of the app rather than on startup directly. By putting it in the dev.sh file we're able to add the seeds for just startup when the API container is ran on its own.

### Legal Boilerplate


adrian-codecov (approved):

core/management/commands/codecovPlans-Jan25.csv
```diff
@@ -2,12 +2,12 @@
...
+5,2025-01-16 04:35:34.152 -0800,2025-01-24 11:35:42.724 -0800,10,"{""Includes 5 seats"",""$10 per additional seat"",""Unlimited public repositories"",""Unlimited private repositories"",""Priority Support""}",annually,true,Sentry Pro,5,,true,users-sentryy,5,price_1Mj1mMGlVGuVgOrkC0ORc6iW
+4,2025-01-16 04:34:33.867 -0800,2025-01-24 11:35:48.218 -0800,12,"{""Includes 5 seats"",""$10 per additional seat"",""Unlimited public repositories"",""Unlimited private repositories"",""Priority Support""}",monthly,true,Sentry Pro,5,,true,users-sentrym,5,price_1Mj1kYGlVGuVgOrk7jucaZAa
+3,2025-01-16 04:32:44.655 -0800,2025-01-24 11:36:09.660 -0800,10,"{""Configurable # of users"",""Unlimited public repositories"",""Unlimited private repositories"",""Priority Support""}",annually,true,Pro,,,true,users-pr-inappy,3,plan_H6P16wij3lUuxg
+2,2025-01-16 04:30:42.897 -0800,2025-01-24 11:36:14.651 -0800,12,"{""Configurable # of users"",""Unlimited public repositories"",""Unlimited private repositories"",""Priority Support""}",monthly,true,Pro,,,true,users-pr-inappm,3,plan_H6P3KZXwmAbqPS
```
nora-codecov, line 10: these are junk values/non-prod right? can you change those last 2 to have `price_` instead of `plan`?

core/management/commands/insert_data_to_db_from_csv.py
```diff
@@ -51,7 +51,12 @@ def handle(self, *args, **kwargs):
                         )
                         continue
 
-                Model.objects.create(**model_data)
+                Model.objects.update_or_create(
+                    defaults=model_data,
+                    # Specify the fields to identify the record
+                    # For example, if 'id' is the unique field:
+                    id=row.get("id"),
+                )
```
nora-codecov, line 59: might get weird specifying id like this, but we can try it out and see how it goes

nora-codecov (approved):

core/management/commands/codecovPlans-Jan25.csv
ajay-sentry: nah they're legit values 😅 We need them to actually checkout with stripe locally / on stage

PR #1123, opened 2025-02-03
fix: Add missing service GQL error defn

JerrySentry:
Reduce noisy errors that occurs when the user sends requests without service param provided. This will no longer raise these unhandled errors, leading to Sentry spams.

First one:
- fixes: https://github.com/codecov/internal-issues/issues/1199
- Sentry link: https://codecov.sentry.io/share/issue/938420f7b3694d78876bb51b25338244/
- Looks like we forgot to add Missing Service to the mutation's error set. The code raises it but we never had a error resolver for it.

Second one:
- fixes: https://github.com/codecov/internal-issues/issues/1200
- Sentry link: https://codecov.sentry.io/share/issue/440fa0cfc3e24606a6e6e60de3e67ef8/
- whenever doing `owner` query if a service string (eg "gh", "gl", etc) is not provided we just return null instead of having it raise a unhandled error later downstream.

### Legal Boilerplate


ajay-sentry (approved):
Nice!

PR #1125, opened 2025-02-03
chore: Update shared, remove amplitude

spalmurray-codecov:
Updates shared to c8dca74. Also removes amplitude as a direct dependency because it's now a dependency of shared.


requirements.txt
```diff
@@ -1,15 +1,13 @@
 #
-# This file is autogenerated by pip-compile with Python 3.13
+# This file is autogenerated by pip-compile with Python 3.12
```
adrian-codecov, line 2: 👀 should be python 13 👀

JerrySentry (approved):

adrian-codecov (approved):

PR #1131, opened 2025-02-05
ref: Plan query optimizations

ajay-sentry:
### Purpose/Motivation

Some more plan optimizations for the update_plan helper and when determining if we want to fire the sentry webhook when moving to a sentry plan. Consequently updates the tests.

The update_plan helper change is the main one, mostly just consolidating the 2 queries to a single one and passing the relevant params down that are needed

This should be merged into the remove reference PR before merging that one, but can also be done separately

### Legal Boilerplate

### DIFF
--- a/services/billing.py
+++ b/services/billing.py
@@ -316,9 +316,21 @@ class StripeService(AbstractPaymentService):
         return stripe.SubscriptionSchedule.retrieve(subscription_schedule_id)
 
     @_log_stripe_error
-    def modify_subscription(self, owner, desired_plan):
+    def modify_subscription(self, owner: Owner, desired_plan: dict):
+        desired_plan_info = Plan.objects.filter(name=desired_plan["value"]).first()
+        if not desired_plan_info:
+            log.error(
+                f"Plan {desired_plan['value']} not found",
+                extra=dict(owner_id=owner.ownerid),
+            )
+            return
+
         subscription = stripe.Subscription.retrieve(owner.stripe_subscription_id)
-        proration_behavior = self._get_proration_params(owner, desired_plan)
+        proration_behavior = self._get_proration_params(
+            owner,
+            desired_plan_info=desired_plan_info,
+            quantity=desired_plan["quantity"],
+        )
         subscription_schedule_id = subscription.schedule
 
         # proration_behavior indicates whether we immediately invoice a user or not. We only immediately
@@ -326,7 +338,11 @@ class StripeService(AbstractPaymentService):
         # An increase in seats and/or plan implies the user is upgrading, hence 'is_upgrading' is a consequence
         # of proration_behavior providing an invoice, in this case, != "none"
         # TODO: change this to "self._is_upgrading_seats(owner, desired_plan) or self._is_extending_term(owner, desired_plan)"
-        is_upgrading = True if proration_behavior != "none" else False
+        is_upgrading = (
+            True
+            if proration_behavior != "none" and desired_plan_info.stripe_id
+            else False
+        )
 
         # Divide logic bw immediate updates and scheduled updates
         # Immediate updates: when user upgrades seats or plan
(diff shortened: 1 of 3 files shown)


services/billing.py
```diff
@@ -404,7 +411,11 @@ def modify_subscription(self, owner, desired_plan):
             )
 
     def _modify_subscription_schedule(
-        self, owner: Owner, subscription, subscription_schedule_id, desired_plan
+        self,
```
ajay-sentry, line 414: just some types

services/billing.py
```diff
@@ -453,20 +464,18 @@ def _modify_subscription_schedule(
             metadata=self._get_checkout_session_and_subscription_metadata(owner),
         )
 
-    def _is_upgrading_seats(self, owner: Owner, desired_plan: dict) -> bool:
+    def _is_upgrading_seats(self, owner: Owner, quantity: int) -> bool:
```
ajay-sentry, line 467: swap this to just quantity

services/billing.py
```diff
@@ -453,20 +464,18 @@ def _modify_subscription_schedule(
...
         """
         Returns `True` if purchasing more seats.
         """
-        return bool(
-            owner.plan_user_count and owner.plan_user_count < desired_plan["quantity"]
-        )
+        return bool(owner.plan_user_count and owner.plan_user_count < quantity)
 
-    def _is_extending_term(self, owner: Owner, desired_plan: dict) -> bool:
+    def _is_extending_term(
```
ajay-sentry, line 473: swap this to just the two plan references

services/billing.py
```diff
@@ -475,24 +484,23 @@ def _is_extending_term(self, owner: Owner, desired_plan: dict) -> bool:
             and desired_plan_info.billing_rate == PlanBillingRate.YEARLY.value
         )
 
-    def _is_similar_plan(self, owner: Owner, desired_plan: dict) -> bool:
+    def _is_similar_plan(
+        self,
+        owner: Owner,
```
ajay-sentry, line 489: this one needs all 4 properties, the owner, both plan references, and the quantity

services/billing.py
```diff
@@ -508,25 +516,35 @@ def _is_similar_plan(self, owner: Owner, desired_plan: dict) -> bool:
 
         return bool(is_same_term and is_same_seats)
 
-    def _get_proration_params(self, owner: Owner, desired_plan: dict) -> str:
+    def _get_proration_params(
+        self, owner: Owner, desired_plan_info: Plan, quantity: int
+    ) -> str:
+        current_plan_info = Plan.objects.select_related("tier").get(name=owner.plan)
```
ajay-sentry, line 522: where we're doing the current_plan_info query, the desired_plan one is one level up (saving us an extra query here)

nora-codecov (approved):

services/billing.py
```diff
@@ -316,17 +316,33 @@ def get_schedule(self, owner: Owner):
...
+                extra=dict(owner_id=owner.ownerid),
+            )
+            return
+
         subscription = stripe.Subscription.retrieve(owner.stripe_subscription_id)
-        proration_behavior = self._get_proration_params(owner, desired_plan)
+        proration_behavior = self._get_proration_params(
+            owner,
+            desired_plan_info=desired_plan_info,
+            quantity=desired_plan["quantity"],
```
adrian-codecov, line 332: Nit: would change this to "desired_quantity"

adrian-codecov (approved):
Lgtm! I'm wondering if there's an easy way to cache these generic values during runtime. Not a huge fan of adding many DB queries per subscription/any time we want plan data that's just going to be static. 

I wonder if we could have longterm a redis cache that caches all plan data from the DB for like a day and then we wouldn't be pinging the DB any time we tell it "gimme the basic plan, gimme the pro year plan" etc (this is different than **your** own plan). And we could make it fancier by invalidating the cache if we see a change in the plan/tier database or something

services/billing.py
ajay-sentry: fair point, I'll make the update in the other PR

ajay-sentry: I lied fixed in[7425e16](https://github.com/codecov/codecov-api/pull/1131/commits/7425e161580a62cb7f39dca600aac6054ae7b320)

PR #1139, opened 2025-02-06
fix: a couple of rest api unhandled errors

JerrySentry:
### Legal Boilerplate

### DIFF
--- a/api/public/v2/compare/serializers.py
+++ b/api/public/v2/compare/serializers.py
@@ -16,10 +16,11 @@ class ComparisonSerializer(BaseComparisonSerializer):
 
     def get_files(self, comparison: Comparison) -> List[dict]:
         data = []
-        for filename in comparison.head_report.files:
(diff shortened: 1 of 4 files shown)


api/public/v2/compare/serializers.py
```diff
@@ -16,10 +16,11 @@ class ComparisonSerializer(BaseComparisonSerializer):
 
     def get_files(self, comparison: Comparison) -> List[dict]:
         data = []
-        for filename in comparison.head_report.files:
-            file = comparison.get_file_comparison(filename, bypass_max_diff=True)
-            if self._should_include_file(file):
-                data.append(FileComparisonSerializer(file).data)
+        if comparison.head_report is not None:
```
ajay-sentry, line 19: wondering if we should raise a validation indicating no head report for comparison?

ajay-sentry: In the event it falls out of the if I mean

rohitvinnakota-codecov (approved):

PR #1141, opened 2025-02-07
Refactor and optimize querying `pathContents`

Swatinem:
First of all, this fixes and adds missing type annotations.

It then deduplicates the code between `pathContents` and the `deprecated` variant, which were 99% identical.

Resolving the path contents was also optimized a little bit by not filtering the report for "files matching *any* session in the report", aka "all files". Other optimizations include avoiding some intermediate allocations by sorting lists in-place, and simplifying some functions.

### DIFF
--- a/graphql_api/types/commit/commit.py
+++ b/graphql_api/types/commit/commit.py
@@ -14,6 +14,7 @@ from shared.reports.types import ReportTotals
 import services.components as components_service
 import services.path as path_service
 from codecov.db import sync_to_async
+from codecov_auth.models import Owner
 from core.models import Commit
 from graphql_api.actions.commits import commit_status, commit_uploads
 from graphql_api.actions.comparison import validate_commit_comparison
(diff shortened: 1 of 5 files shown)


joseph-sentry (approved):
qq about a functionality change in get_sorted_paths, but lgtm to merge

graphql_api/types/commit/commit.py
```diff
@@ -214,32 +211,33 @@ def resolve_path_contents(commit: Commit, info, path: str = None, filters=None):
     if len(report_paths.paths) == 0:
         # we do not know about this path
 
-        if path_service.provider_path_exists(path, commit, current_owner) is False:
+        if (
+            not path
```
joseph-sentry, line 215: this seems to be the only place where you're modifying the functionality of `get_sorted_paths` in comparison to before, do you mind leaving an explanation of why it's safe to make this change? it seems like the `provider_path_exists` function is fine with having the `path` = `None` and has a branch to handle that case

Swatinem: thanks for calling this out. I reverted this change.

PR #1142, opened 2025-02-07
Optimize usage of `match`

Swatinem:
This creates stateful `Matcher`s instead, to avoid re-compiling regex matchers in loops.


Swatinem:
In theory, yes. But that code is not being run in a loop, so would not benefit from a stateful matcher.

ajay-sentry:
@Swatinem Got it, I thought your shared PR removed the stateless match() fn but saw it was just at the bottom of the file, lgtm

ajay-sentry (approved):

PR #1144, opened 2025-02-07
dev: Pytest-split test

ajay-sentry:
### Purpose/Motivation

This PR tests adding the pytest-split package and running it locally to see if it could potentially help speed up our CI. 

On the one hand, it was super straight forward to setup and for our CI it would help a lot with running the tests (if we choose 5 groups for example, each would take ~1-2 minutes to run), and the documentation states that we don't need to run --store-durations all that often, unless our test suite changes a lot since it uses the average test duration for each subsequent test added.

On the other hand, this doesn't help with test runtimes for the full suite locally. While you can be selective with the tests you run, I usually run the full suite initially to see which tests have broken for any PR prior to running individual test suites to fix those tests. If it's found that most folks don't usually run the full suite locally as well then maybe it's not a big deal.

This PR handles everything that's needed from the API side, we'd need to spin up some new GH actions though for each test group that would run after though.

They have a [demo repo on their docs](https://github.com/jerry-git/pytest-split-gh-actions-demo) with how to set up with a generic test suite

### UPDATE:

We got it to work and the results are pretty good! CI times for API look to have gone from ~15:30 -> 6:30, or a 58% reduction in CI run time!

Before run: https://github.com/codecov/codecov-api/actions/runs/13274970819
After run: https://github.com/codecov/codecov-api/actions/runs/13315813179

The test-durations file needs to be committed because it's the thing that pytest-split references when it creates the groups. We can choose to commit it every month or year or anything else, but it's not required outside of the first time since it'll just use the suite's average test runtime for all "new" tests added to the suite.

Again, this doesn't modify anything for your local docker setup either, but you could run the test groups locally if you choose to via `pytest --splits "numSplits" --group "groupNum"`

Screenshot 2025-02-07 at 1 49 36 PM
Screenshot 2025-02-07 at 2 02 25 PM
Screenshot 2025-02-07 at 2 02 33 PM

**Where to go from here?**

- If there's a way to cache the "bring test env" step up that might be another way to shave off a minute across each runner since those times can range from 20 seconds to ~2 minutes if we get unlucky with network bandwidth for the runner
- Continue to move tests from TransactionTestCase -> TestCase as we can, since they're like twice as fast

### Legal Boilerplate


Makefile
```diff
@@ -205,10 +213,18 @@ test_env.container_check_db:
...
-	docker-compose exec api make test.unit
+	@if [ -n "$(GROUP)" ]; then \
+		docker-compose exec api make test.unit SPLIT=${SPLIT} GROUP=${GROUP}; \
+	else \
+		docker-compose exec api make test.unit; \
+	fi
 
 test_env.run_integration:
-	#docker-compose exec api make test.integration
+	# @if [ -n "$(GROUP)" ]; then \
```
ajay-sentry, line 223: comment this out if we ever have integration tests and it should just work

.github/workflows/ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
   test:
     name: Test
     needs: [build]
-    uses: codecov/gha-workflows/.github/workflows/run-tests.yml@v1.2.27
+    uses: codecov/gha-workflows/.github/workflows/run-tests-split.yml@285163a75899bad2018fe960ac9dba7530e009fb
```
ajay-sentry, line 52: TODO: update this to the new release hash before merging

graphql_api/tests/test_pull.py
```diff
@@ -459,46 +459,44 @@ def test_with_complete_pull_request(self):
...
-            commit=head, report_type=CommitReport.ReportType.BUNDLE_ANALYSIS
-        )
-
-        pull = self.fetch_one_pull_request(
-            my_pull.pullid, pull_request_bundle_analysis_missing_reports
-        )
-        assert pull == {
-            "bundleAnalysisCompareWithBase": {"__typename": "MissingBaseReport"}
-        }
+    # def test_compare_bundle_analysis_missing_reports(self):
```
suejung-sentry, line 462: maybe leave a comment why this is commented?

Makefile
suejung-sentry: Do you know why is this command commented to begin with?

.test_durations
suejung-sentry, line 1: Is this file how long it took to run our suite? If it gets recommited with every PR, maybe we should just gitignore it?

suejung-sentry (approved):
This is amazing - thank you for doing it for us!

graphql_api/tests/test_pull.py
ajay-sentry: Yeah good point, we did create a ticket here so it doesn't get lost: https://github.com/codecov/engineering-team/issues/3358, I can link that to the test directly too

.test_durations
ajay-sentry: Re: commented out command, it's bc we don't have integration tests in API (yet) 😅 

Re: .test_durations, yes it is how long each individual test ran and it only gets "created" or "updated" when we pass in the --test-durations flag or something so it shouldn't be created unless we explicitly want to

ajay-sentry: its actually the --store-durations flag, I misspoke

graphql_api/tests/test_pull.py
ajay-sentry: Updated

PR #1146, opened 2025-02-07
fix: Account for Multiple Owners being returned when handling stripe payment intent webhook

ajay-sentry:
### Purpose/Motivation

This PR aims to fix a Sentry issue that popped up after we launched ACH related to a customer trying to update their payment method while having multiple stripe subscription_ids tied to the customer_id in our DB. I noticed that for this customer both of the subscription_ids were the same, but wanted the code to accommodate cases where that wasn't necessarily the case; potentially someone who wants to spread Codecov across the many orgs their apart of ;) 

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/3323

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/billing/views.py
+++ b/billing/views.py
@@ -539,7 +539,6 @@ class StripeWebhookHandler(APIView):
         When verification succeeds, this attaches the payment method to the customer and sets
         it as the default payment method for both the customer and subscription.
         """
-        owner = Owner.objects.get(stripe_customer_id=customer_id)
         payment_method = stripe.PaymentMethod.retrieve(payment_method_id)
 
         is_us_bank_account = payment_method.type == "us_bank_account" and hasattr(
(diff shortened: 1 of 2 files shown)


billing/views.py
```diff
@@ -548,19 +547,34 @@ def _check_and_handle_delayed_notification_payment_methods(
...
-            )
-            stripe.Customer.modify(
-                owner.stripe_customer_id,
-                invoice_settings={"default_payment_method": payment_method},
-            )
-            stripe.Subscription.modify(
-                owner.stripe_subscription_id, default_payment_method=payment_method
+            # retrieve the number of owners to update
+            owners = Owner.objects.filter(
+                stripe_customer_id=customer_id, stripe_subscription_id__isnull=False
```
suejung-sentry, line 554: nice, i like adding the filter for not null

suejung-sentry (approved):
Thanks for fixing that!

PR #1153, opened 2025-02-12
update other repo actions

ajay-sentry:
### Purpose/Motivation

This PR aims to uptake additional GH webhook actions for the repository webhook, specifically the following:
- Created
- Edited (for default branch updates)
- Transferred
- Archived
- Unarchived
- Renamed

I think if we uptake these there's a possibility some load can be removed from sync_repos to find and update stale repos, and I think in the case of archived/unarchived, those are not handled even within sync_repos.

Plus there's some "quality of life" adjustment here with the "created" action, no longer needing to rely on sync_repos to pull it in and create the repo. Generally, it seems like sync_repos has been "bandaged" together to handle all these edge cases when if we solve the problem upstream (in the webhook handler), we might be able to get rid of a lot of that stuff if we can get confidence that these webhooks keep things in sync on their own.

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/2938

### Notes to Reviewer

### Legal Boilerplate

### DIFF
--- a/webhook_handlers/views/github.py
+++ b/webhook_handlers/views/github.py
@@ -103,10 +103,10 @@ class GithubWebhookHandler(APIView):
     def unhandled_webhook_event(self, request, *args, **kwargs):
         return Response(data=WebhookHandlerErrorMessages.UNSUPPORTED_EVENT)
 
-    def _get_repo(self, request):
+    def _get_repo(self):
         """
-        Attempts to fetch the repo first via the index on o(wnerid, service_id),
-        then naively on service, service_id if that fails.
+        Attempts to fetch the repo first via the index on (ownerid, service_id),
+        then naively on (service, service_id) if that fails.
         """
         repo_data = self.request.data.get("repository", {})
         repo_service_id = repo_data.get("id")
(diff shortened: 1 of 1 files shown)


webhook_handlers/views/github.py
```diff
@@ -119,67 +119,74 @@ def _get_repo(self, request):
...
+            repo_data = self.request.data.get("repository", {})
+            repo_service_id = repo_data.get("id")
+            owner_service_id = repo_data.get("owner", {}).get("id")
+            owner, _ = Owner.objects.get_or_create(
+                service=self.service_name, service_id=owner_service_id
+            )
+            repo, _ = Repository.objects.get_or_create(
+                author__ownerid=owner.ownerid, service_id=repo_service_id
+            )
+            return Response()
```
matt-codecov, line 177: does the webhook give us its public/private status? if we can't know whether the repo is meant to be private, i don't think we can create it here

for public repos this would make them show up without having to trigger a sync, but for private repos you still have to sync to populate `owner.permission`

webhook_handlers/views/github.py
```diff
@@ -195,33 +202,78 @@ def repository(self, request, *args, **kwargs):
...
+                "Repository archived",
+                extra=dict(repoid=repo.repoid, github_webhook_event=self.event),
+            )
+        elif action == "unarchived":
+            repo.activated = True
+            repo.save(update_fields=["activated"])
+            log.info(
+                "Repository unarchived",
+                extra=dict(repoid=repo.repoid, github_webhook_event=self.event),
+            )
```
matt-codecov, line 250: is this actually what "activated" means in our system? i figured it had plan implications or something

webhook_handlers/views/github.py
```diff
@@ -195,33 +202,78 @@ def repository(self, request, *args, **kwargs):
...
+                "Repository renamed",
+                extra=dict(repoid=repo.repoid, github_webhook_event=self.event),
+            )
+        elif action == "transferred":
+            repo.author = Owner.objects.get(
+                service=self.service_name,
+                service_id=self.request.data.get("repository", {})
+                .get("owner", {})
+                .get("id"),
+            )
```
matt-codecov, line 231: probably get_or_create here too

webhook_handlers/views/github.py
ajay-sentry: I _think so_, at least it seemed to make sense according to https://www.notion.so/sentry/Database-Field-Terminology-2c501a0844bc410aa35f1ecfd3c3eb61

PR #1155, opened 2025-02-14
Cache and trace fetching the commit yaml

Swatinem:


joseph-sentry (approved):
it would be interesting to see the difference in cache memory usage as a result of this change, as a followup

PR #1161, opened 2025-02-18
fix: CusrorPaginator null encoding bug

joseph-sentry:
the CursorPaginator encodes `None` values as the string "None", and on decoding it interprets that as just another string which is causing bugs when running the SQL, because we'll be comparing random types against a string instead of against null

this fixes that by assigning a specific value (\x1f) that is unlikely to be in any user provided string that we would want to filter by

### DIFF
--- a/graphql_api/helpers/connection.py
+++ b/graphql_api/helpers/connection.py
@@ -1,9 +1,10 @@
 import enum
+from base64 import b64decode
 from dataclasses import dataclass
 from functools import cached_property
 from typing import Any, Dict, List, Optional
 
-from cursor_pagination import CursorPage, CursorPaginator
+from cursor_pagination import CursorPage, CursorPaginator, InvalidCursor
 from django.db.models import QuerySet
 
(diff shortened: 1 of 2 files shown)


graphql_api/helpers/connection.py
```diff
@@ -205,6 +206,14 @@ class DictCursorPaginator(CursorPaginator):
     if the dict access fails then it throws an exception, although it would be a different
     """
 
+    def decode_cursor(self, cursor):
+        try:
+            orderings = b64decode(cursor.encode("ascii")).decode("utf8")
+            orderings = orderings.split(self.delimiter)
+            return [None if ordering == "\x1f" else ordering for ordering in orderings]
```
adrian-codecov, line 213: Damn, weird bug/good find

adrian-codecov: Nit, make a var for \x1f since we use in multiple places and to explain what the val is/why it's needed

adrian-codecov (approved):

PR #1162, opened 2025-02-18
feat: Track Upload Sent events for Coverage/BA/TA

spalmurray-codecov:
Adds event tracking of 'Upload Sent' events for Coverage, Bundles, and test results.

It seems in some cases commit may not have an author set during TA/BA ingest. In these cases we've decided to attribute the events to a special anonymous user that has a special user id constant in shared.

Note that we are using `commit.id` instead of the commit sha because a full commit sha is functionally identifiable information (for public repos at least). No reason to use this when we have an internal id that does the same thing but is only useful to us.

### DIFF
--- a/upload/views/uploads.py
+++ b/upload/views/uploads.py
@@ -171,7 +171,12 @@ def send_analytics_data(
     AmplitudeEventPublisher().publish(
         "Upload Sent",
         {
-            "user_ownerid": commit.author.ownerid,
+            # Attribute this event to the repo owner. For BA/TA uploads we
+            # don't necessarily have the commit author at upload time, so to
+            # align the upload events, we will always attribute uploads to the
+            # repo owner. It's also not necessarily the case that the commit
+            # author is the owner performing the 'Upload Sent' action.
+            "user_ownerid": commit.repository.author.ownerid,
             "ownerid": commit.repository.author.ownerid,
             "repoid": commit.repository.repoid,
             "commitid": commit.id,  # Not commit.commitid, we do not want a commit SHA here!
--- a/upload/views/uploads.py
+++ b/upload/views/uploads.py
@@ -171,15 +171,10 @@ def send_analytics_data(
     AmplitudeEventPublisher().publish(
         "Upload Sent",
         {
-            # Attribute this event to the repo owner. For BA/TA uploads we
-            # don't necessarily have the commit author at upload time, so to
-            # align the upload events, we will always attribute uploads to the
-            # repo owner. It's also not necessarily the case that the commit
-            # author is the owner performing the 'Upload Sent' action.
-            "user_ownerid": commit.repository.author.ownerid,
+            "user_ownerid": commit.author.ownerid,
             "ownerid": commit.repository.author.ownerid,
             "repoid": commit.repository.repoid,
-            "commitid": commit.id,  # Not commit.commitid, we do not want a commit SHA here!
+            "commitid": commit.id,  # Not commit.commitid, we do not want a commit SHA here.
             "pullid": commit.pullid,
             "upload_type": "Coverage report",
         },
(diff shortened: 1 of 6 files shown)


upload/views/uploads.py
```diff
@@ -167,6 +168,22 @@ def send_analytics_data(
     AnalyticsService().account_uploaded_coverage_report(
         commit.repository.author.ownerid, analytics_upload_data
     )
+    AmplitudeEventPublisher().publish(
+        "Upload Sent",
+        {
+            # Attribute this event to the repo owner. For BA/TA uploads we
+            # don't necessarily have the commit author at upload time, so to
```
ajay-sentry, line 175: Are there some cases where we might and we are able to use that and fall back to the repository author otherwise?

ajay-sentry (approved):
lgtm!

upload/views/uploads.py
spalmurray-codecov: I think we can use it for coverage uploads, although I think it's possible for it not to exist even for those. IMO I'd rather have the attribution be less ideal but consistent. Wdyt?

ajay-sentry: It's a question of the rate at which we're losing that level of precision imo. I'd take a more "truthful" attribution 50% of the time, but if it's only like 5% I agree

spalmurray-codecov: My concern is more about usability of the data than accuracy. I agree it's less precise. I think it may be confusing to someone looking at the events in amplitude why uploads are sometimes belonging to the org and sometimes the user. I could see someone building a funnel and it just being wrong because they didn't have knowledge of this edge case. If uploads always belong to the org, then at that problem should be avoided.

spalmurray-codecov: talked about this exensively on slack, we're going with commit author and falling back to anonymous user if it doesn't exist

ajay-sentry (approved):
looks good!

upload/views/uploads.py
```diff
@@ -167,6 +168,17 @@ def send_analytics_data(
     AnalyticsService().account_uploaded_coverage_report(
         commit.repository.author.ownerid, analytics_upload_data
     )
+    AmplitudeEventPublisher().publish(
+        "Upload Sent",
+        {
+            "user_ownerid": commit.author.ownerid,
+            "ownerid": commit.repository.author.ownerid,
+            "repoid": commit.repository.repoid,
+            "commitid": commit.id,  # Not commit.commitid, we do not want a commit SHA here.
```
joseph-sentry, line 177: can you add the explanation of why to the PR description

upload/views/uploads.py
```diff
@@ -167,6 +168,17 @@ def send_analytics_data(
...
+            "pullid": commit.pullid,
```
joseph-sentry, line 178: I think it's also possible that this is `None`

are we setting the ownerid to another value in TA/BA because it's a limitation of Amplitude? or because that's the value we'd prefer to have?

joseph-sentry (approved):
generally lgtm but just a couple questions

upload/views/uploads.py
spalmurray-codecov: `pullid` being `None` is accounted for in the type over in `shared/events/amplitude/types.py`, so either `int | None` is acceptable.

I'm assuming you're talking about `user_ownerid` not `ownerid`, if so, the reason for this is because in local testing it I found if the Commit object is created during BA/TA upload, the commit author will be undefined. Ideally we attribute the event to the commit author, but if it's not defined, we will fall back to a special predefined anonymous user. We're planning to track the frequency of anonymous uploads and potentially take some action to make sure commit author is always defined should it happen frequently. Does this answer your question?

joseph-sentry: > We're planning to track the frequency of anonymous uploads and potentially take some action to make sure commit author is always defined should it happen frequently

sounds good!

PR #1165, opened 2025-02-19
makefile changes for monorepo prep

matt-codecov:
needed for https://github.com/codecov/umbrella/pull/4

we bindmount our local `codecov-api` into the docker container. currently our local `codecov-api` is the actual git repo, so the `git` invocation works. in umbrella, it's not the actual git repo, so the command fails

apparently this is all i need to change...?


adrian-codecov (approved):
Hmmm, wouldnt this be hinting at the left side of the statement having some issue? But it works?

PR #1167, opened 2025-02-20
fix: TA GQL term filtering

joseph-sentry:
instead of using startswith on the polars dataframe we want to filter by tests that contain the filtering term

### DIFF
--- a/graphql_api/tests/test_test_analytics.py
+++ b/graphql_api/tests/test_test_analytics.py
@@ -468,7 +468,7 @@ class TestAnalyticsTestCase(
     def test_test_analytics_term_filter(self, repository, store_in_redis, mock_storage):
         test_results = generate_test_results(
             repoid=repository.repoid,
-            term=rows[0]["name"],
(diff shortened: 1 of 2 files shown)


graphql_api/tests/test_test_analytics.py
```diff
@@ -468,7 +468,7 @@ def test_test_results_pagination_asc(
     def test_test_analytics_term_filter(self, repository, store_in_redis, mock_storage):
         test_results = generate_test_results(
             repoid=repository.repoid,
-            term=rows[0]["name"],
+            term=rows[0]["name"][2:],
```
adrian-codecov, line 471: Why is this 2:?

adrian-codecov (approved):

graphql_api/tests/test_test_analytics.py
joseph-sentry: oh this is just to truncate the beginning of the name so that we can test that it's actually doing the contain search instead of prefix

PR #1173, opened 2025-02-24
deps: update shared

joseph-sentry:


Swatinem (approved):

ajay-sentry (approved):

staticanalysis/tests/test_views.py
```diff
@@ -16,7 +16,7 @@
 def test_simple_static_analysis_call_no_uploads_yet(db, mocker):
     mocked_task_service = mocker.patch.object(TaskService, "schedule_task")
     mocked_presigned_put = mocker.patch(
-        "shared.api_archive.archive.StorageService.create_presigned_put",
+        "shared.storage.MinioStorageService.create_presigned_put",
```
Swatinem, line 19: as we have discussed in the retro, related to "overreliance on mocks", keep in mind that *all* of these mocks are potential breakage in production that we will never catch in tests :-(

upload/views/legacy.py
```diff
@@ -388,6 +388,7 @@ async def get_repo(self):
             raise Http404("Requested report could not be found")
         return repo
 
+    @sync_to_async
```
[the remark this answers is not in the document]
joseph-sentry: this is a workaround because now the initialization of `ArchiveService` involves feature flags which involves Django queries which can't run in an async context by default

sync_to_async allows us to make django queries in an async context

staticanalysis/tests/test_views.py
joseph-sentry: I made an [issue](https://github.com/codecov/engineering-team/issues/3411) to remove these, but i'd like to get this PR in so we don't fall too behind compared to shared

PR #1176, opened 2025-02-26
Adjust call used in comparison file to avoid provider api call

adrian-codecov:
### Purpose/Motivation
This PR intends to improve loading times in the Pull detail page by adjusting the logic in the `has_different_number_of_head_and_base_reports` resolver.  

### Links to relevant tickets
This is one of the many resolvers I'd like to tackle with the investigation done [here](https://www.notion.so/sentry/Coverage-Pulls-Tab-Cacheing-Project-19a8b10e4b5d80138e8ddf6a80b150a7).

### What does this PR do?
This resolver calls the comparison.head_report, which always calls the `apply_diff` function, which ends up calling the provider. In some cases, this call is necessary, but for this particular resolver it is not as the logic inside the resolver compares number of sessions, not the contents of the sessions. Therefore, we're going to shave some time by adding a property that doesn't talk to the provider.

I left note in that implementation, there's an opportunity to improve the algorithm altogether and use the DB only instead of GCS.

### Legal Boilerplate

### DIFF
--- a/graphql_api/types/comparison/comparison.py
+++ b/graphql_api/types/comparison/comparison.py
@@ -252,10 +252,6 @@ def resolve_has_different_number_of_head_and_base_reports(
     if "comparison" not in info.context:
         return False
     comparison: Comparison = info.context["comparison"]
-    try:
-        comparison.validate()
-    except Exception:
-        return False
     return comparison.has_different_number_of_head_and_base_sessions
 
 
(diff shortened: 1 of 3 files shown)


nora-codecov (approved):

graphql_api/types/comparison/comparison.py
```diff
@@ -252,10 +252,6 @@ def resolve_has_different_number_of_head_and_base_reports(
     if "comparison" not in info.context:
         return False
     comparison: Comparison = info.context["comparison"]
-    try:
```
adrian-codecov, line 255: Didnt specify but this is implicitly moved in the try/except inside the resolver logic

PR #1177, opened 2025-02-26
dev: Migrate from pip-compile to uv

ajay-sentry:
### Purpose/Motivation

As part of migrating over to umbrella (our future monorepo), we want to make sure everything is on the same package management system for much easier dependency management across each workspace. This PR mimics the changes done in https://github.com/codecov/worker/pull/1098 by updating the Makefile, README, and dockerfiles to support uv in lieu of pip / pip-compile.

I was honestly shocked it basically just worked after that, there were a couple dependency updates I had to do and I also removed ddtrace since I'm pretty sure we use Sentry exclusively for tracing now (confirmed with @trent-codecov) but nothing really other than that so kudos to @michelletran-codecov for a really easy to follow experience 😅 

I also removed the opentracing and setproctitle packages since I didn't see any references to those either.

I should probably also double check all the package versions uv installed and pin them to actual versions

### Links to relevant tickets

Closes https://github.com/codecov/engineering-team/issues/3371

### Notes to Reviewer

### Legal Boilerplate


pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "idna>=3.7",
+    "minio",
+    "multidict>=6.1.0",
+    "oauth2==1.9.0.post1",
+    "opentelemetry-instrumentation-django>=0.45b0",
+    "opentelemetry-sdk>=1.24.0",
+    "opentracing",
+    "polars==1.12.0",
+    "pre-commit",
+    "psycopg2-binary>=2.9.10",
```
ajay-sentry, line 41: had to add this explicitly

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "freezegun",
+    "pre-commit>=3.4.0",
+    "pytest>=8.1.1",
+    "pytest-cov>=6.0.0",
+    "pytest-asyncio>=0.14.0",
+    "pytest-django",
+    "pytest-mock",
+    "pytest-split",
+    "urllib3==1.26.19",
+    "vcrpy==4.1.*",
```
ajay-sentry, line 80: had to bump this from v6 to v4 like in the worker PR

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "ruff>=0.9.6",
```
ajay-sentry, line 81: ruff now explicitly installed as a dev dependency and updated every time make lint is run mimicking old behavior

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "ariadne==0.23",
+    "ariadne-django==0.3.0",
+    "celery>=5.3.6",
+    "cerberus",
+    "certifi>=2024.7.4",
+    "django-autocomplete-light",
+    "django-better-admin-arrayfield",
+    "django-cors-headers",
+    "django-csp",
+    "django-cursor-pagination",
```
ajay-sentry, line 17: had to add django-cursor-pagination

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "pytest-django",
+    "pytest-mock",
+    "pytest-split",
+    "urllib3==1.26.19",
+    "vcrpy==4.1.*",
+    "ruff>=0.9.6",
+]
+
+[tool.uv.sources]
+codecovopentelem = { git = "https://github.com/codecov/opentelem-python", rev = "df1e241927d4794fc0e24f96431f0fb730edac21" }
```
ajay-sentry, line 85: I can't remember why I added this, might be able to be removed

docker/Dockerfile
```diff
@@ -6,10 +6,10 @@ ARG BERGLAS_VERSION=2.0.6
 FROM us-docker.pkg.dev/berglas/berglas/berglas:$BERGLAS_VERSION as berglas
 
 FROM $REQUIREMENTS_IMAGE as app
-COPY . /app
+
 WORKDIR /app
-RUN pip install setuptools==71.1.0
-RUN python manage.py collectstatic --no-input
+ADD . /app
+# RUN python manage.py collectstatic --no-input
```
ajay-sentry, line 12: I don't _think_ we need this?

.github/workflows/ci.yml
```diff
@@ -21,25 +21,25 @@ permissions:
...
-    uses: codecov/gha-workflows/.github/workflows/lint.yml@v1.2.26
+    uses: codecov/gha-workflows/.github/workflows/lint.yml@v1.2.33
 
   build:
     name: Build API
-    uses: codecov/gha-workflows/.github/workflows/build-app.yml@v1.2.26
+    uses: codecov/gha-workflows/.github/workflows/build-app.yml@v1.2.33
     secrets: inherit
     with:
       repo: ${{ vars.CODECOV_IMAGE_V2 || 'codecov/self-hosted-api' }}
```
michelletran-codecov, line 31: I think we also need a `cache-file` to specify the right file to hash (it's picking the `requirements.txt` by default). The `gha-workflow` change for `build-app` can be seen [here](https://github.com/codecov/gha-workflows/pull/44).

.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
     needs: [build-self-hosted, test]
     secrets: inherit
     if: ${{ github.event_name == 'push' && github.event.ref == 'refs/heads/main' && github.repository_owner == 'codecov' }}
-    uses: codecov/gha-workflows/.github/workflows/self-hosted.yml@v1.2.27
+    uses: codecov/gha-workflows/.github/workflows/self-hosted.yml@v1.2.33
     with:
       push_rolling: true
       repo: ${{ vars.CODECOV_IMAGE_V2 || 'codecov/self-hosted-api' }}
```
michelletran-codecov, line 84: We also might need a `cache_file` param for this `self-hosted` workflow to pick the `uv.lock` file to hash.

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "django-filter",
+    "django-model-utils",
+    "django-postgres-extra>=2.0.8",
+    "django-prometheus",
+    "django>=4.2.16",
+    "djangorestframework==3.15.2",
+    "drf-spectacular",
+    "drf-spectacular-sidecar",
+    "elastic-apm",
+    "freezegun",
```
michelletran-codecov, line 28: I believe this is a dev dependency.

pyproject.toml
michelletran-codecov: Can we add some versioning for the dependencies that are missing them? We can use some more relaxed versioning (i.e. `>=`), but it would be best practice to include some sort of version pinning.

pyproject.toml
```diff
@@ -0,0 +1,86 @@
...
+    "gunicorn>=22.0.0",
+    "idna>=3.7",
+    "minio",
+    "multidict>=6.1.0",
+    "oauth2==1.9.0.post1",
+    "opentelemetry-instrumentation-django>=0.45b0",
+    "opentelemetry-sdk>=1.24.0",
+    "opentracing",
+    "polars==1.12.0",
+    "pre-commit",
```
michelletran-codecov, line 40: This is also a dev dependency.

michelletran-codecov (commented):
I think we can also remove the `requirements.in` and `requirements.txt`.

pyproject.toml
ajay-sentry: updateddd

pyproject.toml
ajay-sentry: updated!

pyproject.toml
ajay-sentry: totally agreed, updated to what the requirement.txt file had previously and pinned to the exact version.

In the process also removed opentracing and setproctitle since they looked unused

pyproject.toml
ajay-sentry: I lied, changed this back to match requirements.txt

docker/Dockerfile
ajay-sentry: Yeah we don't, removed

pyproject.toml
Swatinem: I believe this was part of the deprecated profiling product, so we should be killing this.

Swatinem (approved):
lgtm, I believe one `cache_file` is missing from the GHA workflow as @michelletran-codecov also mentioned.

pyproject.toml
ajay-sentry: ooooo okay good context, will remove

.github/workflows/ci.yml
ajay-sentry: Updated in https://github.com/codecov/codecov-api/pull/1177/commits/f9f22570b72250d9d2b0d2f8ff5c0afdc0de080d

JerrySentry (approved):

PR #1178, opened 2025-02-27
create coverage data in monorepo-friendly way

matt-codecov:
### Legal Boilerplate


docker/Dockerfile
```diff
@@ -25,27 +25,27 @@ RUN pip uninstall -y typing
...
-    rm /app/codecov/settings_prod.py && \
-    rm /app/codecov/settings_test.py && \
-    rm /app/codecov/settings_staging.py && \
-    rm /app/dev.sh && \
-    rm /app/migrate.sh && \
-    rm /app/prod.sh && \
-    rm /app/staging.sh && \
-    rm /app/production.yml && \
-    rm /app/development.yml
+RUN rm /app/apps/codecov-api/codecov/settings_dev.py && \
```
adrian-codecov, line 28: Optional: might be worth adding a const or something for this value to explain what it is

adrian-codecov (approved):

PR #1179, opened 2025-02-27
add max age to cookie

nora-codecov:
This change would cause impersonation to expire after 3 minutes.


adrian-codecov (commented):
Would you have to reimpersonate every 3 minutes? Or this is like 3 minutes after inactivity?

matt-codecov (approved):

adrian-codecov (approved):

PR #1181, opened 2025-02-27
Improve performance time for ImpactedFiles resolver

adrian-codecov:
### Purpose/Motivation
This PR intends to improve loading times in the Pull detail page by adjusting the logic in the `ImpactedFiles` resolver.

### Links to relevant tickets
This is one of the many resolvers I'd like to tackle with the investigation done [here](https://www.notion.so/sentry/Coverage-Pulls-Tab-Cacheing-Project-19a8b10e4b5d80138e8ddf6a80b150a7). 
[GH Ticket](https://github.com/codecov/engineering-team/issues/3397)

### What does this PR do?
I'm improving the time it takes to finish the `ImpactedFiles` resolver. This resolver is used both in the commit + pull details page, and it needlessly calls the provider every time we get to this page. That's right, we call GH/GL/BB every time you hit any of these pages to display data that doesn't need the provider (there's one part that rightfully requests data, but that's when you unfurl the specific file to see the diff data). This is because we use the `head_report` key, which always has an "apply_diff" function to it, calling GH in the process. 

Most of the data displayed in these pages either comes from our DB or from a file stored in GCS. Therefore, I'm adjusting the resolver to use a property that doesn't rely on the data that comes from github, with the caveat that I need to be mindful that I won't need that data, which I dont.

### Legal Boilerplate


JerrySentry (approved):
chatted about this, this is huge.

PR #1183, opened 2025-03-03
Replace usage of `TransactionTestCase`

Swatinem:
The `TransactionTestCase` is documented to be a bit slower than the normal one, and someone mentioned recently that this might be the reason why API tests are taking so long.

So lets try changing that, and see whether the testsuite still works :-)

### DIFF
--- a/core/commands/pull/interactors/tests/test_fetch_pull_request.py
+++ b/core/commands/pull/interactors/tests/test_fetch_pull_request.py
@@ -32,7 +32,6 @@ class FetchPullRequestInteractorTest(TestCase):
         assert pr == self.pr
 
 
-# Not part of the class because TestCase cannot be parametrized
 @freeze_time("2024-07-01 12:00:00")
 @pytest.mark.parametrize(
     "pr_state, updatestamp, expected",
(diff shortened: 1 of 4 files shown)


core/commands/pull/interactors/tests/test_fetch_pull_request.py
```diff
@@ -32,7 +32,7 @@ async def test_fetch_pull_request(self):
         assert pr == self.pr
 
 
-# Not part of the class because TransactionTestCase cannot be parametrized
+# Not part of the class because TestCase cannot be parametrized
```
ajay-sentry, line 35: Does this comment still hold true?

Swatinem: I removed this comment completely. Was probably a victim of just doing find&replace :-D

ajay-sentry: hahahah i know the feel, sounds good

ajay-sentry (approved):
tempted to say send it

Swatinem:
This has been baking on API staging for a while, so I think we are good here :-)

michelletran-codecov (approved):

PR #1184, opened 2025-03-03
Enable timeseries tests, differently

Swatinem:
Rather than messing with ENV in docker, this uses the test-global `config.yml` to enable timeseries tests, and switches to a lighter-weight timescale docker image.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -33,14 +33,14 @@ services:
           size: 2048M
 
   timescale:
-    image: timescale/timescaledb:latest-pg14
+    image: timescale/timescaledb-ha:pg14-latest
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_HOST_AUTH_METHOD=trust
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -39,9 +40,10 @@ services:
       - POSTGRES_PASSWORD=password
     volumes:
       - type: tmpfs
-        target: /var/lib/postgresql/data
+        target: /home/postgres/pgdata/data
```
joseph-sentry, line 43: did an error pop up when you changed the image that caused you to modify this or has this been incorrect the whole time?

joseph-sentry (approved):

docker-compose.yml
Swatinem: so according to docs, the postgres data is in a different directory from the upstream postgres images, but actually using that for the `tmpfs` mount errors because of permissions. really weird, but I ended up deleting the volume definition completely.

PR #1187, opened 2025-03-03
fix: TA tokenless uploads

joseph-sentry:
previously it was broken because it was trying to get the repository and commit information from the request path, when TA requests store that information in the body. So we take the same modifications that the BA tokenless authentication made but change the service to be fetched from the "service" key instead of the "git_service" key

Fixes: https://github.com/codecov/codecov-action/issues/1762

### DIFF
--- a/codecov_auth/authentication/repo_auth.py
+++ b/codecov_auth/authentication/repo_auth.py
@@ -338,6 +338,49 @@ class TokenlessAuthentication(authentication.TokenAuthentication):
             raise exceptions.AuthenticationFailed(self.auth_failed_message)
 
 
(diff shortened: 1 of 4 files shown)


adrian-codecov (approved):
I checked w/ Shelter's logic and it seems to be aligned. Could you double check thought? Should start at [get_token](https://github.com/codecov/shelter/blob/c36bf58719aad2301c93d564b0cc241d506985f5/libs/helpers.py#L495), then should go to tokenless_allowed, and then should go to _tokenless_v2_allowed (and that's the logic that should be on par). This assumes you also accept the newest "true tokenless" method as well btw, so just confirming that's fine too

PR #1195, opened 2025-03-04
feat: Track Upload Sent events for Coverage/BA/TA (attempt 2)

spalmurray-codecov:
Adds event tracking of 'Upload Sent' events for Coverage, Bundles, and test results.

It seems in some cases commit may not have an author set during upload ingest. In these cases we've decided to attribute the events to a special anonymous user that has a special user id constant in shared.

Last time I merged this PR it was breaking uploads because it seems sometimes commit author is also not defined there, which we were not expecting. This second attempt adds the fallback user id to coverage uploads as well.

Note that we are using commit.id instead of the commit sha because a full commit sha is functionally identifiable information (for public repos at least). No reason to use this when we have an internal id that does the same thing but is only useful to us.

### DIFF
--- a/upload/views/uploads.py
+++ b/upload/views/uploads.py
@@ -171,7 +171,12 @@ def send_analytics_data(
     AmplitudeEventPublisher().publish(
         "Upload Sent",
         {
-            "user_ownerid": commit.author.ownerid,
+            # Attribute this event to the repo owner. For BA/TA uploads we
+            # don't necessarily have the commit author at upload time, so to
+            # align the upload events, we will always attribute uploads to the
+            # repo owner. It's also not necessarily the case that the commit
+            # author is the owner performing the 'Upload Sent' action.
+            "user_ownerid": commit.repository.author.ownerid,
             "ownerid": commit.repository.author.ownerid,
             "repoid": commit.repository.repoid,
             "commitid": commit.id,  # Not commit.commitid, we do not want a commit SHA here!
--- a/upload/views/uploads.py
+++ b/upload/views/uploads.py
@@ -171,15 +171,10 @@ def send_analytics_data(
     AmplitudeEventPublisher().publish(
         "Upload Sent",
         {
-            # Attribute this event to the repo owner. For BA/TA uploads we
-            # don't necessarily have the commit author at upload time, so to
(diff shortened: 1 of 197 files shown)


upload/views/uploads.py
```diff
@@ -167,6 +168,19 @@ def send_analytics_data(
     AnalyticsService().account_uploaded_coverage_report(
         commit.repository.author.ownerid, analytics_upload_data
     )
+    AmplitudeEventPublisher().publish(
+        "Upload Sent",
+        {
+            "user_ownerid": commit.author.ownerid
+            if commit.author
+            else UNKNOWN_USER_OWNERID,
```
spalmurray-codecov, line 176: the only change is these lines

upload/views/uploads.py
```diff
@@ -167,6 +168,19 @@ def send_analytics_data(
...
+            "ownerid": commit.repository.author.ownerid,
+            "repoid": commit.repository.repoid,
```
joseph-sentry, line 178: putting a note here (mostly for myself) to figure out whether we would gain anything from using `author_id` and `repository_id` here instead of accessing those values via attrs

joseph-sentry (approved):

upload/views/uploads.py
spalmurray-codecov: not sure I follow. Where are `author_id` and `repository_id` defined?

joseph-sentry: they are not defined explicitly anywhere:  here's some context https://docs.djangoproject.com/en/dev/ref/models/fields/#database-representation

 there's nothing todo on this PR for this comment, i just wanted to make a note to revisit this and check whether we can avoid some reads from the database here

spalmurray-codecov: ahh okay I'm with you. I've been under the assumption that there are no reads for accessing like this. Assumed it would be populated on the already read commit in scope here (though we may need to explicitly add a select_related here to achieve that). Please lmk if I'm wrong I do not want to be doing more than one fetch for the commit's attrs. Side Q does django automatically cache any of this stuff?

ajay-sentry (approved):
looks good!

PR #1196, opened 2025-03-04
tt1

JerrySentry:
### Legal Boilerplate


services/components.py
```diff
@@ -26,7 +26,7 @@ def commit_components(commit: Commit, owner: Owner | None) -> List[Component]:
 
 
 def component_filtered_report(
-    report: Report, components: List[Component]
+    report: Report | None, components: List[Component]
```
suejung-sentry, line 29: corresponds with this issue - https://codecov.sentry.io/issues/5239860853

PR #1216, opened 2025-03-18
feat: Add component filter to badges

spalmurray-codecov:
Adds a component filter for our coverage badge.

Closes https://github.com/codecov/engineering-team/issues/3223


graphs/views.py
```diff
@@ -156,6 +161,39 @@ def flag_coverage(self, flag_name, commit):
...
+
+        Parameters
+        component_id (string): id or name of component
+        commit (obj): commit object containing report
+        """
+        report = commit.full_report
+        if report is None:
+            log.warning(
+                "Commit's report not found",
+                extra=dict(commit=commit, component=component_id),
```
ajay-sentry, line 176: nit: should we just log the commitid rather than the full commit obj?

graphs/views.py
```diff
@@ -156,6 +161,39 @@ def flag_coverage(self, flag_name, commit):
...
+            )
+            return None
+        components = commit_components(commit, None)
+
+        try:
+            component = next(
+                c
+                for c in components
+                if c.component_id == component_id or c.name == component_id
```
ajay-sentry, line 185: In what scenario would the component name be the component_id?

graphs/views.py
```diff
@@ -156,6 +161,39 @@ def flag_coverage(self, flag_name, commit):
...
+            )
+        except StopIteration:
```
ajay-sentry, line 187: Haven't seen this one before, is this the exception that occurs when the iterator fails to fetch the next element?

spalmurray: this occurs when no match is found

graphs/views.py
spalmurray: couldn't think of a better name for `component_id` lol. It can be the `component.component_id` or the `component.name`, we will accept either basically.

graphs/views.py
```diff
@@ -156,6 +161,39 @@ def flag_coverage(self, flag_name, commit):
...
+            # Component not found
+            return None
+
+        component_flags = component.get_matching_flags(report.get_flag_names())
+
+        filtered_report = report.filter(flags=component_flags, paths=component.paths)
+
+        return filtered_report.totals.coverage
```
ajay-sentry, line 195: will totals ever be none on filtered_report?

spalmurray: pretty sure that can happen yes. I will add a test case for this

graphs/views.py
ajay-sentry: ahhh I see, its also doubly confusing because the param being passed into the param is just called "component" 😅 

Maybe something like component_identifier might help ease cognitive load a bit

spalmurray: I like it

PR #1221, opened 2025-03-19
Batch triggering timeseries backfills

Swatinem:
Previously, the "owner measurements" would trigger backfilling of timeseries data for each repo one-by-one, leading to suboptimal N+1 queries.

The logic is now changed to do this as a batch, so we only run a single query to figure out the start/end date of the backfill.

---

Fixes https://github.com/codecov/internal-issues/issues/1222

### DIFF
--- a/core/commands/repository/interactors/activate_measurements.py
+++ b/core/commands/repository/interactors/activate_measurements.py
@@ -20,11 +20,9 @@ class ActivateMeasurementsInteractor(BaseInteractor):
         )
 
         dataset, created = Dataset.objects.get_or_create(
-            name=measurement_type.value,
-            repository_id=repo.pk,
+            name=measurement_type.value, repository_id=repo.pk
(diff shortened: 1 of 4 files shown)


Swatinem:
I verified manually that the query that is generated looks like this:
```sql
select
	"commits"."repoid",
	MIN("commits"."timestamp") as "start_date",
	MAX("commits"."timestamp") as "end_date"
from
	"commits"
where
	"commits"."repoid" in (…)
group by
	"commits"."repoid"
```

… and also that this is using an `Index Only Scan using commits_repoid_timestamp_desc on commits` which is pretty much the fastest query you can do, so I think we are good :-)

adrian-codecov (approved):
Sweet! Good perf improvement! Lets monitor too how those N+1 queries do after (hopefully disappear) and see some improvements!

PR #1223, opened 2025-03-19
fix: Capture uploads for upload-only endpoint also

spalmurray-codecov:
Based on in-person conversation we figured out we were missing metrics from the upload-only endpoint. The metric was only added to the new do it all create-commit create-report create-upload endpoint, but some uploads still go the old ones and we were missing those.


upload/views/upload_coverage.py
spalmurray-codecov, line 1: This function calls `create_upload` below so by moving the event into there we get both new and old endpoints accounted for.

joseph-sentry (approved):
👍

PR #1226, opened 2025-03-20
Avoid an unnecessary async indirection

Swatinem:
For reasons I do not understand, an async call was passed as the single future through a `gather`. This now cuts down on that indirection and uses `async_to_sync` directly.


ajay-sentry (approved):
send it, nice find

PR #1228, opened 2025-03-20
chore: improve activation logging

drazisil-codecov:


Swatinem (commented):
it would be better to use `extra` fields for this, so you can more efficiently search for it.

Swatinem (approved):

PR #1232, opened 2025-03-21
Update API to Python 3.13

Swatinem:
We have been running`worker` on Python 3.13 for a while now, but never ended up updating API.

So lets do this :-)


ajay-sentry (approved):
LOVE IT

nora-codecov (approved):

PR #1235, opened 2025-03-24
feat: Add bundle badge svgs

spalmurray:
Adds a function for creating bundle analysis svg badge from a bundle size in bytes and optionally a level of precision for rounding. They are not yet used anywhere - that will be the next PR.

Screenshot 2025-03-24 at 14 09 17

Closes https://github.com/codecov/engineering-team/issues/3487

### DIFF
--- a/graphs/helpers/badge.py
+++ b/graphs/helpers/badge.py
@@ -50,3 +50,53 @@ def format_coverage_precision(coverage, precision):
     precision = int(precision)
     coverage = float(coverage)
     return ("%%.%sf" % precision) % coverage
+
+
+def get_bundle_badge(bundle_size_bytes: int, precision: int):
+    bundle_size_string = format_bundle_bytes(bundle_size_bytes, precision)
+    char_width = 7  # approximate, looks good on all reasonable inputs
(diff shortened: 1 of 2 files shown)


graphs/helpers/badge.py
```diff
@@ -50,3 +50,53 @@ def format_coverage_precision(coverage, precision):
...
+    static_width = 57  # width of static elements in the svg (text + margins)
+
+    width = static_width + width_in_pixels
+
+    return f""" <svg xmlns="http://www.w3.org/2000/svg" width="{width}" height="20">
+    <linearGradient id="CodecovBadgeGradient" x2="0" y2="100%">
+        <stop offset="0" stop-color="#bbb" stop-opacity=".1" />
+        <stop offset="1" stop-opacity=".1" />
+    </linearGradient>
+    <mask id="CodecovBadgeMask{width}px">
```
spalmurray, line 68: found that ids can leak into other svgs, so trying to make the id very specific.

spalmurray: Nope, figured giving it a sensical id would be better for readability. This particular element will be the same for all svgs of the same width so there won't be side effects of it leaking.

PR #1239, opened 2025-03-25
feat: Create API for bundle size badges

spalmurray:
Adds new graph endpoints for getting new bundle analysis badges.

- Bundle name is required in the url
- Branch name is optional param
- Precision is optional param (one of 0, 1, 2(default))

Example:
[codecov](https://stage-web.codecov.dev/gh/codecov/gazebo/bundles/main/gazebo-staging-system)
[codecov](https://stage-web.codecov.dev/gh/codecov/gazebo/bundles/main/gazebo-staging-system)
[codecov](https://stage-web.codecov.dev/gh/codecov/gazebo/bundles/main/gazebo-staging-system)
`https://stage-api.codecov.dev/gh/codecov/gazebo/main/graph/bundle/gazebo-staging-system/badge.svg?precision=2`

Closes https://github.com/codecov/engineering-team/issues/3488

### DIFF
--- a/graphs/views.py
+++ b/graphs/views.py
@@ -13,9 +13,8 @@ from shared.metrics import Counter, inc_counter
 
 from api.shared.mixins import RepoPropertyMixin
 from core.models import Branch, Pull
-from graphql_api.dataloader.bundle_analysis import load_bundle_analysis_report
 from graphs.settings import settings
-from services.bundle_analysis import BundleAnalysisReport
+from services.bundle_analysis import BundleAnalysisReport, load_report
 from services.components import commit_components
 
 from .helpers.badge import (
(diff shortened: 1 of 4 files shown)


graphs/views.py
```diff
@@ -193,6 +199,83 @@ def component_coverage(self, component_identifier: str, commit: Commit):
...
+        branch_name = self.kwargs.get("branch") or repo.branch
+        branch = Branch.objects.filter(
+            name=branch_name, repository_id=repo.repoid
+        ).first()
+
+        if branch is None:
+            log.warning(
+                "Branch not found", extra=dict(branch_name=branch_name, repo=repo)
+            )
+            return None
```
nicholas-codecov, line 250: Should we throw a 404 here, when a commit is not found, or if the bundle is not found?

nicholas-codecov (approved):

graphs/views.py
spalmurray: Coverage follows this pattern, we turn any None into a badge that says `unknown`. I think the motivation for this is so that the badge will always show *something* instead of looking broken

nicholas-codecov: Ahhh i see i see, sounds good 👍

PR #1260, opened 2025-03-31
feat: add ability to read new file format

joseph-sentry:
i'm introducing 2 things here:
- the ability to read the output of the new ta cache rollups code in
  the new location in GCS
- the ability to read multiple versions of the new rollup file
    - its probable that the "no version" file never exists in prod GCS
      but i'd like to establish this pattern in the code nonetheless

i also add metrics

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - RUN_ENV=DEV
       # Improves pytest-cov performance in python 3.12
       # https://github.com/nedbat/coveragepy/issues/1665#issuecomment-1937075835
-      - COVERAGE_CORE=sysmon 
+      - COVERAGE_CORE=sysmon
     env_file:
       - .testenv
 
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -42,3 +42,16 @@ services:
   redis:
     image: redis:6-alpine
 
+  minio:
+    image: minio/minio:latest
+    command: server /export
+    ports:
+      - "${MINIO_PORT:-9000}:9000"
```
Swatinem, line 49: I noticed recently that shared/worker are conflicting with each other, as they are exposing this port to the host. we don’t need this in order to run tests, so I would just not do this.

Swatinem (approved):

spalmurray (approved):
skimmed and trust Arpad. Just codeowner stamping for ya

spalmurray:
apparently I'm not a member of the team LOL. You are @joseph-sentry can you stamp your own PR 😆

JerrySentry (approved):
lgtm

PR #1263, opened 2025-04-01
Avoid N+1 queries fetching `Upload.errors`

Swatinem:
Previously, this was going through an unnecessarily complex Java-esque design pattern, just to execute a query that can be much better served by using a `prefetch_related` Django combinator.

---

This should fix https://github.com/codecov/engineering-team/issues/3399


graphql_api/types/upload/upload.py
```diff
@@ -44,15 +42,7 @@ def resolve_upload_type(upload, info) -> UploadType:
...
-    command = info.context["executor"].get_command("upload")
-    queryset = await command.get_upload_errors(report_session)
-    result = await queryset_to_connection(
-        queryset,
-        ordering=("updated_at",),
-        ordering_direction=OrderingDirection.ASC,
-        **kwargs,
-    )
-    return result
+    return report_session.errors
```
joseph-sentry, line 45: i think we may still have to `queryset_to_connection` this, connections are this GQL api specific abstraction we have for doing pagination, i'm not sure if report_session.errors is suitable here

PR #1265, opened 2025-04-02
Move trace annotation to correct fn

Swatinem:
By looking at some random traces, I found that the functions being traced were only *creating* querysets, but they were not the functions that actually *evaluated* those querysets.

### DIFF
--- a/graphql_api/actions/repository.py
+++ b/graphql_api/actions/repository.py
@@ -1,7 +1,6 @@
 import logging
 from typing import Any
 
-import sentry_sdk
 from django.db.models import QuerySet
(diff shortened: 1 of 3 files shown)


graphql_api/actions/repository.py
```diff
@@ -72,16 +70,13 @@ def list_repository_for_owner(
     if exclude_okta_enforced_repos:
         queryset = queryset.exclude_accounts_enforced_okta(okta_account_auths)
 
-    if not ai_enabled_filter:
```
Swatinem, line 75: no to reviewers: expand the code above this change, and you will see that we early-return `if ai_enabled_filter` above, so this branch here was always true.

joseph-sentry (approved):

PR #1267, opened 2025-04-02
Bundle Analysis: fix trends when no data in query range

JerrySentry:
### Legal Boilerplate

### DIFF
--- a/graphql_api/actions/measurements.py
+++ b/graphql_api/actions/measurements.py
@@ -16,7 +16,7 @@ def measurements_by_ids(
     before: datetime,
     after: Optional[datetime] = None,
     branch: Optional[str] = None,
-) -> Dict[int, List[Dict[str, Any]]]:
+) -> Dict[Any, List[Dict[str, Any]]]:
     queryset = MeasurementSummary.agg_by(interval).filter(
         name=measurable_name,
         owner_id=repository.author_id,
--- a/services/bundle_analysis.py
+++ b/services/bundle_analysis.py
@@ -454,7 +454,7 @@ class BundleAnalysisMeasurementsService(object):
     @sentry_sdk.trace
     def _compute_measurements(
         self, measurable_name: str, measurable_ids: List[str]
-    ) -> Dict[int, List[Dict[str, Any]]]:
+    ) -> Dict[str, List[Dict[str, Any]]]:
         all_measurements = measurements_by_ids(
             repository=self.repository,
             measurable_name=measurable_name,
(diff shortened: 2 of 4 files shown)


graphql_api/actions/measurements.py
```diff
@@ -16,7 +16,7 @@ def measurements_by_ids(
     before: datetime,
     after: Optional[datetime] = None,
     branch: Optional[str] = None,
-) -> Dict[int, List[Dict[str, Any]]]:
```
ajay-sentry, line 19: was this change on purpose? Should we also change the type on 41 as well

timeseries/tests/test_helpers.py
```diff
@@ -1178,9 +1178,12 @@ def test_no_dataset(self, trigger_backfill):
                 repository_id__in=[self.repo1.pk, self.repo2.pk],
             )
         )
```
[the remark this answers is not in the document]
ajay-sentry: could also confirm the items in the set as well

services/bundle_analysis.py
```diff
@@ -481,8 +483,10 @@ def _compute_measurements(
                 # If there isn't any measurements before the start date range, measurements will be untouched
                 if carryover_measurement:
                     value = Decimal(carryover_measurement[0]["value"])
-                    carryover = dict(measurements[0])
-                    carryover["timestamp_bin"] = self.after
+                    carryover = dict(measurements[0]) if measurements else {}
+                    carryover["timestamp_bin"] = self.after.replace(
```
ajay-sentry, line 487: is this to "zero out" the timestamp?

graphql_api/actions/measurements.py
JerrySentry: yeah because when this function was made the measurable_id was only ints representing the flag IDs for flag measurements, but it can be also str for component names or BA reports.

Yes updating line 41.

services/bundle_analysis.py
JerrySentry: updated.

timeseries/tests/test_helpers.py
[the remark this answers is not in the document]
JerrySentry: updated

services/bundle_analysis.py
JerrySentry: yes, in the downstream when building the list of measurements to return it needs to have 00:00:00.000 as the time, that is expected because normally we pull the data from the aggregate tables, naturally it always agg by n days and so the time is always 0s. But in this case when we carry over the last datapoint we don't query from the aggregate tables, instead it comes from the un-agg'd table so the time is when it got inserted. So this part would zero out the times to comply with the logic later.

ajay-sentry: ahhhh okay makes sense

ajay-sentry (approved):
send it

PR #1269, opened 2025-04-03
Improve BaseLoader and other type annotations

Swatinem:
With this change, we now have proper types and thus autocompletion for all the XYZLoader.loader() calls.


adrian-codecov (approved):
Ty for types 🙇

PR #1271, opened 2025-04-03
Use the `CommitLoader` for `bundleAnalysisCompareWithParent`

Swatinem:
This is loading the `parent_commit` using the async batch-loading `CommitLoader`.

This way, we can avoid one of the N+1 queries being executed when executing the `GetCommits` GraphQL query.

However, we are still loading all the BA `CommitReport`s separately within `load_bundle_analysis_comparison`, which is another 2N+1 problem on its own. Will solve that one another time :-)

### DIFF
--- a/graphql_api/types/commit/commit.py
+++ b/graphql_api/types/commit/commit.py
@@ -393,16 +393,19 @@ def resolve_coverage_components(commit: Commit, info, filters=None) -> List[Comp
 
 
 @commit_bundle_analysis_bindable.field("bundleAnalysisCompareWithParent")
-@sync_to_async
 @sentry_sdk.trace
-def resolve_commit_bundle_analysis_compare_with_parent(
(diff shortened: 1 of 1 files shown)


adrian-codecov (approved):
Can you explain a bit more how making the call async to load_... helps w/ the N+1?

Swatinem:
https://github.com/syrusakbary/aiodataloader?tab=readme-ov-file#batching explains it quite well.

Assuming we have an async framework based on an event-loop. We can call the async fns, and they will run immediately up until the first `await` point. So they will observe a bunch of `load` calls.

On the next tick of the event loop, all the `load` calls that happened on the previous tick are being batched together and executed as one.

The graphql framework in the background is executing all the resolvers concurrently, meaning all of them are running up until the first `await Loader.load()`. Then the dataloader does its magic and batches all those load requests together.

PR #1295, opened 2025-04-23
feat(admin): add action to trigger sync repos

giovanni-guidini:
This new action triggers the "sync_repos" for the selected owner from the admin dashboard. Would have been helpful for me today.

(only tested through the unit tests)

### DIFF
--- a/codecov_auth/admin.py
+++ b/codecov_auth/admin.py
@@ -12,6 +12,8 @@ from django.forms import CheckboxInput, Select, Textarea
 from django.http import HttpRequest
 from django.shortcuts import redirect, render
 from django.utils import timezone
+from services.task import TaskService
+
 from django.utils.html import format_html
 from shared.django_apps.codecov_auth.models import (
     Account,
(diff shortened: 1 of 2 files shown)


codecov_auth/admin.py
```diff
@@ -111,6 +113,40 @@ def impersonate_owner(self, request, queryset):
...
+        manual_trigger=True,
+    )
+
+    self.message_user(
+        request,
+        f"Refresh task triggered for {owner.username} (ownerid: {owner.ownerid})",
+        level=messages.SUCCESS,
+    )
+    History.log(
+        Owner.objects.get(ownerid=owner.ownerid),
```
Swatinem, line 140: this would just be `owner`, right?

codecov_auth/admin.py
```diff
@@ -111,6 +113,40 @@ def impersonate_owner(self, request, queryset):
 impersonate_owner.short_description = "Impersonate the selected owner"
 
 
+def refresh_owner(self, request, queryset):
+    if queryset.count() != 1:
+        self.message_user(
+            request, "You must refresh exactly one Owner.", level=messages.ERROR
```
Swatinem, line 119: I think it should be trivially possible to also support multiple-select.

Swatinem (approved):

codecov_auth/admin.py
giovanni-guidini: SO I copied what `impersonate_owner` does :P

